PR #3, opened 2019-10-24
Disambiguate Polysemy FromJSON instance usage

liftM:
I added some configuration files so `ghcide` doesn't blow up in my face immediately.

The important part of this diff is the change in `Pipenv.hs`. Since `fileInputJson :: (FromJSON a, Members '[ReadFS, Error CLIErr] r) => Path b File -> InterpreterFor (Input a) r`, and the uasge of `fileInputJson` in `Pipenv.hs:77` applies no additional constraints, GHC is unable to disambiguate the type variable `a`. This diff explicitly types `a` to be `PipfileLock` to remove that ambiguity.

### DIFF
--- a/src/Strategy/Python/Pipenv.hs
+++ b/src/Strategy/Python/Pipenv.hs
@@ -24,6 +24,7 @@ import Prologue
 import qualified Data.Map.Strict as M
 import qualified Data.Text as T
 import           Polysemy
+import           Polysemy.Error hiding (catch)
 import           Polysemy.Input
 import           Polysemy.State
 import           Polysemy.Output
@@ -35,7 +36,6 @@ import           Effect.ReadFS
 import qualified Graph as G
 import           Types
 
-import           Polysemy.Error hiding (catch)
 import Diagnostics
 
 discover :: Discover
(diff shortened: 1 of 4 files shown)


src/Strategy/Python/Pipenv.hs
```diff
@@ -71,7 +74,7 @@ strategyWithCmd = Strategy
 strategyNoCmd :: Strategy BasicFileOpts
 strategyNoCmd = Strategy
   { strategyName = "python-pipfile"
-  , strategyAnalyze = \opts -> analyzeNoCmd & fileInputJson (targetFile opts)
+  , strategyAnalyze = \opts -> analyzeNoCmd & (fileInputJson :: (Members '[ReadFS, Error CLIErr] r) => Path b File -> InterpreterFor (Input PipfileLock) r) (targetFile opts)
```
cnr: There's a really useful language extension called `TypeApplications` that lets you specify concrete types for type variables:

For example, with `const`:
```hs
> :type const
a -> b -> a
> :type const @Int
Int -> b -> Int
> :type const @Int @String
Int -> String -> Int
```

We use this [elsewhere in the codebase](https://github.com/fossas/spectrometer/blob/master/src/App/Scan.hs#L55) when disambiguating types

You should be able to use something similar here with `fileInputJson @... (targetFile opts)`

PR #15, opened 2019-12-12
Glide lockfile parsing support

zlav:
This PR adds support for parsing `glide.lock` files. The file is yaml formatted and as a result this PR added supporting functions to parse yaml files into Aeson data. Fortunately the `Data.yaml` package supports this out of the box.

### DIFF
--- a/src/Effect/ReadFS.hs
+++ b/src/Effect/ReadFS.hs
@@ -12,9 +12,11 @@ module Effect.ReadFS
 
   , readContentsParser
   , readContentsJson
+  , readContentsYaml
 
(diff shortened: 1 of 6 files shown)


src/Effect/ReadFS.hs
```diff
@@ -71,6 +74,14 @@ readContentsJson file = do
     Left err -> throw (FileParseError (toFilePath file) (T.pack err))
     Right a -> pure a
 
+-- | Read JSON from a file
```
cnr:
```suggestion
-- | Read YAML from a file
```

PR #16, opened 2019-12-18
fix go.mod parsing

cnr:
- `replace` directives may have a semver on the left side of the fat arrow
- Package names may contain underscores


src/Strategy/Go/Gomod.hs
```diff
@@ -128,7 +128,7 @@ gomodParser = do
 
   -- parse the body of a single replace (wihtout the leading "replace" lexeme)
```
zlav: [nit grammar] comment "wihtout"

PR #17, opened 2020-01-06
Add strategy design docs

cnr:
[rendered docs](https://github.com/fossas/spectrometer/blob/designdocs/docs/strategies.md)

This moves much of the content from `rfcs` into this repo

Open questions:

- should `gradle` and `maven` be combined into a single `java` category?
- is there additional detail we should include for some of these?

### DIFF
--- a/docs/strategies.md
+++ b/docs/strategies.md
@@ -1,8 +1,9 @@
 # Analysis Strategies
 
-We support dependency analysis for the following languages:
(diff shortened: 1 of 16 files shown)


scruffystuffs:
Is there anything preventing us from merging this?

scruffystuffs:
Confirmed with @cnr, we're merging this.

PR #18, opened 2020-01-07
Support for the .NET Package Reference lockfile

zlav:
This PR adds support for reading and interpreting the `project.assets.json` file and constructing an accurate graph.

This PR also includes the .NET support RFC located [here](https://github.com/fossas/spectrometer/blob/feat/packagereference-lockfile/docs/languages/dotnet.md)


src/Strategy/NuGet/ProjectAssetsJson.hs
```diff
@@ -0,0 +1,93 @@
...
+  { targets     :: M.Map Text (M.Map Text NugetDep)
+  } deriving Show
+
+instance FromJSON ProjectAssetsJson where
+  parseJSON = withObject "ProjectAssetsJson" $ \obj ->
+    ProjectAssetsJson <$> obj .: "targets"
+
+data NugetDep = NugetDep
+  { depType    :: Text
+  , deepDeps   :: Maybe (M.Map Text Text)
```
cnr: I'd recommend making this field just `deepDeps :: M.Map Text Text` -- it'll make this easier to consume in `buildGraph`

in the FromJSON parser below, you can use `(.!=) :: Parser (Maybe a) -> a -> Parser a` like:
```
instance FromJSON NugetDep where
  parseJSON = withObject "Dependency" $ \obj ->
    NugetDep <$> obj .: "type"
             <*> obj .:? "dependencies" .!= M.empty
```

src/Strategy/NuGet/ProjectAssetsJson.hs
```diff
@@ -0,0 +1,93 @@
...
+  } deriving Show
+   
+data CompleteNugetDep = CompleteNugetDep
+  { depName            :: Text
+  , depVersion         :: Text
+  , completeDepType    :: Text
+  , completeDeepDeps   :: Maybe (M.Map Text Text)
```
cnr: (similar to NugetDep comment)
change to `completeDeepDeps :: M.Map Text Text` -- and use `M.empty` as appropriate

src/Strategy/NuGet/ProjectAssetsJson.hs
```diff
@@ -0,0 +1,93 @@
...
+    NugetDep <$> obj .: "type"
+             <*> obj .:? "dependencies"
+
+analyze :: Member (Input ProjectAssetsJson) r => Sem r (Graphing Dependency)
+analyze = buildGraph <$> input
+
+buildGraph :: ProjectAssetsJson -> Graphing Dependency
+buildGraph project = unfold direct deepList toDependency
+    where
+    direct = concat $ (\(_,a) -> ((\(x,y) -> CompleteNugetDep (head $ T.splitOn "/" x) (last $ T.splitOn "/" x) (depType y) (deepDeps y)) <$> (M.toList a))) <$> (M.toList (targets project))
```
cnr: `concat $ f <$> xs`
is the same as
`concat (map f xs)`
is the same as
`concatMap f xs`
(or `f =<< xs`, but `concatMap` is a clearer name)

let's use `concatMap` here

cnr: `head` and `last` are problematic: they're "partial" functions (i.e. they can throw runtime exceptions) when the list is empty:
```hs
head :: [a] -> a
head (x:_) = x
head _ = error "empty list"

last :: [a] -> a
last [x] = x
last (_:xs) = last xs
last _ = error "empty list"
```

unfortunately the prelude contains a bunch of these partial functions.

we do have a safer alternative, though: pattern matching!
```hs
case T.splitOn "/" depString of
    [name,ver] -> undefined -- we know we have two parts
    _ -> undefined -- and this forces us to consider the empty case (or explicitly ignore it)
```

src/Strategy/NuGet/ProjectAssetsJson.hs
```diff
@@ -0,0 +1,93 @@
...
+  , discoverFunc = discover'
+  }
+
+discover' :: Members '[Embed IO, Output ConfiguredStrategy] r => Path Abs Dir -> Sem r ()
+discover' = walk $ \_ subdirs files -> do
+  case find (\f -> fileName f == "project.assets.json") files of
+    Just file -> output (configure file)
+    Nothing -> pure ()
+
+  walkSkipNamed ["node_modules/"] subdirs
```
cnr: do we actually want to skip `node_modules`?

src/Strategy/NuGet/ProjectAssetsJson.hs
zlav: This is a really great solution and saved me from having to care about the Maybe as well. Good catch, thanks!

src/Strategy/NuGet/ProjectAssetsJson.hs
zlav: Mistake

src/Strategy/NuGet/ProjectJson.hs
```diff
@@ -0,0 +1,100 @@
...
+    Nothing -> pure ()
+
+  walkContinue
+
+strategy :: Strategy BasicFileOpts
+strategy = Strategy
+  { strategyName = "nuget-projectjson"
+  , strategyAnalyze = \opts -> analyze
+      & fileInputJson @ProjectJson (targetFile opts)
+  , strategyModule = parent. targetFile
```
cnr: [nit] space before the dot

src/Discovery.hs
```diff
@@ -55,7 +58,10 @@ discoverFuncs =
 
 strategyGroups :: [StrategyGroup]
 strategyGroups =
-  [ StrategyGroup "gradle"
+  [ StrategyGroup "dotnet"
+    [ SomeStrategy ProjectAssetsJson.strategy
```
cnr: Make sure to add to the strategygroup/discoverfuncs

PR #19, opened 2020-01-07
Don't export partial functions in Prologue

cnr:
partial functions like these can crash at runtime, so we'd rather not use them by default

### DIFF
--- a/src/Strategy/Go/Types.hs
+++ b/src/Strategy/Go/Types.hs
@@ -8,7 +8,8 @@ module Strategy.Go.Types
   )
   where
 
-import Prologue hiding (parent)
+import qualified Prelude as Unsafe
+import           Prologue hiding (parent)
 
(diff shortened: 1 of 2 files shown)


src/Strategy/Go/Types.hs
```diff
@@ -59,8 +60,8 @@ golangPackageToDependency pkg = foldr applyLabel start
 
 -- replace "v0.0.0-20191212000000-abcdef+incompatible" with "abcdef"
 fixVersion :: Text -> Text
-fixVersion = last . T.splitOn "-" . T.replace "+incompatible" ""
+fixVersion = Unsafe.last . T.splitOn "-" . T.replace "+incompatible" ""
```
cnr: Using `last` here is safe because `splitOn` _always_ returns a non-empty list.

(I wish it'd return a `Data.List.NonEmpty` instead)

PR #20, opened 2020-01-15
Nuspec, packages.config, and package reference support

zlav:
This PR adds support for parsing `.nuspec`, `packages.config`, and package reference files for dependencies.

### DIFF
--- a/hscli.cabal
+++ b/hscli.cabal
@@ -120,6 +120,7 @@ library
     Strategy.Node.YarnLock
     Strategy.NpmList
     Strategy.NuGet.PackagesConfig
+    Strategy.NuGet.PackageReference
     Strategy.NuGet.ProjectAssetsJson
     Strategy.NuGet.ProjectJson
     Strategy.NuGet.Nuspec
@@ -163,6 +164,7 @@ test-suite tests
     Node.NpmLockTest
     Node.PackageJsonTest
     NuGet.PackagesConfigTest
+    NuGet.PackageReferenceTest
     NuGet.ProjectAssetsJsonTest
     NuGet.ProjectJsonTest
     NuGet.NuspecTest
--- a/hscli.cabal
+++ b/hscli.cabal
@@ -75,7 +75,7 @@ common deps
     , vector                       ^>=0.12.0.3
     , yaml                         ^>=0.11.1
     , yarn-lock                    ^>=0.6.2
-    , xml
+    , xml                          ^>=1.3.14
 
 library
   import:          lang
--- a/hscli.cabal
+++ b/hscli.cabal
@@ -73,9 +73,9 @@ common deps
     , typed-process                ^>=0.2.6
     , unordered-containers         ^>=0.2.10
     , vector                       ^>=0.12.0.3
+    , xml                          ^>=1.3.14
     , yaml                         ^>=0.11.1
     , yarn-lock                    ^>=0.6.2
-    , xml                          ^>=1.3.14
 
 library
   import:          lang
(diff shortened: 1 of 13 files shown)


src/Strategy/NuGet/Nuspec.hs
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+analyze :: Members '[ReadFS, Error ReadFSErr] r => BasicFileOpts -> Sem r (Graphing Dependency)
+analyze BasicFileOpts{..} = do
+      contents <- readContentsBS targetFile
+      let nuspecGroups = parseNuspec =<< XML.parseXMLDoc contents
+      
+      case nuspecGroups of
+        Just gs -> pure $ buildGraph gs
+        Nothing -> undefined
```
cnr: `undefined` 😱

src/Strategy/NuGet/Nuspec.hs
```diff
@@ -0,0 +1,109 @@
...
+  , strategyAnalyze = analyze
+  , strategyModule = parent . targetFile
+  , strategyOptimal = NotOptimal
+  , strategyComplete = NotComplete
+  }
+
+analyze :: Members '[ReadFS, Error ReadFSErr] r => BasicFileOpts -> Sem r (Graphing Dependency)
+analyze BasicFileOpts{..} = do
+      contents <- readContentsBS targetFile
+      let nuspecGroups = parseNuspec =<< XML.parseXMLDoc contents
```
cnr: can we extract this pattern into helpers in `ReadFS`? -- similar to json/parser/yaml

src/Strategy/NuGet/Nuspec.hs
```diff
@@ -0,0 +1,109 @@
...
+  childByName :: String -> XML.Element -> Maybe XML.Element
+  childByName name = XML.filterChildName (\elName -> XML.qName elName == name)
+
+  childrenByName :: String -> XML.Element -> [XML.Element]
+  childrenByName name = XML.filterChildrenName (\elName -> XML.qName elName == name)
+
+buildGraph :: [Group] -> Graphing Dependency
+buildGraph project = unfold direct (const []) toDependency
+    where
+    direct = foldr (\x y -> y ++ dependencies x) [] project
```
cnr: `concatMap :: (a -> [b]) -> [a] -> [b]` codifies this pattern -- which might be a little easier to understand later. Let's use that instead

src/Strategy/NuGet/PackagesConfig.hs
```diff
@@ -0,0 +1,93 @@
...
+  , strategyComplete = NotComplete
+  }
+
+analyze :: Members '[ReadFS, Error ReadFSErr] r => BasicFileOpts -> Sem r (Graphing Dependency)
+analyze BasicFileOpts{..} = do
+      contents <- readContentsBS targetFile
+      let packagesConfig = parsePackagesConfig =<< XML.parseXMLDoc contents
+      case packagesConfig of
+        Just gs -> pure $ buildGraph gs
+        Nothing -> undefined
```
cnr: `undefined` again 😱

src/Strategy/NuGet/PackagesConfig.hs
```diff
@@ -0,0 +1,93 @@
...
+  , strategyAnalyze = analyze
+  , strategyModule = parent . targetFile
+  , strategyOptimal = NotOptimal
+  , strategyComplete = NotComplete
+  }
+
+analyze :: Members '[ReadFS, Error ReadFSErr] r => BasicFileOpts -> Sem r (Graphing Dependency)
+analyze BasicFileOpts{..} = do
+      contents <- readContentsBS targetFile
+      let packagesConfig = parsePackagesConfig =<< XML.parseXMLDoc contents
```
cnr: Can we extract this pattern to a helper in `ReadFS`? similar to json/parsec/yaml. We can use that helper in both of these strategies

hscli.cabal
```diff
@@ -75,6 +75,7 @@ common deps
     , vector                       ^>=0.12.0.3
     , yaml                         ^>=0.11.1
     , yarn-lock                    ^>=0.6.2
+    , xml
```
cnr:
```suggestion
    , xml                          ^>=1.3.14
```
(adding the version bound)

src/Strategy/NuGet/Nuspec.hs
zlav: Didn't know about that, good suggestion

PR #21, opened 2020-01-18
Use github actions for CI

cnr:
This also uploads artifacts. While there isn't an API to download those artifacts yet, it's on their roadmap in the next couple of weeks


cabal.project.ci
```diff
@@ -0,0 +1,4 @@
+packages: .
+
+package hscli
+  ghc-options: -Werror
```
cnr: This prevents the build from succeeding when there are warnings. I'd like to do something similar with `hlint` as well

.github/workflows/build.yml
```diff
@@ -0,0 +1,102 @@
...
+    name: build-windows
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/cache@v1
+      name: Cache cabal store
+      with:
+        path: ~/AppData/Roaming/cabal/store
```
cnr: Looks like this isn't caching on windows. I'll look into it on Monday

cnr: it's a [known issue](https://github.com/actions/cache/issues/144) that should resolve itself

src/Strategy/Gradle.hs
```diff
@@ -94,7 +94,7 @@ analyze BasicDirOpts{..} =
 
     let text = decodeUtf8 $ BL.toStrict stdout
         textLines :: [Text]
-        textLines = T.lines text
+        textLines = T.lines (T.filter (/= '\r') text)
```
zlav: We should add a test for this by duplicating the current test file and changing the endings.

cnr: I'll follow up today with another PR for normalizing output from Exec/ReadFS. There's another case we need to properly handle too -- and that's when Exec/ReadFS outputs aren't UTF8

PR #24, opened 2020-01-22
Add aeson-ish XML parser

cnr:
This adds an XML parser that gives decent error messages when XML parsing fails. It also substantially simplifies XML parsing

A future PR will add better documentation

### DIFF
--- a/src/Effect/ReadFS.hs
+++ b/src/Effect/ReadFS.hs
@@ -15,7 +15,6 @@ module Effect.ReadFS
   -- * Parsing file contents
   , readContentsParser
   , readContentsJson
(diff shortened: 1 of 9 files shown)


src/Effect/ReadFS.hs
```diff
@@ -2,19 +2,23 @@
 {-# language TemplateHaskell #-}
 
 module Effect.ReadFS
-  ( ReadFS(..)
+  ( -- * ReadFS Effect
+    ReadFS(..)
   , readFSToIO
 
+  -- * Reading raw file contents
```
zlav: I like this!

PR #30, opened 2020-02-06
Build static linux binaries (again)

cnr:
When updating to GHC 8.8, I ran into issues with building under alpine. I've since [worked around](https://github.com/fossas/haskell-static-alpine/compare/ceb27070f9e755d9115a24c9ed2cd9ff5e9044ce...master) those issues, so we can start building static linux binaries again.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,9 +20,10 @@ name: Build & Test
 on: [push]
 
 jobs:
-  build-ubuntu:
-    name: build-ubuntu
+  build-linux:
+    name: build-linux
     runs-on: ubuntu-latest
+    container: quay.io/fossa/haskell-static-alpine
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -20,9 +20,10 @@ name: Build & Test
 on: [push]
 
 jobs:
-  build-ubuntu:
-    name: build-ubuntu
+  build-linux:
+    name: build-linux
     runs-on: ubuntu-latest
```
zlav: I think this `runs-on` line may be redundant based on this documentation here https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer, although I do see where `runs-on` is a required field so it looks like they are contradicting themselves in their own docs. Either way this looks a little confusing to me and I can't tell if we're using ubuntu or alpine.

cnr: turns out this is required -- I guess you can run some steps in docker and others in the host (ubuntu in this case)

PR #34, opened 2020-02-11
Declared license detector

zlav:
This PR extends spectrometers ability by creating a new binary called `pathfinder` which can be used to detect all declared licenses in a directory. This process leverages the existing parser work that has been done to output license information in the following format which is an array of strategies. License value can be one of `expression`, `url`, `file`, or `unknown`.
```JSON
[
  {
    "name": "nuget-nuspec",
    "files": [
      {
        "filepath": "test/NuGet/testdata/test.nuspec",
        "licenses": [
          {
            "value": "https://licence.location.com/LICENSE.md",
            "type": "url"
          },
          {
            "value": "license-file",
            "type": "file"
          }
        ]
      }
    ],
    "type": "dotnet"
  }
]
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,7 +61,7 @@ jobs:
       run: |
         strip dist-newstyle/build/x86_64-linux/ghc-8.8.2/pathfinder-0.1.0.0/x/pathfinder/opt/build/pathfinder/pathfinder
 
-    - uses: actions/upload-pathfinder-artifact@v1
+    - uses: actions/upload-artifact@v1
       with:
         name: pathfinder-linux
         path: dist-newstyle/build/x86_64-linux/ghc-8.8.2/pathfinder-0.1.0.0/x/pathfinder/opt/build/pathfinder/pathfinder
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -166,7 +166,6 @@ jobs:
       id: cache-package
       with:
         path: ~/AppData/Roaming/cabal/packages
-        path: ~/.cabal/packages
         key: ${{ runner.os }}-cabal-packages
 
     - uses: actions/cache@v1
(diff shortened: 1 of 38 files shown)


src/Strategy/NuGet/Nuspec.hs
```diff
@@ -66,11 +97,17 @@ data NuGetDependency = NuGetDependency
...
+           <*> optional (child "licenseUrl" metadata)
+
+instance FromXML NuspecLicense where
+  parseElement el =
+    NuspecLicense <$> attr "type" el
+                  <*> content el
 
 instance FromXML Group where
-  parseElement el = Group <$> children "dependency" el
+  parseElement el = Group <$> optional (children "dependency" el) `defaultsTo` []
```
cnr: [nit] I think this `defaultsTo` doesn't do anything -- `children` already returns an empty list if there aren't any

.github/workflows/build.yml
```diff
@@ -113,10 +132,11 @@ jobs:
     - uses: actions/checkout@v2
 
     - uses: actions/cache@v1
-      name: Cache cabal store
+      name: Cache ~/.cabal/store
+      id: cache-store
       with:
         path: ~/AppData/Roaming/cabal/store
-        key: ${{ runner.os }}-cabal-store
+        key: ${{ hashFiles ('**\hscli.cabal') }}-${{ runner.os }}-cabal-store
```
cnr: does this have to be prefixed with `**\`? if so, that's annoying

zlav: Windows is the only one I haven't been able to get to work. This way it does find a filehash but I'm not 100% confident that it is for `hscli.cabal`. I had a lot of issues trying to get `AppData/Roaming/hsci.cabal` to work. I'm putting this on hold for now. Mac and Linux being cached are helpful enough to get the tests running faster.

PR #38, opened 2020-02-19
fork strategies

cnr:
working title: _Discovery **IS** analysis_

The old `Strategy` abstraction was broken for a number of reasons:

- strategies couldn't determine completeness at analysis time -- it was hardcoded as a field on the strategy. this caused a split between the `pipfile.lock-with-pipenv-cmd` strategy and `pipfile.lock-without-pipenv-cmd` strategy, because one has edges and the other doesn't. now they're the same strategy.
- strategy "options" had to be serializable, which was too inflexible for the pom.xml strategy (and upcoming multi-project dotnet strategy)
  - this seemed like a great answer to user configuration, but turned out to be the wrong abstraction. user configuration has been removed.
- strategy "options" made writing strategies and testing them awkward, with weird workarounds like `fileInputJson` and `runInputConst`.
  - these workarounds have been eliminated, and we're adopting a `buildGraph`-focused testing approach for the relevant unit tests. Integration tests will run under the _real_ Exec and ReadFS interpreters

In place of `Strategy`s, discover tasks now directly produce `ProjectClosure`s

This PR also moves us over to `fused-effects`. While it requires more boilerplate to use than `polysemy`, we get a bunch of benefits:

- the interop story with the rest of the haskell ecosystem is strong, and is set to get stronger: there are issues being worked on wrt, e.g., UnliftIO, that will make our lives even easier
- the amount of testing in the Real World (via `semantic`) and test coverage is huge. There's a bunch of weirdness in the semantics of polysemy that fused-effects avoids by default
- it _should_ feel more familiar to those used to monad transformers, and there's less "magic". implementation details are transparent. (no more TemplateHaskell!)
- there are fewer hacks to get asynchrony working from within our actions
- compile times are much faster
- we get a free performance boost

Other minor changes:
- [x] we log to stderr now instead of stdout
- [x] analysis results are piped to stdout by default (overridable with `--outfile` flag)
- [x] we try all files matching `req*.txt` in the requirements.txt strategy

TODO before merging:
- [x] reimplement project grouping
- [ ] associate logs with projects


src/App/Scan/GraphBuilder.hs
```diff
@@ -12,42 +11,50 @@ module App.Scan.GraphBuilder
...
 
-data GraphBuilder m a where
-  AddNode   :: Dependency -> GraphBuilder m G.DepRef
-  AddEdge   :: G.DepRef -> G.DepRef -> GraphBuilder m ()
-  AddDirect :: G.DepRef -> GraphBuilder m ()
+data GraphBuilder m k
+  = AddNode Dependency (G.DepRef -> m k)
+  | AddEdge G.DepRef G.DepRef (m k)
+  | AddDirect G.DepRef (m k)
+  deriving Generic1
```
zlav: Is `Generic1` an accident or is this something different?

PR #40, opened 2020-03-04
Include project failures in the analysis blob

cnr:
This includes failures and failure reasons in the analysis blob (for use in manual/automatic comparisons)

This also:
- fixes minor parsing issues with the PodfileLock parser
- disables `PipList` strategy (this is more "environment scanning", which we'll visit later)
- disables transitive dependency scanning for golang strategies (golang needs to be revisited)

### DIFF
--- a/test/Cocoapods/PodfileLockTest.hs
+++ b/test/Cocoapods/PodfileLockTest.hs
@@ -12,6 +12,7 @@ import DepTypes
 import Strategy.Cocoapods.PodfileLock
 import GraphUtil
 
+import qualified Test.Hspec.Megaparsec as T
 import qualified Test.Tasty.Hspec as T
 
 dependencyOne :: Dependency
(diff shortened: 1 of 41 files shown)


test/Cocoapods/PodfileLockTest.hs
```diff
@@ -68,8 +69,4 @@ spec_analyze = do
   podLockFile <- T.runIO (TIO.readFile "test/Cocoapods/testdata/Podfile.lock")
   T.describe "podfile lock parser" $
     T.it "parses error messages into an empty list" $
-      case runParser findSections "" podLockFile of
-        Left _ -> T.expectationFailure "failed to parse"
-        Right result -> do
-            result `T.shouldContain` [podSection]
-            result `T.shouldContain` [dependencySection]
+      runParser findSections "" podLockFile `T.parseSatisfies` \result -> podSection `elem` result && dependencySection `elem` result
```
zlav: I'm not sure I like this option better. We lose the failed to parse exception and it feels more difficult to understand what we're asserting on.

PR #41, opened 2020-03-06
Add `DepEnvironment`

cnr:
This formalizes environments as a concrete sum type, so we can filter out test/dev dependencies when converting to the v1 format

This also fixes a bug wrt `classifier` in the maven pom strategy

### DIFF
--- a/src/DepTypes.hs
+++ b/src/DepTypes.hs
@@ -7,12 +7,15 @@ module DepTypes
 
 import Prologue
 
+-- FIXME: this needs a smart constructor with empty tags/environments/etc.
+-- We've historically relied on the compile error for making sure we fill all
+-- of these fields. Tests should be used to ensure this instead.
 data Dependency = Dependency
   { dependencyType         :: DepType
   , dependencyName         :: Text
   , dependencyVersion      :: Maybe VerConstraint
   , dependencyLocations    :: [Text]
-  , dependencyEnvironments :: [DepEnvironment]
+  , dependencyEnvironments :: [DepEnvironment] -- FIXME: this should be a Set
   , dependencyTags         :: Map Text [Text]
   } deriving (Eq, Ord, Show, Generic)
 
--- a/src/DepTypes.hs
+++ b/src/DepTypes.hs
@@ -20,7 +20,6 @@ data DepEnvironment =
     EnvProduction
   | EnvDevelopment
   | EnvTesting
-  | EnvOther Text
   deriving (Eq, Ord, Show, Generic)
 
 -- | A Dependency type. This corresponds to a "fetcher" on the backend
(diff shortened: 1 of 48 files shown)


zlav:
I'm not sure I agree with environment becoming a sum type. We should still have the ability to filter dependencies if their only scopes are "test" or "compileTest". My biggest reservation is that I don't see really see a difference between `DepEnvironment` and `scope`, especially if we need to retain both of them. I would prefer that we either do `DepEnvironment` that can be an array of all possible scopes or we stick to tags where we don't have to make a determination.

cnr:
I definitely should've added more details:

This falls out of a short-term need to filter out non-production dependencies when converting to the v1 upload format. We need to make this determination somewhere, and doing it at the Dependency creation site leads to fewer buildtool-specific implementation details leaking downstream

> We should still have the ability to filter dependencies if their only scopes are "test" or "compileTest".

We can absolutely do this without the addition of DepEnvironments. The difficulty is that this structure of scope tags (and tags in general) isn't well-defined (as in, a change to the way we tag scopes in the maven analyzer could break downstream code that relies on a specific structure). This is an easy win for type safety, I think

This does remind me that environments aren't properly tagged for all maven scopes, so I'll fix that

> I would prefer that we either do DepEnvironment that can be an array of all possible scopes

The reason I chose production/development/test is that I consider those to be the three big buckets to categorize dependencies -- but you're probably right in that capturing "compileTest" vs "test" will be important for our future feature-set. Do you think we should model DepEnvironment based on maven's scopes? They seem pretty comprehensive (or at least a really good starting point).

(and yes environments are a list on Dependency -- but this reminds me that they should be a Set)

zlav:
Doing this to support the CLI v1 format makes a lot of sense, it would be very unintuitive for users if we included dev deps so that's a great call. I think all of your opinions here make a lot of sense but they do leave me still concerned about the scenario where we see configurations we don't anticipate. The one thing I really like about having untyped tags is that if we see a maven configuration that we haven't before we will automatically pick it up. One struggle we had with v1 was handling configurations which we hadn't yet discovered. Examples:
[CLI PR with configurations we didn't know about](https://github.com/fossas/fossa-cli/pull/411)
[Custom config "war" that we would miss](https://discuss.gradle.org/t/subproject-depends-on-configuration-from-other-subproject-but-subproject-build-order-doesnt-respect-that/19173)

I'm not sold on using dependencyEnvironments in the long run, but for the purposes of graph conversion, I have no arguments. My personal ideal solution would be to check for known test or dev dep tags while doing the conversion, but I'm not sure if that's feasible so I'm hesitant to say it is the right way forward. 

As long as we have the ability to capture tags without having to know their name then I have no issues with what we choose to do. One option might be to do the following:
```haskell
data DepEnviornment = 
    Production
    | Development
    | Test
    | Unknown Text
```
That would at least allow us to handle scopes we aren't aware of.

cnr:
Added back an `EnvOther` constructor for the gradle cases you've described. We'll definitely revisit/change this approach when we build out tagging support and have clear requirements there

PR #42, opened 2020-03-31
Googlesource xml

spatten:
Parse `manifest.xml` files read by the `repo` tool: https://gerrit.googlesource.com/git-repo/

`repo` is used to manage groups of related git repositories. A `repo` manifest can be parsed to understand what git repositories are being pulled from and what branch or tag or commit is being checked out for each of those repositories.

The format of the XML file and the information we want to extract from it is described in the "AOSP Static Analysis" RFC: https://docs.google.com/document/d/1X3nvw6_qsoDxpa4fqRtMLOQmB7pVHFPIlsfhEQNlaII/edit#heading=h.4g7h78a327rl

A manifest will consist of:

* zero or one `default` tags
* zero or more `remote` tags
* zero or more `project` tags
* zero or more `include` tags

The project tags are the things we are interested in. They, combined with information from the `remote` and `default` tags, provide a list of the dependencies. We want to go from a manifest like this:

```
<?xml version="1.0" encoding="UTF-8"?>
<manifest>
  <remote name="aosp" fetch="https://android.googlesource.com" />
  <remote fetch="https://android.othersource.com" name="othersource" revision="google/android-6.0.1_r74" />
  <default revision="refs/tags/android-10.0.0_r29"
           remote="aosp"
           sync-j="4" />

  <project path="art" name="platform/art" groups="pdk" remote="aosp"/>
  <project path="bionic" name="platform/bionic" groups="pdk" revision="57b7d1574276f5e7f895c884df29f45859da74b6" />
  <project path="bootable/recovery" name="platform/bootable/recovery" groups="pdk" remote="othersource" />
  <project path="cts" name="platform/cts" groups="cts,pdk-cw-fs,pdk-fs" remote="othersource" revision="1111"/>
  <project path="dalvik" name="platform/dalvik" groups="pdk-cw-fs,pdk-fs" />
</manifest>
```

To a list of dependencies that (in JSON format) would look like this:

```
[
   {  "name": "platform/art",
       "version": {
         "value": "refs/tags/android-10.0.0_r29",
         "type": "EQUAL"
       },
       "type": "GooglesourceType",
       "locations": [
         "https://android.googlesource.com/platform/art"
       ],
       "tags": {}
    },
...
   {  "name": "platform/dalvik",
       "version": {
         "value": "refs/tags/android-10.0.0_r29",
         "type": "EQUAL"
       },
       "type": "GooglesourceType",
       "locations": [
         "https://android.googlesource.com/platform/dalvik"
       ],
       "tags": {}
    }
]
```

The manifest can also just consist of an `include` tag which points at another manifest to be parsed. In this example, the manifest is pointing at a file called `default.xml`, which will be found at `manifests/default.xml`, where the `manifests` directory is a sibling of the original file being read.

```
<manifest>
  <include name="default.xml" />
</manifest>
```


src/Strategy/Googlesource/RepoManifest.hs
```diff
@@ -0,0 +1,218 @@
...
+  case find (\f -> (==) "manifest.xml" (fileName f)) files of
+    Nothing -> walkContinue
+    Just file -> do
+      if (dirname $ parent file) == $(mkRelDir ".repo") then do
+        runSimpleStrategy "googlesource-repomanifest" GooglesourceGroup $ analyze file
+        walkSkipAll subdirs
+      else walkContinue
+
+analyze :: (Has ReadFS sig m, Has (Error ReadFSErr) sig m, MonadFail m) => Path Rel File -> m ProjectClosureBody
+analyze file = do
```
cnr, line 45: [style] another way to write this is
```hs
analyze file = mkProjectClosure file <$> nestedValidatedProjects file
```

Anytime you see:
```hs
something = do
  bar <- doSomething
  pure (foo bar)
```
it's the same as
```hs
something = foo <$> doSomething
```

src/Strategy/Googlesource/RepoManifest.hs
```diff
@@ -0,0 +1,218 @@
...
+-- </manifest>
+-- If you see that, you need to look for `manifests/default.xml`, where the manifests directory will
+-- be a sibling to the original manifest you were parsing.
+validatedProjectsFromIncludes :: (Has ReadFS sig m, Has (Error ReadFSErr) sig m, MonadFail m) => RepoManifest -> Path Rel Dir -> m [ValidatedProject]
+validatedProjectsFromIncludes manifest parentDir = do
+    let dirPath = toFilePath parentDir
+        manifestIncludeFiles :: [Text]
+        manifestIncludeFiles = map includeName $ manifestIncludes manifest
+        manifestFiles :: Maybe [Path Rel File]
+        manifestFiles = traverse (\file -> parseRelFile (dirPath ++ "manifests/" ++ T.unpack file)) manifestIncludeFiles
```
cnr, line 72: can `manifestIncludeFiles` contain paths with `..`? `parseRelFile` doesn't like those.

you may actually want to use `resolveFile` from the `ReadFS` effect here anyway:
```hs
resolveFile :: (Has ReadFS sig m, Has (Error ReadFSErr) sig m) => Path Abs Dir -> Text -> m (Path Abs File)
```

You can do a similar `traverse`, and bind (`<-`) out the result

---

an alternative (though I generally prefer the option above) is to keep `parentDir` as a `Path Rel Dir`, and do something like this:
```hs
manifestFiles = traverse (\file -> (parentDir </>) <$> parseRelFile (T.unpack file)) manifestIncludeFiles
```

but maybe written a little more verbosely for readability

src/Strategy/Googlesource/RepoManifest.hs
```diff
@@ -0,0 +1,218 @@
...
+--   * If the project does not have a revision and there is no default revision from either its remote or the default, then blow up
+revisionForProject :: RepoManifest -> ManifestProject -> Maybe Text
+revisionForProject manifest project =
+      projectRevision project
+  <|> (remoteForProject manifest project >>= remoteRevision)
+  <|> (manifestDefault manifest >>= defaultRevision)
+
+-- The URL for a project is the project's name appended to the fetch attribute of the project's remote
+urlForProject :: RepoManifest -> ManifestProject -> Maybe Text
+urlForProject manifest project =
```
cnr, line 149: [style] another way to write this (with `Maybe` do notation)
```hs
urlForProject manifest project = do
  remote <- remoteForProject manifest project
  pure $ remoteFetch r <> "/" <> projectName project
```

src/Strategy/Googlesource/RepoManifest.hs
```diff
@@ -0,0 +1,218 @@
...
+remoteForProject manifest project =
+  remoteNameString >>= remoteByName manifest
+  where
+    remoteNameString = projectRemote project <|> (manifestDefault manifest >>= defaultRemote)
+
+remoteByName :: RepoManifest -> Text -> Maybe ManifestRemote
+remoteByName manifest remoteNameString =
+  find (\r -> remoteName r == remoteNameString) (manifestRemotes manifest)
+
+validatedProjects :: RepoManifest -> Maybe [ValidatedProject]
```
cnr, line 166: [nit] naming: I'd prefer something like `validateProjects` -- `validatedProjects` sounds like it might be a field on some record named `ValidatedFooBar`

(same with `validatedProject` below)

src/Strategy/Googlesource/RepoManifest.hs
```diff
@@ -0,0 +1,218 @@
...
+validatedProjects manifest =
+    traverse (validatedProject manifest) (manifestProjects manifest)
+
+validatedProject :: RepoManifest -> ManifestProject -> Maybe ValidatedProject
+validatedProject manifest project =
```
cnr, line 171: [style] another way to write this with do notation:
```hs
validatedProject manifest project = do
  revision <- revisionForProject manifest project
  url <- urlForProject manifest project
  pure $ ValidatedProject (projectName project) (projectPathOrName project) url revision

File: src/Strategy/Googlesource/RepoManifest.hs (hunk 2 above)
Comment (spatten) [reply src/Strategy/Googlesource/RepoManifest.hs:72]: `manifestIncludeFiles` cannot contain paths with `..`. As far as I can tell from the code (it's not really specified anywhere), it can only contain a filename without a directory -- the file it's pointing at must be in the `manifests` directory, and the `manifests` directory must be a sibling to the current `manifest.xml` that we're parsing.

If that's true, does that change your recommendation? I think I've only been dealing with `Rel` files, not `Abs` files, so I'm not sure if I can just drop in `resolveFile`.

File: src/Strategy/Googlesource/RepoManifest.hs (hunk 5 above)
Comment (spatten) [reply src/Strategy/Googlesource/RepoManifest.hs]: So much nicer. Thanks!

File: src/Strategy/Googlesource/RepoManifest.hs (hunk 4 above)
Comment (spatten) [reply src/Strategy/Googlesource/RepoManifest.hs]: Done!

File: src/Strategy/Googlesource/RepoManifest.hs (hunk 3 above)
Comment (spatten) [reply src/Strategy/Googlesource/RepoManifest.hs]: Again, so much nicer :)

File: src/Strategy/Googlesource/RepoManifest.hs (hunk 1 above)
Comment (spatten) [reply src/Strategy/Googlesource/RepoManifest.hs]: Nice!

File: src/Strategy/Googlesource/RepoManifest.hs [hunk 6]
```diff
@@ -0,0 +1,218 @@
...
+import Discovery.Walk
+import Effect.ReadFS
+import Graphing (Graphing, unfold)
+import Parse.XML
+import Types
+
+-- We're looking for a file called "manifest.xml" in a directory called ".repo"
+discover :: HasDiscover sig m => Path Abs Dir -> m ()
+discover = walk $ \_ subdirs files -> do
+  case find (\f -> (==) "manifest.xml" (fileName f)) files of
```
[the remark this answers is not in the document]
spatten: Yeah, not sure what I was thinking there. Thanks!

src/Strategy/Googlesource/RepoManifest.hs
cnr: Yeah that definitely does change my recommendation -- something like the second option would be preferred in that case

spatten: :+1:

spatten: Okay, done. This looks much better too :)

PR #45, opened 2020-04-02
Translate graphs to SourceUnits and upload to the CLIv1 endpoint

cnr:
This adds a new `--upload APIKEY` argument to `scan`. When present, we upload results to the v1 API endpoint

- [x] convert `ProjectClosure` to `SourceUnit`
- [x] add cli command for running analysis and uploading to the old endpoint

### DIFF
--- a/src/App/Scan/FossaV1.hs
+++ b/src/App/Scan/FossaV1.hs
@@ -1,11 +1,16 @@
 module App.Scan.FossaV1
   ( uploadAnalysis
+  , UploadResponse(..)
+  , FossaError(..)
   ) where
 
+import Control.Carrier.Error.Either
 import Data.List (isInfixOf)
-import Data.Text.Encoding (decodeUtf8, encodeUtf8)
+import Data.Text.Encoding (encodeUtf8)
 import Effect.Logger
+import qualified Network.HTTP.Client as HTTP
 import Network.HTTP.Req
+import qualified Network.HTTP.Types as HTTP
 import Prologue
 import Srclib.Converter (toSourceUnit)
 import Srclib.Types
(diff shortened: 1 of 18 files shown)


src/App/Scan/FossaV1.hs
```diff
@@ -0,0 +1,98 @@
...
+cliVersion = "spectrometer"
+
+uploadUrl :: Url 'Https
+uploadUrl = https "app.fossa.com" /: "api" /: "builds" /: "custom"
+
+newtype FossaReq m a = FossaReq { runFossaReq :: ErrorC FossaError m a }
+  deriving (Functor, Applicative, Monad, MonadIO)
+
+instance (Algebra sig m, Effect sig, MonadIO m) => MonadHttp (FossaReq m) where
+  handleHttpException = FossaReq . throwError . mangleError
```
cnr, line 30: This is disgusting, but unfortunately the best option for now. Normally, `Req` throws exceptions in IO (i.e., the type doesn't reflect the kinds of exceptions it can throw) -- but we'd like to be explicit and return out `Either FossaError UploadResponse`.

This bit of confusing boilerplate re-throws the exceptions with the `Error FossaError` effect

src/Graphing.hs
```diff
@@ -1,17 +1,32 @@
+-- | A graph augmented with a set of "direct" vertices
```
cnr, line 1: Additionally added some documentation to `Graphing`

src/Graphing.hs
```diff
@@ -68,3 +90,16 @@ unfold seed getDeps toDependency = foldr addNode empty seed
     where
     children = getDeps dep
     res = toDependency dep
+
+-- | Build a graphing from a list, where all list elements are treated as direct
+-- dependencies
+fromList :: Ord ty => [ty] -> Graphing ty
```
cnr, line 96: This is a new convenience method for building Graphings. Rather than using `unfold` with `const []` as an argument:
```hs
buildGraph nodes = Graphing.unfold nodes (const []) toDependency
```
we instead can do:
```hs
buildGraph nodes = Graphing.fromList (map toDependency nodes)
```

or sometimes point-free:
```hs
buildGraph = Graphing.fromList . map toDependency
```

---

I updated the previous callsites for `unfold` to use `fromList` instead, wherever possible

src/Srclib/Converter.hs
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+verConstraintToRevision :: VerConstraint -> Maybe Text
+verConstraintToRevision = \case
+  CEq ver -> Just ver
+  CURI _ -> Nothing -- we can't represent this in a locator
+  CCompatible ver -> Just ver
+  CAnd a b -> verConstraintToRevision a <|> verConstraintToRevision b
+  COr a b -> verConstraintToRevision a <|> verConstraintToRevision b
+  CLess ver -> Just ver -- ugh
```
cnr, line 80: A resolution backend for cliv2 can't come soon enough

src/Srclib/Converter.hs
```diff
@@ -0,0 +1,96 @@
...
+  CLessOrEq ver -> Just ver -- ugh
+  CGreater ver -> Just ver -- ugh
+  CGreaterOrEq ver -> Just ver -- ugh
+  CNot _ -> Nothing -- we can't represent this in a locator
+
+depTypeToFetcher :: DepType -> Text
+depTypeToFetcher = \case
+  SubprojectType -> "mvn" -- FIXME. I knew SubprojectType would come back to bite us.
```
cnr, line 88: I don't know what to do about this yet. Should we model project relationships somehow? We'd probably strip that information on the client side, but at least we wouldn't have to have hacks like this.

src/Strategy/Python/PipList.hs
```diff
@@ -51,7 +52,7 @@ mkProjectClosure dir deps = ProjectClosureBody
     }
 
 buildGraph :: [PipListDep] -> Graphing Dependency
-buildGraph xs = unfold xs (const []) toDependency
+buildGraph = Graphing.fromList . map toDependency
```
cnr, line 55: This reads so much better than the old version

PR #46, opened 2020-04-10
Simplify and document filetree walking

cnr:
This adds a thin wrapper around `path-io`'s `walkDirRel`

`walkDirRel` isn't great because the subdirs and files passed into the walk function are relative to _the current directory_, not the base directory of the filetree traversal.

Additionally, having our own interface opens up potential fusing of filetree traversals / moving to other filetree traversal backends that provide fusing.

### DIFF
--- a/src/Strategy/Carthage.hs
+++ b/src/Strategy/Carthage.hs
@@ -26,12 +26,12 @@ import qualified Graphing as G
 import Types
 
 discover :: HasDiscover sig m => Path Abs Dir -> m ()
-discover = walk $ \_ subdirs files ->
+discover = walk $ \_ _ files ->
   case find (\f -> fileName f == "Cartfile.resolved") files of
-    Nothing -> walkContinue
+    Nothing -> pure WalkContinue
     Just file -> do
       runSimpleStrategy "carthage-lock" CarthageGroup $ fmap (mkProjectClosure file) (analyze file)
-      walkSkipAll subdirs
+      pure WalkSkipAll
 
 mkProjectClosure :: Path Rel File -> G.Graphing ResolvedEntry -> ProjectClosureBody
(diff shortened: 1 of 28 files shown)


src/Strategy/Carthage.hs
```diff
@@ -26,12 +26,12 @@ import qualified Graphing as G
...
 discover :: HasDiscover sig m => Path Abs Dir -> m ()
-discover = walk $ \_ subdirs files ->
+discover = walk $ \_ _ files ->
   case find (\f -> fileName f == "Cartfile.resolved") files of
-    Nothing -> walkContinue
+    Nothing -> pure WalkContinue
     Just file -> do
       runSimpleStrategy "carthage-lock" CarthageGroup $ fmap (mkProjectClosure file) (analyze file)
-      walkSkipAll subdirs
+      pure WalkSkipAll
```
cnr, line 34: This is much clearer than the previous construction

src/Strategy/Go/GoList.hs
```diff
@@ -22,12 +24,12 @@ import Strategy.Go.Types
...
 
 discover :: HasDiscover sig m => Path Abs Dir -> m ()
-discover = walk $ \_ subdirs files -> do
+discover = walk $ \_ _ files -> do
   case find (\f -> fileName f == "go.mod") files of
     Nothing -> pure ()
     Just file  -> runSimpleStrategy "golang-golist" GolangGroup $ analyze (parent file)
 
-  walkSkipNamed ["vendor/"] subdirs
+  pure $ WalkSkipSome [$(mkRelDir "vendor")]
```
cnr, line 32: This is both clearer _and_ safer -- the old `walkSkipNamed` had some dangerous caveats

PR #47, opened 2020-04-11
Sherlock cli

spatten:


hscli.cabal
```diff
@@ -7,8 +7,73 @@ extra-source-files:
   scripts/*.jar
   scripts/*.gradle
 
-common lang
-  build-depends:      base >=4.12 && <4.14
+library
```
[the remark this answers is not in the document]
cnr: weird. I don't think we intentionally did this -- it was probably something with `cabal-fmt`. I'll fix this up

src/App/Scan.hs
```diff
@@ -2,71 +2,77 @@
...
-import Types
+import Network.HTTP.Req (HttpException)
+
+import VPSScan.RunSherlock
+import VPSScan.ScotlandYard
+import VPSScan.RunIPR
 
 data ScanCmdOpts = ScanCmdOpts
   { cmdBasedir :: FilePath
   , cmdDebug   :: Bool
```
cnr, line 22: we should kill these debug/outfile switches too

src/App/Scan.hs
```diff
@@ -2,71 +2,77 @@
...
+  ( Has ScotlandYard sig m
+  , Has IPR sig m
+  , Has Sherlock sig m
+  , Has (Error VPSError) sig m
+  ) => Path Abs Dir -> ScanCmdOpts -> m ()
+vpsScan basedir ScanCmdOpts{..} = do
+  let VPSOpts{..} = scanVpsOpts
+  response <- tagError Couldn'tGetScanId =<< createScotlandYardScan vpsScotlandYard
+ 
+  traceM $ "Running scan on directory " ++ show basedir
```
cnr, line 62: rather than using `traceM` here, let's use the `Trace` effect, which gives us:
```hs
trace :: Has Trace sig m => String -> m ()
```

it has a few carriers/interpreters, exported from these three modules:
- `Control.Carrier.Trace.Printing` - this mirrors the behavior of `traceM`
- `Control.Carrier.Trace.Ignoring` - this ignores the traces
- `Control.Carrier.Trace.Returning` - this collects the trace messages as a list and returns them out of the computation (most often useful for testing)

you'll have to use `runTrace` (imported from one of the carrier modules) at the callsite of vpsScan

src/VPSScan/RunSherlock.hs
```diff
@@ -0,0 +1,66 @@
...
+import qualified Data.Text as T
+
+data SherlockOpts = SherlockOpts
+  { sherlockCmdPath :: String
+  , sherlockUrl :: String
+  , sherlockClientToken :: String
+  , sherlockClientID :: String
+  } deriving (Eq, Ord, Show, Generic)
+
+scan :: (Has Exec sig m, Has (Error ExecErr) sig m) => Path Abs Dir -> Text -> SherlockOpts -> m ()
```
cnr, line 22: I think this is no longer used

src/App/Scan.hs
spatten: I knew using `traceM` in production code was a bad idea. Meant to ask you about that. Thanks!

spatten:
Closing this in favour of #50, which does the same thing but also creates a new `vpscli` binary.

PR #49, opened 2020-04-21
fossa commandline UI

cnr:
This PR:
- adds a new command, `fossa analyze`. This supports most of the same options as the ones found in cliv1
- adds a new command -- `fossa test`
- removes the old `hscli` binary

TODO:
- [x] update install scripts to install `fossa2` instead of `hscli`
- [x] update cliv1 to run `fossa2` instead of `hscli` for comparison data

Coming in a future PR:
- uniform error handling/rendering for API errors


src/App/Fossa/Analyze.hs
```diff
@@ -53,77 +52,63 @@ import qualified Strategy.Ruby.BundleShow as BundleShow
...
   , Has Logger sig m
   , Has Threaded sig m
   , MonadIO m
   , Effect sig
   )
   => Path Abs Dir
-  -> Maybe FilePath
-  -> Maybe Text -- ^ API key for fossa core. when present, we'll upload results
+  -> ScanDestination
+  -> Maybe Text -- ^ cli override for name
```
cnr, line 74: This is gross: we should pass in the exact `ProjectRevision` here -- will require a future refactor to move project inference outside of analyze/test.

currently blocked on a usable `Async` implementation for fused-effects

src/App/Fossa/Analyze/FossaV1.hs
```diff
@@ -0,0 +1,307 @@
...
+          <> "v" =: cliVersion
+          <> "managedBuild" =: True
+          <> "title" =: fromMaybe projectName (projectTitle metadata)
+          <> header "Authorization" ("token " <> encodeUtf8 key)
+          <> mkMetadataOpts metadata
+  resp <- req POST (uploadUrl baseurl) (ReqBodyJson sourceUnits) jsonResponse opts
+  pure (responseBody resp)
+
+mkMetadataOpts :: ProjectMetadata -> Option scheme
+mkMetadataOpts ProjectMetadata{..} = mconcat $ catMaybes $
```
cnr, line 112: this is clever, but also probably the best way to do this(?)

we have a bunch of optional flags in ProjectMetadata that were passed from the command line, and we need to turn them into query options:

some intermediate types here:
`projectBranch :: Maybe Text`
fmapping in the query option constructor gives us:
`("branch" =:) <$> projectBranch :: Maybe (Option scheme)`
we can use catMaybes to filter out the `Nothing`s:
`catMaybes :: [Maybe (Option scheme)] -> [Option scheme]`
and finally we can use mconcat to combine the query param Options with `<>`
`mconcat :: [Option scheme] -> Option scheme`

src/App/Fossa/Analyze/FossaV1.hs
```diff
@@ -0,0 +1,307 @@
...
+      HTTP.Status 404 _ -> InvalidProjectOrRevision err
+      HTTP.Status 403 _ -> NoPermission err
+      _                 -> OtherError err
+  JsonHttpException msg -> JsonDeserializeError msg
+  _ -> OtherError err
+
+-- we specifically want Rel paths here: parent directories shouldn't affect path
+-- filtering
+isProductionPath :: Path Rel fd -> Bool
+isProductionPath path = not $ any (`isInfixOf` toFilePath path)
```
cnr, line 134: this emulates the exact filtering logic as found in cliv1

src/App/Fossa/Main.hs
```diff
@@ -0,0 +1,130 @@
...
+
+testOpts :: Parser TestOptions
+testOpts =
+  TestOptions
+    <$> option auto (long "timeout" <> help "Duration to wait for build completion (in seconds)" <> value 600)
+
+data CmdOptions = CmdOptions
+  { optDebug   :: Bool
+  , optBasedir :: FilePath
+  , optBaseUrl :: Url 'Https
```
cnr, line 113: Note that this field can also contain a non-https url -- the Req library is just annoyingly picky about this. We coerce any url we parse into this type, despite the underlying value being http

src/App/Fossa/Test.hs
```diff
@@ -0,0 +1,174 @@
...
+  -> Text -- ^ api key
+  -> Severity
+  -> Int -- ^ timeout (seconds)
+  -> Maybe Text -- ^ cli override for name
+  -> Maybe Text -- ^ cli override for revision
+  -> IO ()
+testMain baseurl apiKey logSeverity timeoutSeconds overrideName overrideRevision= do
+  basedir <- getCurrentDir
+
+  void $ timeout timeoutSeconds $ withLogger logSeverity $ do
```
cnr, line 37: This magic timeout is very cool, but unfortunately forces us to directly use IO here. The existing fused-effects machinery doesn't allow us to express timeout.

As soon as it does, the code in Fossa.Main, Fossa.Analyze, and Fossa.Test gets substantially simpler

src/App/Fossa/Test.hs
```diff
@@ -0,0 +1,174 @@
...
+        liftIO exitFailure
+      Right _ -> liftIO exitSuccess
+
+  -- we call exitSuccess/exitFailure in each branch above. the only way we get
+  -- here is if we time out
+  hPutStrLn stderr "Timed out while waiting for issues scan"
+  exitFailure
+
+renderedIssues :: Fossa.Issues -> Doc ann
+renderedIssues issues = rendered
```
cnr, line 74: cliv1 used a template renderer here -- and we don't have something similar for haskell. Instead, we're using the `prettyprinter` library

src/App/Fossa/Test.hs
```diff
@@ -0,0 +1,174 @@
...
+        "WAITING" -> do
+          liftIO $ threadDelay (pollDelaySeconds * 1_000_000)
+          waitForIssues baseurl key revision
+        _ -> pure issues
+
+timeout
+  :: Int -- ^ number of seconds before timeout
+  -> IO a
+  -> IO (Maybe a)
+timeout seconds act = either id id <$> Async.race (Just <$> act) (threadDelay (seconds * 1_000_000) *> pure Nothing)
```
cnr, line 174: `Control.Concurrent.Async` gives us a `race` primitive (run two actions, returning the value from whichever finishes first), which allows us to pretty cleanly write a timeout function.

src/Effect/Logger.hs
```diff
@@ -78,21 +80,21 @@ logWarn = log SevWarn
...
+      flushLogger tid = sendIO $ do
+        atomically $ closeTMQueue queue
+        void (wait tid)
 
-  liftIO $ do
-    atomically $ closeTMQueue queue
-    void (wait tid)
-
-  pure res
+  bracket mkLogger
```
cnr, line 95: This now ensures that the log queue is flushed before exiting, even in the face of exceptions

PR #52, opened 2020-05-01
Integrate properly with the Scotland Yard uploadDiscoveredLicenses endpoint

spatten:
The UploadDiscoveredLicenses endpoint was expecting a slightly different JSON object

Instead of

```
[list of files]
```

We should send up

```
{"Files": [list of files]}
```

### DIFF
--- a/src/App/VPSScan/Scan/ScotlandYard.hs
+++ b/src/App/VPSScan/Scan/ScotlandYard.hs
@@ -55,14 +55,14 @@ createScan VPSOpts{..} = do
 postIprResults :: ToJSON a => VPSOpts -> Text -> a -> HTTP ()
 postIprResults VPSOpts{..} scanId value = do
   let ScotlandYardOpts{..} = vpsScotlandYard
-  _ <- req POST (scanDataEndpoint scotlandYardUrl projectID scanId) (ReqBodyJson value) ignoreResponse $ port scotlandYardPort
+  _ <- req POST (scanDataEndpoint scotlandYardUrl projectID scanId) (ReqBodyJson value) ignoreResponse $ (port scotlandYardPort <> header "Content-Type" "application/json")
   pure ()
 
 ----- scotland yard effect
 
 data ScotlandYard m k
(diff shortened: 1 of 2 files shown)


src/App/VPSScan/Scan/ScotlandYard.hs
```diff
@@ -55,14 +55,14 @@ createScan VPSOpts{..} = do
 postIprResults :: ToJSON a => VPSOpts -> Text -> a -> HTTP ()
 postIprResults VPSOpts{..} scanId value = do
   let ScotlandYardOpts{..} = vpsScotlandYard
-  _ <- req POST (scanDataEndpoint scotlandYardUrl projectID scanId) (ReqBodyJson value) ignoreResponse $ port scotlandYardPort
+  _ <- req POST (scanDataEndpoint scotlandYardUrl projectID scanId) (ReqBodyJson value) ignoreResponse $ (port scotlandYardPort <> header "Content-Type" "application/json")
```
cnr, line 58: *grumble*

src/App/VPSScan/Scan/ScotlandYard.hs
```diff
@@ -71,7 +71,7 @@ instance Effect ScotlandYard
 createScotlandYardScan :: Has ScotlandYard sig m => VPSOpts -> m (Either HttpException ScanResponse)
 createScotlandYardScan vpsOpts = send (CreateScotlandYardScan vpsOpts pure)
 
-uploadIPRResults :: Has ScotlandYard sig m => VPSOpts -> Text -> Array -> m (Either HttpException ())
+uploadIPRResults :: Has ScotlandYard sig m => VPSOpts -> Text -> Value -> m (Either HttpException ())
```
spatten, line 74: @cnr Would it be better if the signature of this was more general, like this?

```
uploadIPRResults :: Has ScotlandYard sig m, ToJSON a => VPSOpts -> Text -> a -> m (Either HttpException ())
```

src/App/VPSScan/Scan/ScotlandYard.hs
spatten: Agreed :D

PR #53, opened 2020-05-01
Extend the urlOption option parser to pass along the port and other options

spatten:
Before, if you passed in a URL like `http://localhost:8080` as a command line arg, the port was ignored. Now it is not.

This required changing the `urlOption` option parser to return both the URL and the URL options, so a new data type called `UrlOption` was used for this.

One unfortunate side-effect of this is that `UrlOption` has a `Network.HTTP.Req.Option` as one of its members, and `Option` does not implement the `Eq`, `Ord` or `Show` classes. This flows through to a bunch of the data types. It doesn't seem to cause a problem here, and I'm not sure how I'd fix it if it was.

### DIFF
--- a/src/App/Fossa/Main.hs
+++ b/src/App/Fossa/Main.hs
@@ -15,9 +15,7 @@ import System.Environment (lookupEnv)
 import System.Exit (die)
 import qualified Data.Text as T
 import OptionExtensions
-import Control.Monad.Except
-import Text.URI
-import Network.HTTP.Req (useURI, port)
+import Network.HTTP.Req (https)
 
 appMain :: IO ()
 appMain = do
@@ -27,7 +25,7 @@ appMain = do
   maybeApiKey <- fmap T.pack <$> lookupEnv "FOSSA_API_KEY"
 
   basedir <- validateDir optBasedir
- 
+
   setCurrentDir basedir
 
   case optCommand of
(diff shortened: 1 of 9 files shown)


src/App/Fossa/Main.hs
```diff
@@ -56,11 +56,14 @@ opts =
...
     <*> strOption (long "basedir" <> short 'd' <> metavar "DIR" <> help "Set the base directory for scanning (default: current directory)" <> value ".")
-    <*> urlOption (long "endpoint" <> metavar "URL" <> help "The FOSSA API server base URL" <> value (https "app.fossa.com"))
+    <*> urlOption (long "endpoint" <> metavar "URL" <> help "The FOSSA API server base URL" <> value urlOpts)
     <*> optional (strOption (long "project" <> help "this repository's URL or VCS endpoint (default: VCS remote 'origin')"))
     <*> optional (strOption (long "revision" <> help "this repository's current revision hash (default: VCS hash HEAD)"))
     <*> comm
     <**> helper
+    where
+      baseUrl = https "app.fossa.com"
+      urlOpts = UrlOption baseUrl $ mempty
```
cnr, line 66: I'd recommend running hlint over this -- e.g., this `$` is redundant

spatten: Oops. Yeah, I missed that one. Thanks!

BTW, fun alias:

```
alias hlintdiff="git diff --numstat master... | awk '{print \$3}' | xargs hlint --color=always | less"
```

PR #54, opened 2020-05-04
sbt strategy

cnr:
The sbt strategy leverages our existing multi-project maven-pom machinery, in conjunction with sbt's pom exporter (`sbt makePom`)

`sbt makePom` has one caveat: the pom files it exports _do not contain multi-project references_ via `<parent>` or `<module>` declarations. The only non-trivial logic in this strategy is introducing those references as appropriate.


src/Strategy/Scala.hs
```diff
@@ -0,0 +1,112 @@
...
+
+  case pomLocations of
+    Nothing -> throwError (CommandParseError "sbt makePom" "Could not parse pom locations")
+    Just [] -> throwError (CommandParseError "sbt makePom" "No projects found")
+    Just paths -> do
+      globalClosure <- buildGlobalClosure paths
+
+      -- The pom files generated by sbt do not include the proper <parent> references.
+      -- We need to introduce these edges ourselves.
+      let pomEdges :: AM.AdjacencyMap MavenCoordinate
```
cnr, line 91: This is the non-trivial bit. Everything else is shared with the maven-pom code. I think focusing on an integration test later will be more worthwhile than building out a unit test now

src/Strategy/Scala.hs
```diff
@@ -0,0 +1,112 @@
...
+  stdoutBL <- execThrow (parent file) makePomCmd []
+
+  -- stdout for "sbt makePom" looks something like:
+  --
+  -- > ...
+  -- > [info] Wrote /absolute/path/to/pom.xml
+  -- > [info] Wrote /absolute/path/to/other/pom.xml
+  -- > ...
+  let stdoutLText = decodeUtf8 stdoutBL
+      stdout = TL.toStrict stdoutLText
```
zlav, line 72: Why not `stdout = TL.toStrict $ decodeUtf8 stdoutBL`?

src/Strategy/Scala.hs
zlav: To clarify here, is what you're saying that this bit is already unit tested because maven pom code is unit tested?

PR #58, opened 2020-05-06
Re-enable the Googlesource.RepoManifest strategy

spatten:
I'm not sure when this was turned off, but we'll eventually need to turn it back on for Motorola.


spatten:
@cnr if this was turned off on purpose because we didn't want to ship this quite yet, just ignore the PR for now.

cnr:
Definitely wasn't intentionally disabled 👍

PR #62, opened 2020-05-07
Rust analyzer

scruffystuffs:
Rust analyzer

check https://github.com/fossas/spectrometer/blob/a1b57a99721af76261d6165456cd84580d3ff34c/docs/strategies/rust.md for details on the abstract analysis process.

Please note that `Strategy.Cargo (Package, PackageDependency)` are added for future use, when determining licenses.  Since we don't hold that info yet, we are not using that code.

### DIFF
--- a/src/Strategy/Cargo.hs
+++ b/src/Strategy/Cargo.hs
@@ -21,7 +21,7 @@ import qualified Data.Map.Strict as M
 import Discovery.Walk
 import Effect.Exec
 import Effect.Grapher
-import Graphing (Graphing)
+import Graphing (Graphing, stripRoot)
 import Types
 
 import qualified Data.Text as T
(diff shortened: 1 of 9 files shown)


scruffystuffs:
Question: should the top level analyzed project be considered a direct dependency?  Current iteration is doing this, but that doesn't make sense to me.  I'll confirm before merging.

scruffystuffs:
Pretty-printed output from a rust project of mine, found at https://github.com/scruffystuffs/jfmt.rs
Uploaded as text because github doesn't allow `.json` files (too scary).
[jfmt.json.txt](https://github.com/fossas/spectrometer/files/4618790/jfmt.json.txt)

scruffystuffs:
> Question: should the top level analyzed project be considered a direct dependency? Current iteration is doing this, but that doesn't make sense to me. I'll confirm before merging.

Btw, the answer was yes, which is why #65 was created and merged.

src/Strategy/Cargo.hs
```diff
@@ -0,0 +1,207 @@
+module Strategy.Cargo
+  ( discover
+  --
+  , CargoMetadata(..)
+  , NodeDependency(..)
+  , NodeDepKind(..)
+  -- , Package(..)
```
cnr, line 7: [nit] we should kill the comments here and elsewhere

PR #63, opened 2020-05-07
Add logStdout to Effect.Logger and use it for analyze --output

cnr:
Previously, the json output from `fossa analyze --output` was getting truncated


scruffystuffs:
giphub

PR #65, opened 2020-05-12
Add stripRoot to Graphing module

scruffystuffs:
Expected semantics:
Given some `Graphing a`, remove all items from direct set, and promote the direct descendants of that direct set to be the new direct set.

This is useful when a project is built as the top layer, and should not be included in the dependency graph.

### DIFF
--- a/src/Graphing.hs
+++ b/src/Graphing.hs
@@ -16,6 +16,7 @@ module Graphing
   , gmap
   , filter
   , pruneUnreachable
+  , stripRoot
 
   -- * Building simple Graphings
   , fromList
@@ -60,6 +61,14 @@ filter f gr = gr { graphingDirect = direct', graphingAdjacent = adjacent' }
 -- | The empty Graphing
 empty :: Graphing ty
 empty = Graphing S.empty AM.empty
+
+-- | Strip a given item from the direct set, promote its immediate children to direct items
+stripRoot :: Ord ty => Graphing ty -> Graphing ty
+stripRoot gr = gr { graphingDirect = direct' }
+  where
+  topLayer = graphingDirect gr
+  newDirects = S.unions $ map (flip AM.postSet $ graphingAdjacent gr) $ S.toList topLayer
+  direct' = foldr S.insert S.empty newDirects
 
 -- | Add a direct dependency to this Graphing
 direct :: Ord ty => ty -> Graphing ty -> Graphing ty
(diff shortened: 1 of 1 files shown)


src/Graphing.hs
```diff
@@ -61,6 +62,14 @@ filter f gr = gr { graphingDirect = direct', graphingAdjacent = adjacent' }
 empty :: Graphing ty
 empty = Graphing S.empty AM.empty
 
+-- | Strip a given item from the direct set, promote its immediate children to direct items
```
cnr, line 65: doc here is silghtly confusing -- there isn't really a 'given item'  here, and it'd apply to all direct dependencies

src/Graphing.hs
```diff
@@ -61,6 +62,14 @@ filter f gr = gr { graphingDirect = direct', graphingAdjacent = adjacent' }
...
+stripRoot :: Ord ty => Graphing ty -> Graphing ty
+stripRoot gr = gr { graphingDirect = direct' }
+  where
+  topLayer = graphingDirect gr
+  newDirects = S.unions $ map (flip AM.postSet $ graphingAdjacent gr) $ S.toList topLayer
+  direct' = foldr S.insert S.empty newDirects
```
cnr, line 71: I think `direct'` is going to have the same value as `newDirects` -- `newDirects` is already `:: Set ty`. good intuition on the foldr here in general, though 🙂

src/Graphing.hs
```diff
@@ -61,6 +62,14 @@ filter f gr = gr { graphingDirect = direct', graphingAdjacent = adjacent' }
 empty :: Graphing ty
 empty = Graphing S.empty AM.empty
 
+-- | Strip a given item from the direct set, promote its immediate children to direct items
+stripRoot :: Ord ty => Graphing ty -> Graphing ty
+stripRoot gr = gr { graphingDirect = direct' }
+  where
+  topLayer = graphingDirect gr
+  newDirects = S.unions $ map (flip AM.postSet $ graphingAdjacent gr) $ S.toList topLayer
```
cnr, line 70: [style] another way to write `(flip AM.postSet $ graphingAdjacent gr)` is to use postSet infix:
```
(`AM.postSet` graphingAdjacent gr)
```

scruffystuffs: That hurts my brain.  Infix makes sense to me, but infix as a pipeline inside of parens is really hard for me to put together mentally without more experience seeing it.

I'm going to leave this one as-is.

src/Graphing.hs
scruffystuffs: Ah yeah.  That came from my first attempt, which was based on removing a named item from `directs`.

src/Graphing.hs
scruffystuffs: This came from my first attempt, I'll rewrite it.

PR #67, opened 2020-05-14
Decide to upload based on API key existence

scruffystuffs:
Short version:  If api key exists, try to upload results.  Otherwise output to stdout.
Can override either case with `-U/--upload` or `-o/--output`.

Our current default assumes that users will want to upload results.  This is sensible, but mildly taxing for those users who don't.  Since upload requires a key in all cases, we can use that to infer whether to upload analysis results or just print.  We must also give the user a way to decide against our inference.

This is compatible with the `--fossa-api-key` PR (#66).


scruffystuffs:
Closing this.  In the current context, it doesn't make sense.

PR #68, opened 2020-05-15
Docs update

scruffystuffs:
Obviously we could add a lot more info here, but I wanted to focus on the README before anything else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,16 +6,19 @@ report dependency trees for many languages and package managers.  See the
 languages/managers.
 
 ```sh
-me@mydesk:~/myproject$ fossa2 analyze -o
-# Dependency analysis output
+export FOSSA_API_KEY=your-key-goes-here
+me@mydesk:~/myproject$ fossa2 analyze
 ```
+
+If you do not have an API key, please check the [FOSSA Documentation](https://docs.fossa.com/docs/api-reference)
+for instructions on creating API tokens.
 
 ## Table of Contents
 
 1. [Installation](#installation)
-1. [Basic Usage](#basic-usage)
-3. [Supported Strategies](#supported-languagesmanagers)
-4. [Contributing/Building From Source](#contributingbuilding-from-source)
+2. [Basic Usage](#basic-usage)
+3. [Supported Strategies](#supported-languages)
+4. [Contributing](#contributing)
 
 ## Installation
 
@@ -38,41 +41,40 @@ easier.
 The tool requires little-to-no configuration to run on its own.
 
 ```sh
-fossa2 analyze --output # Analyze the current working directory
-# OR
-fossa2 analyze --output --basedir /path/to/project  # Analyze a specific directory.
-```
-
-The `--output` flag tells the analyzer not to connect to a FOSSA server,
-instead printing out the analysis result to the console.
-
-While we're on the subject, you can (and should!) connect your output directly
-to your FOSSA server!  You can do this by simply supplying the FOSSA API key
-from your server ([more info](https://docs.fossa.com/docs/api-reference)) and
-using it as shown:
-
-```sh
 export FOSSA_API_KEY=your-key-goes-here
 fossa2 analyze
 # OR
 fossa2 analyze --fossa-api-key your-key-goes-here
 ```
 
-This not only lets you re-examine the results later, but allows you to check
-your output against your preset policies.
-
-**NOTE** If leaked, your FOSSA API key can grant an attacker access to your FOSSA
+**NOTE:** If leaked, your FOSSA API key can grant an attacker access to your FOSSA
 projects, and should be kept secret.  For this reason, we recommend supplying
 the API key with the environment variable, especially in a shared environment
 like a CI/CD server.
 
-## Supported Languages/Managers
+By default, the program will analyze the current working directory.  If you need
+to run in a different directory, you can simply pass that directory as an argument.
+
+```sh
+fossa2 analyze /path/to/project
+```
+
+The `--output` flag tells the analyzer not to connect to a FOSSA server,
+instead printing out the analysis result to the console.
+
+```sh
+fossa2 analyze --output
+# OR
+fossa2 analyze -o
+```
+
+## Supported Languages
 
 A non-exhaustive list of supported languages and managers can be found
 [here](docs/strategies.md).  This list is a work-in-progress, as some existing 
 strategies are not yet documented, but are implemented.
 
-## Building From Source
+## Contributing
 
 ### Building
 
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ In the base directory, run `cabal build`
 ### Running
 
 ```sh
-$ cabal run fossa -- analyze -d path/to/basedir/ -o
+$ cabal run fossa2 -- analyze -o path/to/basedir/
 ```
 
 This will produce analysis results on stdout
--- a/README.md
+++ b/README.md
@@ -47,10 +47,7 @@ fossa2 analyze
 fossa2 analyze --fossa-api-key your-key-goes-here
 ```
 
-**NOTE:** If leaked, your FOSSA API key can grant an attacker access to your FOSSA
-projects, and should be kept secret.  For this reason, we recommend supplying
-the API key with the environment variable, especially in a shared environment
-like a CI/CD server.
+**NOTE:** *If leaked, your FOSSA API key can grant an attacker access to your FOSSA projects, and should be kept secret.  For this reason, we recommend the API key with the environment variable, especially in a shared environment like a CI/CD server.*
 
 By default, the program will analyze the current working directory.  If you need
 to run in a different directory, you can simply pass that directory as an argument.
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ me@mydesk:~/myproject$ fossa2 analyze -o
 ## Table of Contents
 
 1. [Installation](#installation)
-1. [Usage](#usage)
+1. [Basic Usage](#basic-usage)
 3. [Supported Strategies](#supported-languagesmanagers)
 4. [Contributing/Building From Source](#contributingbuilding-from-source)
 
@@ -33,7 +33,7 @@ Set-ExecutionPolicy Bypass -Scope Process -Force; iex  ((New-Object System.Net.W
 The installer will report the installation path of the executable to make this
 easier.
 
-## Usage
+## Basic Usage
 
 The tool requires little-to-no configuration to run on its own.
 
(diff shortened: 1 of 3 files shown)


docs/strategies/rust.md
```diff
@@ -71,249 +75,3 @@ version, whether it is platform-specific, what build category it's included in
 (build, dev, or standard), and what features it compiles with.  For any further
 information, we have to compare this package and version with the corresponding
 entry in `.packages`.
-
-## Schema
```
liftM, line 75: Why delete this?

README.md
```diff
@@ -1,3 +1,22 @@
+# Spectrometer
+
+Spectrometer is a polyglot dependency analysis tool, designed to identify and
+report dependency trees for many languages and package managers.  See the
+[docs](docs/strategies.md) for a non-exhaustive list of supported
+languages/managers.
+
+```sh
+me@mydesk:~/myproject$ fossa2 analyze -o
```
liftM, line 9: We should rename this to `fossa` before we merge this `README` cc @cnr.

README.md
```diff
@@ -10,7 +29,52 @@ curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/spect
...
+easier.
+
+## Basic Usage
+
+The tool requires little-to-no configuration to run on its own.
+
+```sh
+fossa2 analyze --output # Analyze the current working directory
+# OR
+fossa2 analyze --output --basedir /path/to/project  # Analyze a specific directory.
```
liftM, line 43: Why do we take the directory as a flag and not an argument? That UX seems unintuitive, and I'd prefer to resolve that before merging the `README`.

README.md
```diff
@@ -10,7 +29,52 @@ curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/spect
...
+fossa2 analyze
+# OR
+fossa2 analyze --fossa-api-key your-key-goes-here
+```
+
+This not only lets you re-examine the results later, but allows you to check
+your output against your preset policies.
+
+**NOTE** If leaked, your FOSSA API key can grant an attacker access to your FOSSA
+projects, and should be kept secret.  For this reason, we reccommend supplying
```
liftM, line 65: Typo: "recommend".

README.md
```diff
@@ -10,7 +29,52 @@ curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/spect
...
+the API key with the environment variable, especially in a shared environment
+like a CI/CD server.
+
+## Supported Languages/Managers
```
liftM, line 69: I'd just call this "Supported Languages".

README.md
```diff
@@ -10,7 +29,52 @@ curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/spect
...
+
+A non-exhaustive list of supported languages and managers can be found
+[here](docs/strategies.md).  This list is a work-in-progress, as some existing 
+strategies are not yet documented, but are implmented.
+
+## Contributing/Building From Source
```
liftM, line 75: I'd just call this "Contributing".

README.md
```diff
@@ -10,7 +29,52 @@ curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/spect
...
+**NOTE** If leaked, your FOSSA API key can grant an attacker access to your FOSSA
+projects, and should be kept secret.  For this reason, we reccommend supplying
+the API key with the environment variable, especially in a shared environment
+like a CI/CD server.
+
+## Supported Languages/Managers
+
+A non-exhaustive list of supported languages and managers can be found
+[here](docs/strategies.md).  This list is a work-in-progress, as some existing 
+strategies are not yet documented, but are implmented.
```
liftM, line 73: Typo: "implemented".

README.md
```diff
@@ -1,3 +1,22 @@
...
+```
+
+## Table of Contents
+
+1. [Installation](#installation)
+1. [Basic Usage](#basic-usage)
+3. [Supported Strategies](#supported-languagesmanagers)
+4. [Contributing/Building From Source](#contributingbuilding-from-source)
+
 ## Installation
```
liftM, line 20: Do we have automated testing in this repository for the installation commands yet?

README.md
liftM, line 9: I would use the quick-start to demonstrate an analysis upload. Most people will read this snippet and just run it without reading anything else, and providing the output snippet first will confuse them.

README.md
scruffystuffs: I agree with this.  We can just set up an optional positional arg.  I'll PR that tomorrow, it should be quick.

README.md
scruffystuffs: I have never spelled that right on the first try.

README.md
scruffystuffs: It's probably a good idea to show people how to build from source as a line item.  A lot of developers like to do this when possible even if not contributing.  I'd rather leave this as just "Building from Source".

README.md
scruffystuffs: If a potential user that does not use the full FOSSA suite tries that, it won't work.  That behavior in a quickstart section would make me seriously reconsider using any tool published by that user/org.

README.md
scruffystuffs: Not likely.

README.md
scruffystuffs: That's a bit of a misnomer, considering that mostly we're analyzing files related to package managers, not languages.  I don't like the current title, but I'd like to revisit this.

docs/strategies/rust.md
scruffystuffs: I don't think it provides much value to our users.  Especially since this is not any output that they would normally see.  Also, there is a link to the schema in the text already, which will be kept up-to-date if there are any changes, so no documentation rot.

README.md
scruffystuffs: Turns out, that change isn't straightforward.  The `--basedir` option is part of the global options for the fossa executable.  I have a potential solution, but it's not great, and I don't think we should do it.  I'll link the PR here once it's uploaded.

README.md
scruffystuffs: I'll continue that logic here: 

The point of a quickstart should be "How do I check that this thing works?", not "Now that I've finished my other setup, how does this piece fit in?".  That belong in the user manual, and for our product, it belongs at the top of the user manual, since that is really what we're offering/selling.

I think a good analog to this is setting up a database server.  The quickstart there would not include things like security, HighAvailability, sharding, etc.  That's all critical for running in the real world, and should be documented well, but it's not the first thing you want to see during your first experiment with the tool.

README.md
liftM: I agree that it's a bit of a misnomer, but I would prefer easier comprehension over more pedantic correctness in user documentation.

README.md
liftM: I don't think a developer will be confused by this header not saying "Building From Source". For user documentation, I'd prefer easier comprehension over technical correctness.

README.md
scruffystuffs: "Pedantic" might be a little strong.  But I don't disagree with the rest of your point.  I'll change it to "Supported Package Managers".

README.md
liftM: For almost all users, the first thing they want to do is upload an analysis. We can provide the output snippet later down, but the first line in the quickstart should show them how to upload an analysis.

scruffystuffs: Ok, I have an idea how these can coexist.  I can make a small quickstart section that has both options and starts with the upload bit, but is prefaced with "If you have an API key".

I think we're underplaying the "I have literally nothing else set up yet" scenario, and in that case, we're responsible here for making a great first impression.

cnr: I genuinely don't know what to do about the naming, because I suspect in some cases customers may need to use both the old version and the new version -- at least until we have support for archive uploads and the languages from v1

README.md
cnr: We don't. As soon as we can dogfood with a haskell analyzer, it'll be part of the test suite (for at least the linux installer)

README.md
```diff
@@ -10,7 +29,52 @@ curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/spect
...
+the API key with the environment variable, especially in a shared environment
+like a CI/CD server.
+
+## Supported Languages/Managers
+
+A non-exhaustive list of supported languages and managers can be found
+[here](docs/strategies.md).  This list is a work-in-progress, as some existing 
+strategies are not yet documented, but are implemented.
+
+## Building From Source
```
cnr, line 75: I'd love to move this to a CONTRIBUTING.md or HACKING.md, but that can wait for another pr

README.md
liftM: Will they ever use both in the same environment? I don't think so, and we can instruct SE to specifically avoid integrating that way. Is there a likely I-need-both scenario that I'm missing?

liftM: > I think we're underplaying the "I have literally nothing else set up yet" scenario, and in that case, we're responsible here for making a great first impression.

Can you tell me more about why this is a prominent scenario?

In the last 2 years I've worked on this project, people almost never care about integrating their own code directly against the output. Generally, people only care about using this together with the service. In fact, I wouldn't even preface with "if you have an API key" - I would add a link to the web application docs and direct people to create an API key if they don't have one.

For comparison, think about Snyk's CLI or GitHub's CLI. Would you expect to use those _without_ integrating with the web application? The output mode is really only for serious power users, who can figure it out themselves without needing a quick start (besides, we'll write a power user guide later anyway).

scruffystuffs: I'm less concerned about power users and more concerned about someone whose first interaction with FOSSA is through this project, and that leads them to other FOSSA services.  But I'm OK to defer it to a power user guide.  And the historic context is a part I definitely haven't seen much of.  I'll shift it the way you're suggesting.

zlav: I generally agree with Leo here, but I will have a quick chat with an SE to make sure that it makes sense to them as well.

README.md
```diff
@@ -10,7 +29,52 @@ curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/spect
...
+fossa2 analyze --output # Analyze the current working directory
+# OR
+fossa2 analyze --output --basedir /path/to/project  # Analyze a specific directory.
+```
+
+The `--output` flag tells the analyzer not to connect to a FOSSA server,
+instead printing out the analysis result to the console.
+
+While we're on the subject, you can (and should!) connect your output directly
+to your FOSSA server!  You can do this by simply supplying the FOSSA API key
```
zlav, line 50: [nit] I would change "your" to "the" to avoid confusion around users thinking they need to run their own FOSSA server.

PR #69, opened 2020-05-19
Fix graph unfold utility

cnr:
`unfold` is only used in two strategies -- and it was incorrectly stopping at the second level of dependencies.

### DIFF
--- a/src/Graphing.hs
+++ b/src/Graphing.hs
@@ -92,13 +92,17 @@ edge parent child gr = gr { graphingAdjacent = adjacent' }
 -- - A way to convert a dependency @toDependency@
 --
 -- __Unfold does not work for recursive inputs__
-unfold :: Ord res => [dep] -> (dep -> [dep]) -> (dep -> res) -> Graphing res
-unfold seed getDeps toDependency = foldr addNode empty seed
+unfold :: forall dep res. Ord res => [dep] -> (dep -> [dep]) -> (dep -> res) -> Graphing res
+unfold seed getDeps toDependency = Graphing
+  { graphingDirect = S.fromList (map toDependency seed)
+  , graphingAdjacent = AM.vertices (map toDependency seed) `AM.overlay` AM.edges [(toDependency parentDep, toDependency childDep) | (parentDep,childDep) <- edgesFrom seed]
+  }
   where
-  addNode dep gr = direct res (foldr (edge res . toDependency) gr children)
-    where
-    children = getDeps dep
-    res = toDependency dep
+    edgesFrom :: [dep] -> [(dep,dep)]
+    edgesFrom nodes = do
+      node <- nodes
+      let children = getDeps node
+      map (node,) children ++ edgesFrom children
 
 -- | Build a graphing from a list, where all list elements are treated as direct
 -- dependencies
--- a/spectrometer.cabal
+++ b/spectrometer.cabal
@@ -211,6 +211,7 @@ test-suite unit-tests
     Googlesource.RepoManifestSpec
     Gradle.GradleSpec
     GraphUtil
+    GraphingSpec
     Maven.PluginStrategySpec
     Node.NpmLockSpec
     Node.PackageJsonSpec
--- a/src/Strategy/Node/NpmLock.hs
+++ b/src/Strategy/Node/NpmLock.hs
@@ -70,7 +70,7 @@ mkProjectClosure file lock = ProjectClosureBody
   dependencies = ProjectDependencies
     { dependenciesGraph    = buildGraph lock
     , dependenciesOptimal  = Optimal
-    , dependenciesComplete = Complete
+    , dependenciesComplete = NotComplete
     }
 
 data NpmPackage = NpmPackage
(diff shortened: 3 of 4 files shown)


src/Graphing.hs
```diff
@@ -92,13 +92,17 @@ edge parent child gr = gr { graphingAdjacent = adjacent' }
 -- - A way to convert a dependency @toDependency@
 --
 -- __Unfold does not work for recursive inputs__
-unfold :: Ord res => [dep] -> (dep -> [dep]) -> (dep -> res) -> Graphing res
-unfold seed getDeps toDependency = foldr addNode empty seed
+unfold :: forall dep res. Ord res => [dep] -> (dep -> [dep]) -> (dep -> res) -> Graphing res
```
scruffystuffs, line 95: Why do we need to add `forall dep res` here?

src/Graphing.hs
```diff
@@ -92,13 +92,17 @@ edge parent child gr = gr { graphingAdjacent = adjacent' }
...
+unfold seed getDeps toDependency = Graphing
+  { graphingDirect = S.fromList (map toDependency seed)
+  , graphingAdjacent = AM.vertices (map toDependency seed) `AM.overlay` AM.edges [(toDependency parentDep, toDependency childDep) | (parentDep,childDep) <- edgesFrom seed]
```
scruffystuffs, line 98: Can you explain this single-pipe syntax?

scruffystuffs:
Are we able to unit-test this fixed graphing logic?

src/Graphing.hs
scruffystuffs: Nvm, found [this](https://wiki.haskell.org/List_comprehension) explanation.

src/Graphing.hs
cnr: By default, this `forall` is automatically generated whenever you use a type variable, e.g., if I define
```hs
const :: a -> b -> a
```
the `forall` is implicit:
```hs
const :: forall a b. a -> b -> a
```

with `-XScopedTypeVariables`, the use of the explicit forall here means I can use `dep` in the type signature of `edgesFrom` in the `where` clause below.

---

in rust, `forall` is analogous to (pseudocode?):
`fn const<A,B>(a: A, b: B) -> A`

cnr:
Good idea; added a test

test/GraphingTest.hs
```diff
@@ -0,0 +1,20 @@
...
+import Prelude
+import Graphing
+import GraphUtil
+
+spec_unfold :: Spec
+spec_unfold = do
+  describe "graphing unfold" $ do
+    it "should unfold deeply" $ do
+      let graph :: Graphing Int
+          graph = unfold [10] (\x -> if x > 0 then [x-2] else []) id
```
cnr, line 16: the `id` here highlights that `unfold` has the wrong type. but it sure is convenient. (same issue as `withLabeling`)

src/Graphing.hs
```diff
@@ -92,13 +92,17 @@ edge parent child gr = gr { graphingAdjacent = adjacent' }
...
+  }
   where
-  addNode dep gr = direct res (foldr (edge res . toDependency) gr children)
-    where
-    children = getDeps dep
-    res = toDependency dep
+    edgesFrom :: [dep] -> [(dep,dep)]
+    edgesFrom nodes = do
+      node <- nodes
```
scruffystuffs, line 103: [nit]: This is probably a little confusing to most readers, including me.  I had to really think about how this would work.  I don't know how many people will read this AND need to fully understand it AND not be able to figure out like I did, but it may be worth writing the iteration more explicitly.

src/Graphing.hs
scruffystuffs, line 98: [nit] Style-wise, when do you suggest using infix notation?  I find it hard to read sometimes, probably due to my more limited experience.

src/Graphing.hs
cnr: I agree it's likely to be confusing the first few times it's encountered -- but I think the alternatives are potentially more-confusing.

Having an intuition for how the list monad works (or: "not every monad is a container with one thing in it") is useful in other places too, especially for effects like `NonDet` in fused-effects, or reasoning about megaparsec/etc `Parser` monads. I wish there was a better way to signal that "this is a list do block"

scruffystuffs: I'm on the fence here.  I marked this as a nit, because it shouldn't block the PR, but I'm not convinced that this is the most readable syntax for what is essentially a nested map pipeline.  But this code may not change, so no need to block it here.

src/Graphing.hs
cnr: depends on the situation -- I usually use it in one of these cases:
1. I'm using a binary function in a section, e.g.,
```hs
(`M.member` someMap)
```
2. A function is named in a way that's meant to be used infix, e.g., `Data.Function.on` or `Text.URI.relativeTo`
3. To combine two larger expressions _of the same type_. `AM.overlay (somereallylongexpression) (anotherreallylongexpression)` is fine too. In general, the intuition around infix functions is that they have very low precedence, so
```hs
foo bar `and` baz

-- is almost always going to be interpreted as
(foo bar) `and` (baz)
```

PR #70, opened 2020-05-20
Erlang support with rebar3

zlav:
This PR adds support for Erlang through the rebar3 package manager by parsing the output from the command `rebar3 tree -v` for dependencies. Example output
```
   ├─ one─1.0.0 (https://github.com/dep/one)
   │  ├─ two─2.0.0 (https://github.com/dep/two)
   │  │  └─ three─3.0.0 (https://github.com/dep/three)
   │  └─ four─4.0.0 (https://github.com/dep/four)
   └─ five─5.0.0 (hex package)
```

### DIFF
--- a/src/Strategy/Erlang/Rebar3Tree.hs
+++ b/src/Strategy/Erlang/Rebar3Tree.hs
@@ -12,6 +12,7 @@ import Prologue
 
 import Control.Carrier.Error.Either
 import qualified Data.Map.Strict as M
+import qualified Data.Text as T
 import Text.Megaparsec
 import Text.Megaparsec.Char
 
@@ -72,8 +73,14 @@ buildGraph deps = run . withLabeling toDependency $ do
     applyLabel (RebarSource src) dep = dep { dependencyLocations = src : dependencyLocations dep }
 
     start =
-      Dependency { dependencyType = HexType
-                 , dependencyName = depName pkg
+      Dependency { dependencyType =
+                      case T.isInfixOf (T.pack "github.com") (depLocation pkg) of
+                      True -> GitType
+                      False -> HexType
+                 , dependencyName =
+                      case T.isInfixOf (T.pack "github.com") (depLocation pkg) of
+                      True -> depLocation pkg
+                      False -> depName pkg
                  , dependencyVersion = Just (CEq (depVersion pkg))
                  , dependencyLocations = []
                  , dependencyEnvironments = []
--- a/src/Strategy/Erlang/Rebar3Tree.hs
+++ b/src/Strategy/Erlang/Rebar3Tree.hs
@@ -54,6 +54,12 @@ mkProjectClosure dir deps = ProjectClosureBody
     , dependenciesComplete = NotComplete
     }
 
+type RebarGrapher = LabeledGrapher Rebar3Dep RebarLabel
+
+data RebarLabel =
+    RebarSource Text -- location
+  deriving (Eq, Ord, Show, Generic)
+
 buildGraph :: [Rebar3Dep] -> Graphing Dependency
 buildGraph deps = run . withLabeling toDependency $ do
   traverse_ direct deps
--- a/src/Strategy/Erlang/Rebar3Tree.hs
+++ b/src/Strategy/Erlang/Rebar3Tree.hs
@@ -1,0 +1,128 @@
+module Strategy.Erlang.Rebar3Tree
+  ( discover
+  , analyze
+
+  , buildGraph
+  , rebar3TreeParser
+  , Rebar3Dep(..)
+  )
+  where
+
+import Prologue
+
+import Control.Carrier.Error.Either
+import qualified Data.Map.Strict as M
+import Text.Megaparsec
+import Text.Megaparsec.Char
+
+import DepTypes
+import Discovery.Walk
+import Effect.Exec
+import Graphing (Graphing)
+import qualified Graphing
+import Types
+
+discover :: HasDiscover sig m => Path Abs Dir -> m ()
+discover = walk $ \dir _ files -> do
+  case find (\f -> fileName f `elem` ["rebar.config"]) files of
+    Nothing -> pure ()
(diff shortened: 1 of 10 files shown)


src/Strategy/Erlang/Rebar3Tree.hs
```diff
@@ -0,0 +1,156 @@
...
+import DepTypes
+import Discovery.Walk
+import Effect.Exec
+import Graphing (Graphing)
+import Effect.Grapher
+import Types
+
+discover :: HasDiscover sig m => Path Abs Dir -> m ()
+discover = walk $ \dir _ files -> do
+  case find (\f -> fileName f `elem` ["rebar.config"]) files of
```
cnr, line 28: [nit] use `== "rebar.config"` instead of `elem` here

src/Strategy/Erlang/Rebar3Tree.hs
```diff
@@ -0,0 +1,156 @@
...
+  dependencies = ProjectDependencies
+    { dependenciesGraph    = buildGraph deps
+    , dependenciesOptimal  = NotOptimal
+    , dependenciesComplete = NotComplete
+    }
+
+type RebarGrapher = LabeledGrapher Rebar3Dep RebarLabel
+
+data RebarLabel =
+    RebarSource Text -- location
```
cnr, line 60: this label isn't actually used -- so let's not use `LabeledGrapher` here. can we use `unfold` instead for `buildGraph`?

src/Strategy/Erlang/Rebar3Tree.hs
```diff
@@ -0,0 +1,156 @@
...
+  where
+  toDependency :: Rebar3Dep -> Set RebarLabel -> Dependency
+  toDependency pkg = foldr applyLabel start
+    where 
+    applyLabel :: RebarLabel -> Dependency -> Dependency
+    applyLabel (RebarSource src) dep = dep { dependencyLocations = src : dependencyLocations dep }
+
+    start =
+      Dependency { dependencyType =
+                      case T.isInfixOf (T.pack "github.com") (depLocation pkg) of
```
cnr, line 77: [nit] `T.pack` on string literals is redunant because of `OverloadedStrings` -- so this can just be `T.isInfixOf "github.com" (depLocation pkg)`

cnr, line 77: [nit]
```hs
case ... of
  False -> ...
  True -> ...
```

is the same as `if ... then ... else ...`

src/Strategy/Erlang/Rebar3Tree.hs
```diff
@@ -0,0 +1,156 @@
...
+
+  findVersion :: Parser Text
+  findVersion = takeWhileP (Just "version") (/= ' ')
+
+  findLocation :: Parser Text
+  findLocation = takeWhileP (Just "location") (/= ')')
+
+  rebarDep :: Integer -> Parser [Rebar3Dep]
+  rebarDep depth = do
+    _ <- chunk " "
```
cnr, line 136: let's revisit this to use the lexer machinery soon

cnr: I bet we could even use `indentBlock` somehow

src/Strategy/Erlang/Rebar3Tree.hs
```diff
@@ -0,0 +1,156 @@
...
+  findName :: Parser Text
+  findName = takeWhileP (Just "dep") (/= '─')
+
+  findVersion :: Parser Text
+  findVersion = takeWhileP (Just "version") (/= ' ')
+
+  findLocation :: Parser Text
+  findLocation = takeWhileP (Just "location") (/= ')')
+
+  rebarDep :: Integer -> Parser [Rebar3Dep]
```
cnr, line 134: [nit] use `Int` here

src/Strategy/Erlang/Rebar3Tree.hs
```diff
@@ -0,0 +1,156 @@
...
+  rebarDep depth = do
+    _ <- chunk " "
+    countSlash <- many "  │"
+    _ <- satisfy (\_ -> (fromIntegral (length countSlash)) == depth)
```
cnr, line 138: this is clever! this also has another name from `Control.Monad`: `guard :: Alternative f => Bool -> f ()`:
```hs
guard (length countSlash == depth)
```

you've used `Alternative` elsewhere (it defines `<|>` and `empty`), and `guard` is defined as:
```hs
guard :: Alternative f => Bool -> f ()
guard b = if b then pure () else empty
```

zlav: So I can't seem to get this to work, and I believe the reason has to do with how `satisfy` fails the parser if the condition is not met. `guard` seems to be causing the parser to ignore the initial `concat` in `rebar3TreeParser = concat <$> ((try (rebarDep 0) <|> ignoredLine) `sepBy` eol) <* eof` when I use it like this `_ <- guard (length slashCount == depth)`. Do you have any idea why this would be happening?

zlav: I've also tried `guard $ not ....` but that doesn't seem to work. `satisfy` will backtrack the parser which I would expect `guard` to do as well but for some reason it is not.

src/Strategy/Erlang/Rebar3Tree.hs
zlav: So I implemented this and I can't get the dependency graph construction to work for dependencies greater than 1 level. I think this may have been a known error and is one of the reasons I went with the `LabeledGrapher` to begin with. Is this familiar with you? I can't find an example where `unfold` is being used with deps deeper than 1 level.

zlav: From the code, It appears that the `unfold` function only handles deep deps for direct dependencies. It feel like the best path forward here would be to maintain the labeled grapher as it can handle deep deps and the locations that will be found once we can correctly handle `git` dependencies.

PR #71, opened 2020-05-20
Move --basedir to common positional argument

scruffystuffs:
I'm going to say upfront, I don't like this.  I created this to show the effects of the proposed change in [this thread](https://github.com/fossas/spectrometer/pull/68#discussion_r427705628), and I'm not a fan of it.

That said, it's more intuitive for the user, and more "correct" cli behavior.  But it adds maintenance cost for every new command we add.

EDIT: Some suggestions below have modified the initial implementation, which I did not like, into a refactor where the command modules are taking responsibility for their own validation, which I definitely support.  As our command structure grows, this helps manage the growing complexity of the application.

### DIFF
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -75,7 +75,6 @@ analyze ::
   -> Maybe Text -- ^ cli override for revision
   -> m ()
 analyze basedir destination overrideName overrideRevision = do
-  setCurrentDir basedir
   capabilities <- liftIO getNumCapabilities
 
   (closures,(failures,())) <- runOutput @ProjectClosure $ runOutput @ProjectFailure $
--- a/src/App/Fossa/Main.hs
+++ b/src/App/Fossa/Main.hs
@@ -92,8 +92,8 @@ analyzeOpts :: Parser AnalyzeOptions
 analyzeOpts =
   AnalyzeOptions
     <$> switch (long "output" <> short 'o' <> help "Output results to stdout instead of uploading to fossa")
-    <*> baseDirArg
     <*> metadataOpts
+    <*> baseDirArg
 
 metadataOpts :: Parser ProjectMetadata
 metadataOpts =
@@ -127,8 +127,8 @@ data Command
 
 data AnalyzeOptions = AnalyzeOptions
   { analyzeOutput :: Bool
-  , analyzeBaseDir :: FilePath
   , analyzeMetadata :: ProjectMetadata
+  , analyzeBaseDir :: FilePath
   }
 
 data TestOptions = TestOptions
--- a/src/App/Fossa/Main.hs
+++ b/src/App/Fossa/Main.hs
@@ -26,7 +26,7 @@ appMain = do
 
   case optCommand of
     AnalyzeCommand AnalyzeOptions{..} -> do
-      _ <- changeDir analyzeBaseDir
+      changeDir analyzeBaseDir
       if analyzeOutput
         then analyzeMain logSeverity OutputStdout optProjectName optProjectRevision
         else case maybeApiKey of
@@ -34,7 +34,7 @@ appMain = do
           Just key -> analyzeMain logSeverity (UploadScan optBaseUrl key analyzeMetadata) optProjectName optProjectRevision
 
     TestCommand TestOptions{..} -> do
-      _ <- changeDir testBaseDir
+      changeDir testBaseDir
       case maybeApiKey of
         Nothing -> die "A FOSSA API key is required to run this command"
         Just key -> testMain optBaseUrl key logSeverity testTimeout optProjectName optProjectRevision
--- a/src/App/Fossa/Main.hs
+++ b/src/App/Fossa/Main.hs
@@ -24,22 +24,23 @@ appMain = do
 
   maybeApiKey <- checkAPIKey optAPIKey
 
-  basedir <- validateDir optBasedir
-
-  setCurrentDir basedir
-
   case optCommand of
-    AnalyzeCommand AnalyzeOptions{..} ->
+    AnalyzeCommand AnalyzeOptions{..} -> do
+      _ <- changeDir analyzeBaseDir
       if analyzeOutput
         then analyzeMain logSeverity OutputStdout optProjectName optProjectRevision
         else case maybeApiKey of
           Nothing -> die "A FOSSA API key is required to run this command"
           Just key -> analyzeMain logSeverity (UploadScan optBaseUrl key analyzeMetadata) optProjectName optProjectRevision
 
-    TestCommand TestOptions{..} ->
+    TestCommand TestOptions{..} -> do
+      _ <- changeDir testBaseDir
       case maybeApiKey of
         Nothing -> die "A FOSSA API key is required to run this command"
         Just key -> testMain optBaseUrl key logSeverity testTimeout optProjectName optProjectRevision
+
+changeDir :: FilePath -> IO ()
+changeDir path = validateDir path >>= setCurrentDir
 
 -- | Try to fetch FOSSA_API_KEY from env if not supplied from cmdline
 checkAPIKey :: Maybe Text -> IO (Maybe Text)
@@ -56,11 +57,13 @@ validateDir dir = do
 
   pure absolute
 
+baseDirArg :: Parser String
+baseDirArg = argument str (metavar "DIR" <> help "Set the base directory for scanning (default: current directory)" <> value ".")
+
 opts :: Parser CmdOptions
 opts =
   CmdOptions
     <$> switch (long "debug" <> help "Enable debug logging")
-    <*> strOption (long "basedir" <> short 'd' <> metavar "DIR" <> help "Set the base directory for scanning (default: current directory)" <> value ".")
     <*> urlOption (long "endpoint" <> metavar "URL" <> help "The FOSSA API server base URL" <> value urlOpts)
     <*> optional (strOption (long "project" <> help "this repository's URL or VCS endpoint (default: VCS remote 'origin')"))
     <*> optional (strOption (long "revision" <> help "this repository's current revision hash (default: VCS hash HEAD)"))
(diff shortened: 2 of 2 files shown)


liftM:
> But it adds maintenance cost for every new command we add.

Can you elaborate on this maintenance cost?

Why is this PR also adding a base directory for `fossa test`?

scruffystuffs:
For every new command we add, we may have to move to the correct directory, and we have to do it in the main module, since that's the only place we know about the basedir.  We could abstract this, but that's more work, and likely requires some good haskell intuition, meaning @cnr has to do it/design it.  Currently all commands get this behavior for free.

`fossa test` has to know which project it's polling against to see if there are any issues.  It uses the current dir to infer the project info.

cnr:
The repetition here is okay, and I agree this change feels more correct.

One minor issue: if you do `fossa2 analyze --help`, `[DIR]` appears in a pretty awkward spot in the argument list. The order in which you apply parsers changes the order of the arguments/switches in the help output

scruffystuffs:
I'm not a fan of this repetition.  We'll have to remember it forever, and the typechecker won't help us to do that.  I'd be happy with this if we passed the basedir to the consuming functions, since that would leverage the typechecker and make us actively decide whether we we're going to use it or not.
They (`analyzeMain` and `testMain`) are already returning `IO ()`, so adding the same changedir logic would be quite simple.

The other option is integration tests, and while we should have as many integration tests as we need, we should try to **need** as few as possible.

If I'm outvoted, I'm obviously not going to refuse to merge this, but I do feel strongly about this.

src/App/Fossa/Analyze.hs
```diff
@@ -59,10 +60,10 @@ data ScanDestination
   | OutputStdout
   deriving (Generic)
  
-analyzeMain :: Severity -> ScanDestination -> Maybe Text -> Maybe Text -> IO ()
-analyzeMain logSeverity destination name revision = do
-  basedir <- getCurrentDir
-  withLogger logSeverity $ analyze basedir destination name revision
+analyzeMain :: Severity -> FilePath -> ScanDestination -> Maybe Text -> Maybe Text -> IO ()
+analyzeMain logSeverity basedir destination name revision = do
+  basedir' <- validateDir basedir
```
liftM, line 65: We should probably validate these _before_ we call `analyzeMain`, and maybe use a `newtype` to enforce validation.

src/App/Fossa/Test.hs
```diff
@@ -27,16 +27,17 @@ testMain
...
   -> Severity
+  -> FilePath
   -> Int -- ^ timeout (seconds)
   -> Maybe Text -- ^ cli override for name
   -> Maybe Text -- ^ cli override for revision
   -> IO ()
-testMain baseurl apiKey logSeverity timeoutSeconds overrideName overrideRevision= do
-  basedir <- getCurrentDir
+testMain baseurl apiKey logSeverity baseDir timeoutSeconds overrideName overrideRevision= do
+  basedir' <- validateDir baseDir
```
liftM, line 36: Same here re: validating beforehand.

cnr:
if you're up for it, I would take this a step further: `analyzeMain` and `testMain` should be in charge of all of the validation of their arguments.

To do this, you'd pass the entire `AnalyzeOptions` to `analyzeMain` and the entire `TestOptions` to `testMain` -- and move the rest of the validation logic from `App.Fossa.Main` to their respective command modules.

---

one thing that was missed in the most recent commit is the `setCurrentDir` for each command

liftM:
> one thing that was missed in the most recent commit is the `setCurrentDir` for each command

Can we add a test for catching this kind of issue? I imagine this would need to be an integration test on some fixtures on disk.

scruffystuffs:
> one thing that was missed in the most recent commit is the setCurrentDir for each command

Actually, that was intentional.  From my understanding, we were using that as a shortcut to get the correct directory to work from.  Not only does that seem less Haskelly (meaning that the type system can't communicate that to us) it's also less explicit.  If anyone needs more convincing that it wasn't explicit before, consider that I couldn't know about it without grepping the codebase for "getCurrentDir".

We are now passing the operational directory (which *might* be the current directory) as an argument.

scruffystuffs:
@liftM I'm happy with this change if we do @cnr 's suggestion of moving the commands into their respective modules.  That help us to understand what we get from main, what we use and what we don't.

As for path validation, the `validateDir` function has type `FilePath -> IO (Path Abs Dir)`, which forces us to validate.  Our operations all require `Path a b`, so the typechecker is forcing us to validate.  Is there another piece you think we need to have tests for?

src/BaseDir.hs
```diff
@@ -0,0 +1,24 @@
...
+    )
+    where
+
+import Prologue
+
+import Options.Applicative
+import Path.IO (resolveDir', doesDirExist)
+import System.Exit (die)
+
+-- | Validate that a filepath points to a directory and the directory exists
```
cnr, line 13: I think the type makes it clear, but adding another sentence about how the program exits if the filepath doesn't validate could be helpful

src/App/Fossa/Test.hs
```diff
@@ -1,42 +1,67 @@
...
   -> Severity
+  -> Path Abs Dir
   -> Int -- ^ timeout (seconds)
   -> Maybe Text -- ^ cli override for name
   -> Maybe Text -- ^ cli override for revision
   -> IO ()
-testMain baseurl apiKey logSeverity timeoutSeconds overrideName overrideRevision= do
-  basedir <- getCurrentDir
+testMainInner baseurl apiKey logSeverity basedir' timeoutSeconds overrideName overrideRevision = do
+  -- basedir' <- validateDir baseDir
```
cnr, line 60: [nit] kill the comment, rename `basedir'` to `basedir`?

src/App/Fossa/Test.hs
```diff
@@ -1,42 +1,67 @@
...
 import System.IO (stderr)
-import System.Exit (exitSuccess, exitFailure)
+import System.Exit (exitSuccess, exitFailure, die)
 import OptionExtensions
 
 pollDelaySeconds :: Int
 pollDelaySeconds = 8
 
-testMain
+testOpts :: Parser TestOptions
```
cnr, line 29: I like this. I hadn't considered moving the `Parser`s in here too

src/BaseDir.hs
```diff
@@ -0,0 +1,24 @@
...
+validateDir :: FilePath -> IO (Path Abs Dir)
+validateDir dir = do
+  absolute <- resolveDir' dir
+  exists <- doesDirExist absolute
+
+  unless exists (die $ "ERROR: Directory " <> show absolute <> " does not exist")
+
+  pure absolute
+
+baseDirArg :: Parser String
```
cnr, line 23: to make the intent clearer -- and because `String` is so vague (and `FilePath` is just an alias for `String`) -- you can use a newtype here:

```hs
newtype Basedir = Basedir { unBasedir :: FilePath }
```

and then change the type of `validateDir`/`baseDirArg`:
```hs
validateBasedir :: Basedir -> IO (Path Abs Dir)
baseDirArg :: Parser Basedir
```

src/App/Fossa/Analyze.hs
```diff
@@ -51,18 +55,50 @@ import qualified Strategy.Python.ReqTxt as ReqTxt
...
+  ProjectMetadata
+    <$> optional (strOption (long "title" <> help "the title of the FOSSA project. (default: the project name)"))
+    <*> optional (strOption (long "branch" <> help "this repository's current branch (default: current VCS branch)"))
+    <*> optional (strOption (long "project-url" <> help "this repository's home page"))
+    <*> optional (strOption (long "jira-project-key" <> help "this repository's JIRA project key"))
+    <*> optional (strOption (long "link" <> help "a link to attach to the current build"))
+    <*> optional (strOption (long "team" <> help "this repository's team inside your organization"))
+    <*> optional (strOption (long "policy" <> help "the policy to assign to this project in FOSSA"))
+
+mkScanDestination :: Bool -> Maybe Text -> UrlOption -> ProjectMetadata -> IO ScanDestination
```
cnr, line 91: This type signature is very vague to me. What's the `Bool`? what's the `Maybe Text`? There are two ways to address this -- one is to use a newtype wrapper for API key (which I should've done when I first wrote this, and is kind of a pervasive issue across these modules) + a clearer datatype than `Bool`.

The other way is to name the arguments and use `case`:

```hs
mkScanDestination forceStdout apiKey baseUrl metadata =
  case (forceStdout, apiKey) of
    (True, _) -> ...
    (_, Nothing) -> die ...
    (_, Just key) -> ...
```

src/App/Fossa/Main.hs
```diff
@@ -24,43 +21,22 @@ appMain = do
...
-
   case optCommand of
-    AnalyzeCommand AnalyzeOptions{..} ->
-      if analyzeOutput
-        then analyzeMain logSeverity OutputStdout optProjectName optProjectRevision
-        else case maybeApiKey of
-          Nothing -> die "A FOSSA API key is required to run this command"
-          Just key -> analyzeMain logSeverity (UploadScan optBaseUrl key analyzeMetadata) optProjectName optProjectRevision
+    AnalyzeCommand parsedOpts ->
+      analyzeMain parsedOpts maybeApiKey logSeverity optBaseUrl optProjectName optProjectRevision
```
cnr, line 26: it looks like we're just passing in everything from the `CmdOptions` (now poorly-named -- `GlobalOptions`?) -- would it make it clearer to pass in the entire `GlobalOptions`?

src/App/Fossa/Analyze.hs
scruffystuffs: Agree with your point but not sure how to proceed.  Let's pair on this.

src/App/Fossa/Main.hs
scruffystuffs: Agree with your point but not sure how to proceed.  Let's pair on this.

scruffystuffs:
After discussing with @cnr, the level of change to make this new implementation worth it should be part of a greater "CLI Refactoring" effort.  We should continue with the original change proposed, and take on the technical debt along with other CLI-based technical debt we currently have.  I will create a tracking issue for these items, but we can move forward with this change without blocking on that issue.

PR #73, opened 2020-05-28
Add --json command switch to `fossa test`

cnr:
This mirrors the behavior of cliv1's `fossa test --json`


src/App/Fossa/Main.hs
```diff
@@ -1,43 +1,41 @@
-
 module App.Fossa.Main
-  ( appMain
-  ) where
-
-import Prologue
+  ( appMain,
```
cnr, line 2: There's a lot of line noise in this file because I ran the formatter on it

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -250,6 +250,7 @@ renderIssueType = \case
 
 data Issue = Issue
   { issueId :: Int
+  , issuePriorityString :: Maybe Text -- we only use this field for `fossa test --json`
```
cnr, line 253: It feels weird to do a roundtrip serialization for json output, but I think this is correct

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -283,10 +302,22 @@ instance FromJSON IssueType where
     "outdated_dependency" -> pure IssueOutdatedDependency
     other -> pure (IssueOther other)
 
+instance ToJSON IssueType where
```
cnr, line 305: wish we were using a bidirectional parser 🤔 . lots of room for error in defining FromJSON/ToJSON separately

scruffystuffs: Can we unit test for this so that those potential errors get caught?

test/App/Fossa/FossaAPIV1Spec.hs
```diff
@@ -0,0 +1,54 @@
...
+import Test.Hspec
+import Test.Hspec.Hedgehog
+import Prelude
+
+spec :: Spec
+spec = do
+  describe "Issues ToJSON/FromJSON instances" $ do
+    it "are roundtrippable" $ hedgehog $ do
+      issues <- forAll genIssues
+      roundtripJson issues
```
cnr, line 17: This is the property-based test for roundtripping Issues serialization

test/App/Fossa/FossaAPIV1Spec.hs
```diff
@@ -0,0 +1,54 @@
...
+
+genIssues :: Gen Issues
+genIssues =
+  Issues
+    <$> Gen.int (Range.linear minBound maxBound)
+    <*> Gen.list (Range.linear 0 100) genIssue
+    <*> arbitraryText
+
+genIssue :: Gen Issue
+genIssue =
```
cnr, line 27: These are annoying to write -- there's lots of boilerplate. Upside: the test was two lines of code with the boilerplate out of the way

cnr:
There are a lot of unrelated changes because I also switch here from `tasty` to `hspec` + `hspec-hedgehog`. I've been meaning to do that for a while now, and this served as a forcing function.

PR #74, opened 2020-05-28
Only run IPR scan if all of the IPR options are provided

spatten:
You can also forcibly disable the IPR scan using the `--ipr-disabled` command
line flag

### DIFF
--- a/src/App/VPSScan/Main.hs
+++ b/src/App/VPSScan/Main.hs
@@ -45,10 +45,12 @@ runIPROpts = RunIPR.IPROpts
                   <$> iprCmdPathOpt
                   <*> nomosCmdPathOpt
                   <*> pathfinderCmdPathOpt
+                  <*> iprEnabledOpt
                 where
-                    iprCmdPathOpt = strOption (long "ipr" <> metavar "STRING" <> help "Path to the IPR executable")
-                    nomosCmdPathOpt = strOption (long "nomossa" <> metavar "STRING" <> help "Path to the nomossa executable")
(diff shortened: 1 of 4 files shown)


spatten:
> I'm not super convinced that a disable flag is something we want, especially if we plan running IPR only if all of the flags are present. It's just as easy to omit the options instead of passing an additional --ipr-disabled flag
> 
> This is a good overall approach, with one minor issue. As a general principle, we try to validate things upstream -- so passing a concrete `IPROpts` to `runIPRScan`, rather than one we have to validate, is usually preferred.
> 
> With that in mind, here's an alternative implementation plan that I think will require fewer changes/additions:
> 
>     1. keep `IPROpts` as-is
> 
>     2. make the field in `VPSOpts` `Maybe IPROpts` (and use `optional` as appropriate)
> 
>     3. if we're keeping the disableIpr flag, add it to VPSOpts
>        3b. wrap the `runIPRScan basedir scanId vpsOpts` with `unless :: Applicative f => Bool -> f () -> f ()`

You're right, that's a much cleaner way to do it. I went back and forth on adding the `--ipr-disabled` flag, so I'm happy to take it out :)

I tried to implement the validation in `VPSScan/Main.hs` without adding the `RunIPR.validateIprOpts` function and `RunIPR.UnvalidatedIPROpts` record type, but couldn't get that working. If there's an elegant way to do that, I'd love to see it. If not, this seems reasonable to me.

spatten:
This closes https://github.com/fossas/issues/issues/131

PR #78, opened 2020-06-04
Fossa report data structures

scruffystuffs:
Part of fossas/issues#93.

Items to handle:
- [x] Attributions
- [ ] Dependencies
- [ ] Licenses
- [ ] `newtype`s for each so we can print to `stdout`


test/App/Fossa/Report/AttributionSpec.hs
```diff
@@ -0,0 +1,77 @@
...
+genAttribution :: Gen Attribution
+genAttribution =
+  Attribution
+    <$> genProject
+    <*> Gen.list defaultRange genDependency
+    <*> Gen.list defaultRange genDependency
+    <*> genLicenseMap
+
+tuplify :: Monad m => m a -> m b -> m (a, b)
+tuplify a b = do
```
cnr, line 52: [nit] this is the same as `(,) <$> a <*> b` with applicative syntax
or `liftA2 (,) a b`

src/App/Fossa/Report/Attribution.hs
```diff
@@ -0,0 +1,141 @@
...
+    LicenseName (..),
+    Project (..),
+  )
+where
+
+import Prologue
+
+newtype LicenseName
+  = LicenseName {rawName :: Text}
+  deriving (Eq, Ord, Show, FromJSON, FromJSONKey, ToJSON, ToJSONKey)
```
cnr, line 15: I wouldn't have thought of this -- this is good. nit: is `ToJSON` and `FromJSON` required here?

scruffystuffs: Yeah it's used in values in other places.

test/App/Fossa/Report/AttributionSpec.hs
```diff
@@ -0,0 +1,74 @@
...
+  License
+    <$> (LicenseName <$> arbitraryText)
+    <*> Gen.maybe (LicenseContents <$> arbitraryText)
+
+genProject :: Gen Project
+genProject = Project <$> arbitraryText <*> arbitraryText
+
+genDependency :: Gen Dependency
+genDependency =
+  Dependency
```
zlav, line 27: I don't entirely understand what is happening here but it appears to be doing the correct thing. We should chat about this tomorrow because I'm interested to know what is happening!

PR #79, opened 2020-06-17
Implement 80/20 spec analyzer

scruffystuffs:
The tests should cover the overview of how we actually parse these files.  I'll be writing a doc soon for what we're currently doing vs what we could potentially do, but if not done in time, we might be able to submit this PR without that doc.

### DIFF
--- a/src/Strategy/RPM.hs
+++ b/src/Strategy/RPM.hs
@@ -45,11 +45,9 @@ data Dependencies
 
 discover :: HasDiscover sig m => Path Abs Dir -> m ()
 discover = walk $ \dir _ files ->
-  case find (\f -> ".spec" `isSuffixOf` fileName f) files of
-    Nothing -> pure WalkContinue
-    Just specFile -> do
-      runSimpleStrategy "rpm-spec" RPMGroup $ fmap (mkProjectClosure dir) (analyze specFile)
-      pure WalkSkipAll
+  let specs = find (\f -> ".spec" `isSuffixOf` fileName f) files
+      analzyeFile specFile = runSimpleStrategy "rpm-spec" RPMGroup $ fmap (mkProjectClosure dir) (analyze specFile)
+   in traverse_ analzyeFile specs >> pure WalkSkipAll
 
 analyze :: (Has ReadFS sig m, Has (Error ReadFSErr) sig m) => Path Rel File -> m (Graphing Dependency)
 analyze specFile = do
@@ -121,8 +119,14 @@ getTypeFromLine line = safeReq
   where
     (header, value) = splitOnceOn ": " line
     (pkgName, rawConstraint) = splitOnceOn " " $ T.strip value
+    --
+    isSafeName :: Text -> Bool
     isSafeName name = not $ "%{" `T.isInfixOf` name
+    -- TODO: temporarily ignore names with macros, until we support expansion
+    safeReq :: Maybe RequiresType
     safeReq = if isSafeName pkgName then req else Nothing
+    --
+    req :: Maybe RequiresType
     req = case header of
       "BuildRequires" -> Just . BuildRequires . RPMDependency pkgName $ buildConstraint rawConstraint
       "Requires" -> Just . RuntimeRequires . RPMDependency pkgName $ buildConstraint rawConstraint
@@ -131,6 +135,7 @@ getTypeFromLine line = safeReq
 buildDependencies :: [RequiresType] -> Dependencies
 buildDependencies = foldr addDep blankDeps
   where
+    addDep :: RequiresType -> Dependencies -> Dependencies
     addDep req deps = case req of
       BuildRequires dep -> deps {depBuildRequires = dep : depBuildRequires deps}
       RuntimeRequires dep -> deps {depRuntimeRequires = dep : depRuntimeRequires deps}
--- a/src/Strategy/RPM.hs
+++ b/src/Strategy/RPM.hs
@@ -31,28 +31,24 @@ data RPMDependency
       }
   deriving (Eq, Ord, Show, Generic)
 
-{- These two structures should make it very easy to add other headers like `Requires`
--}
-newtype RequiresType
+data RequiresType
   = BuildRequires RPMDependency
+  | RuntimeRequires RPMDependency
   deriving (Eq, Ord, Show, Generic)
 
-newtype Dependencies
+data Dependencies
   = Dependencies
-      { depBuildRequires :: [RPMDependency]
+      { depBuildRequires :: [RPMDependency],
+        depRuntimeRequires :: [RPMDependency]
       }
   deriving (Eq, Ord, Show, Generic)
-
-{-
-Yep...  Those ones right above me.  Kinda rude.  They're actually standing on my head.
--}
 
 discover :: HasDiscover sig m => Path Abs Dir -> m ()
 discover = walk $ \dir _ files ->
   case find (\f -> ".spec" `isSuffixOf` fileName f) files of
     Nothing -> pure WalkContinue
     Just specFile -> do
-      runSimpleStrategy "rpm-spec" RustGroup $ fmap (mkProjectClosure dir) (analyze specFile)
+      runSimpleStrategy "rpm-spec" RPMGroup $ fmap (mkProjectClosure dir) (analyze specFile)
       pure WalkSkipAll
 
 analyze :: (Has ReadFS sig m, Has (Error ReadFSErr) sig m) => Path Rel File -> m (Graphing Dependency)
@@ -80,7 +76,7 @@ toDependency pkg =
   foldr
     applyLabel
     Dependency
-      { dependencyType = CargoType,
+      { dependencyType = RPMType,
         dependencyName = rpmDepName pkg,
         dependencyVersion = rpmConstraint pkg,
         dependencyLocations = [],
@@ -94,9 +90,11 @@ toDependency pkg =
 buildGraph :: Dependencies -> Graphing Dependency
 buildGraph Dependencies {..} = run . withLabeling toDependency $ do
   traverse_ direct depBuildRequires
-  traverse (flip label $ RequiresType EnvDevelopment) depBuildRequires
+  traverse_ direct depRuntimeRequires
+  --
+  traverse_ (flip label $ RequiresType EnvDevelopment) depBuildRequires
+  traverse_ (flip label $ RequiresType EnvProduction) depRuntimeRequires
 
--- buildConstraint tail = RPMConstraint version <$> comparator
 buildConstraint :: Text -> Maybe VerConstraint
 buildConstraint tail = constraint
   where
@@ -127,6 +125,7 @@ getTypeFromLine line = safeReq
     safeReq = if isSafeName pkgName then req else Nothing
     req = case header of
       "BuildRequires" -> Just . BuildRequires . RPMDependency pkgName $ buildConstraint rawConstraint
+      "Requires" -> Just . RuntimeRequires . RPMDependency pkgName $ buildConstraint rawConstraint
       _ -> Nothing
 
 buildDependencies :: [RequiresType] -> Dependencies
@@ -134,8 +133,8 @@ buildDependencies = foldr addDep blankDeps
   where
     addDep req deps = case req of
       BuildRequires dep -> deps {depBuildRequires = dep : depBuildRequires deps}
-    blankDeps = Dependencies []
+      RuntimeRequires dep -> deps {depRuntimeRequires = dep : depRuntimeRequires deps}
+    blankDeps = Dependencies [] []
 
 getSpecDeps :: Text -> Dependencies
 getSpecDeps = buildDependencies . mapMaybe getTypeFromLine . T.lines
--- getSpecDeps = buildDependencies . concatMap (parseLine . splitLine) . T.lines
--- a/src/Strategy/RPM.hs
+++ b/src/Strategy/RPM.hs
@@ -1,0 +1,132 @@
+module Strategy.RPM (
+  discover,
+  getSpecDeps,
+  getTypeFromLine,
+
+  RPMDependency(..),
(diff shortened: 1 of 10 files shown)


src/Strategy/RPM.hs
```diff
@@ -0,0 +1,132 @@
...
+buildDependencies :: [RequiresType] -> Dependencies
+buildDependencies = foldr addDep blankDeps
+  where
+    addDep req deps = case req of
+      BuildRequires dep -> deps { depBuildRequires = dep : depBuildRequires deps }
+    blankDeps = Dependencies []
+
+getSpecDeps :: Text -> Dependencies
+getSpecDeps = buildDependencies . mapMaybe getTypeFromLine . T.lines
+-- getSpecDeps = buildDependencies . concatMap (parseLine . splitLine) . T.lines
```
zlav, line 141: [nit] did you mean to leave this?

src/Strategy/RPM.hs
```diff
@@ -0,0 +1,132 @@
...
+    len = T.length needle
+    (head, tail) = T.breakOn needle haystack
+    strippedTail = T.drop len tail
+
+getTypeFromLine :: Text -> Maybe RequiresType
+getTypeFromLine line = safeReq
+  where
+    (header, value) = splitOnceOn ": " line
+    (pkgName, rawConstraint) = splitOnceOn " " $ T.strip value
+    isSafeName name = not $ "%{" `T.isInfixOf` name
```
zlav, line 126: [nit] I think I would prefer either not defining `isSafeName` and inlining the logic or adding a type signature for the function.

src/Strategy/RPM.hs
```diff
@@ -0,0 +1,132 @@
...
+    safeReq = if isSafeName pkgName then req else Nothing
+    req = case header of
+      "BuildRequires" -> Just . BuildRequires . RPMDependency pkgName $ buildConstraint rawConstraint
```
zlav, line 129: This seems to be another instance where we need a style guide since I'm not sure if we should be doing repeated `$`s or `.`s. I don't think thats too important right now but we should do that soon

src/Strategy/RPM.hs
```diff
@@ -0,0 +1,132 @@
...
+      _ -> Nothing
+
+buildDependencies :: [RequiresType] -> Dependencies
+buildDependencies = foldr addDep blankDeps
```
zlav, line 133: [nit] I find functions like this easier to read if they are not written in point free. Not a really strong opinion here though. Another thing for us to discuss in a style guide.

test/RPM/testdata/simple-test.spec
```diff
@@ -0,0 +1,30 @@
...
+BuildRequires: xz >> 1.2.3
+
+# Lots of spaces, should be fine
+BuildRequires:   libxml2     >=   5.4.98
+
+# Complex version text (shouldn't change behavior)
+BuildRequires:  gmock = 1.6.0-1.aapps.el7
+
+# Ignored due to macro use
+BuildRequires:  cloudistics-core%{?_isa} = %{version}-%{release}
```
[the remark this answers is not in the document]
scruffystuffs: Real world is slightly more complex, but even so, it's not really applicable here.  A common pattern is to have something like `package` with ISA-agnostic code, and `package-isa` with ISA-specific code.  There's a distinct possibly of giving out incorrect licenses since much of the ISA-specific code can be maintained and released by third-parties.  I don't know how common this is, but even with my limited experience with these packages, I've seen that pattern a few times.  Makes me think it's common enough.

src/Strategy/RPM.hs
```diff
@@ -0,0 +1,132 @@
...
+newtype SpecFileLabel =
+  RequiresType DepEnvironment
+  deriving (Eq, Ord, Show, Generic)
+  
+data RPMDependency = RPMDependency 
+  { rpmDepName :: Text,
+    rpmConstraint :: Maybe VerConstraint
+  } deriving (Eq, Ord, Show, Generic)
+
+{- These two structures should make it very easy to add other headers like `Requires`
```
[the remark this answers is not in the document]
scruffystuffs: Sure, I can do that.

src/Strategy/RPM.hs
```diff
@@ -0,0 +1,132 @@
...
+  case find (\f -> ".spec" `isSuffixOf` fileName f) files of
+    Nothing -> pure WalkContinue
+    Just specFile -> do
+      runSimpleStrategy "rpm-spec" RustGroup $ fmap (mkProjectClosure dir) (analyze specFile)
+      pure WalkSkipAll
+
+analyze :: (Has ReadFS sig m, Has (Error ReadFSErr) sig m) => Path Rel File -> m (Graphing Dependency)
+analyze specFile = do
+  specFileText <- readContentsText specFile
+  pure . buildGraph $ getSpecDeps specFileText
```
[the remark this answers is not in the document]
scruffystuffs: In this context, `pure . buildGraph` is exactly equivalent to `pure $ buildGraph`.  However, `buildGraph $ getSpecDeps specFileText` requires a `$`, since the RHS of the `$` is type `a`, rather than `a -> b` (i.e.: it's a value, rather than a function).  I prefer to use `.` as much as possible, since it's much smaller and visually stays out of the way of the names which I care a lot more about.

test/RPM/testdata/simple-test.spec
```diff
@@ -0,0 +1,30 @@
...
+# Simple version text
+BuildRequires: cmake >= 2.8
+BuildRequires: jsoncpp <= 0.10.5
+BuildRequires: gtest > 6.4
+BuildRequires: libevent < 9.8.1
+BuildRequires: libvirt == 3.9.0
+BuildRequires: openssl = 1.23.45
+
+# Bad operator, will ignore version
+BuildRequires: xz >> 1.2.3
```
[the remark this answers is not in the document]
scruffystuffs: It's a fake operator, `rpm` has no understanding of this operator.  I could have put something like `#%%*#` and performed the same test.

src/Strategy/RPM.hs
```diff
@@ -0,0 +1,132 @@
...
+mkProjectClosure :: Path Rel Dir ->  Graphing Dependency -> ProjectClosureBody
+mkProjectClosure dir graph = ProjectClosureBody
+  { bodyModuleDir    = dir
+  , bodyDependencies = dependencies
+  , bodyLicenses     = []
+  }
+  where
+  dependencies = ProjectDependencies
+    { dependenciesGraph    = graph
+    , dependenciesOptimal  = NotOptimal
```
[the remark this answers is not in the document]
scruffystuffs: I'm not sure about this, I wasn't sure when I wrote it either.  I think we should discuss what we mean by `Optimal` here, then we can decide how to go with this.

src/Strategy/RPM.hs
scruffystuffs: giphub

src/Strategy/RPM.hs
scruffystuffs: This is a great example of why I prefer `.` over `$`.  Neither is very obtrusive when you need it, but if you had to pick, `.` is way less noisy in a visual sense.

src/Strategy/RPM.hs
scruffystuffs: I disagree here.  `isSafeName` takes a name and tells you if it's safe.  My instinct here would assume a `Text -> Bool`, which is exactly what it is (if that wasn't it, I would assume a newtype wrapper around a `Text` or other `String`-like).  Inlining decreases readability for sure, and I think a type sig is pretty unnecessary.

src/Strategy/RPM.hs
scruffystuffs: I think with complex type sigs or tons of internal logic, it can be pretty tough to read.  When you have a single monadic parameter, I think it reduces noise slightly, especially with collections like `List`.  The type sig tells you what you need, and the one-liner is the logic.  Without the where clause, point-free would be super confusing, I'd agree then.

src/Strategy/RPM.hs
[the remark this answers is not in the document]
scruffystuffs: Actually, this brings up an interesting point.  Should `BuildRequires` and `Requires` both be treated as production deps, or should `BuildRequires` be treated as a dev dependency?  Currently it's dev.

src/Strategy/RPM.hs
```diff
@@ -0,0 +1,140 @@
...
+data Dependencies
+  = Dependencies
+      { depBuildRequires :: [RPMDependency],
+        depRuntimeRequires :: [RPMDependency]
+      }
+  deriving (Eq, Ord, Show, Generic)
+
+discover :: HasDiscover sig m => Path Abs Dir -> m ()
+discover = walk $ \dir _ files ->
+  case find (\f -> ".spec" `isSuffixOf` fileName f) files of
```
cnr, line 48: Can there be more than one `.spec` file in a directory? if so, we'll want to make this a `filter` instead (and use something like `traverse` or `sequence` below to run the strategy on all of the files found)

src/Strategy/RPM.hs
```diff
@@ -0,0 +1,140 @@
...
+      "<=" -> Just $ CLessOrEq version
+      "<" -> Just $ CLess version
+      ">=" -> Just $ CGreaterOrEq version
+      ">" -> Just $ CGreater version
+      "=" -> Just $ CEq version
+      "==" -> Just $ CEq version
+      _ -> Nothing
+
+splitOnceOn :: Text -> Text -> (Text, Text)
+splitOnceOn needle haystack = (head, strippedTail)
```
cnr, line 113: [no action] I hate that this isn't in `base`; it's super useful

src/Strategy/RPM.hs
cnr: Type signatures on all of the non-`splitOnceOn` bindings would be nice. It took me a while to mentally map the types (especially on `req`)

src/Strategy/RPM.hs
cnr: hlint would warn here if the variable was present -- this is a pretty typical form found elsewhere in the codebase, e.g., in `toDependency` above.

Something that would help here, though, are type signatures on at least `addDep`

cnr: hlint would warn here if the variable was present -- this is a pretty typical form found elsewhere in the codebase, e.g., in `toDependency` above.

Something that would help here, though, are type signatures on at least `addDep`

scruffystuffs: Good counter example to point-free is on the `isSafeName` in the `where` clause of `getTypeFromLine`.  Point-free would be more confusing there, even though it's easily done.

src/Strategy/RPM.hs
scruffystuffs: I was pissed when I didn't find it.  But the `splitOn` function also seems to be partial, which is confusing and unexpected.  Weird shit.

src/Strategy/RPM.hs
scruffystuffs: If the other bindings need it, then consistency dictates that `isSafeName` should also get it.

src/Strategy/RPM.hs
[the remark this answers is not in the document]
zlav: (Chatted offline) We should not tag `BuildRequires` anything and omit `Requires` from the dependency graph until we have a need for including it. Right now we wanted to make sure that we also parse `Requires` to prove that it can be easily added in the future if needed.

src/Strategy/RPM.hs
```diff
@@ -0,0 +1,137 @@
...
+      { dependencyType = RPMType,
+        dependencyName = rpmDepName pkg,
+        dependencyVersion = rpmConstraint pkg,
+        dependencyLocations = [],
+        dependencyEnvironments = [],
+        dependencyTags = M.empty
+      }
+
+buildGraph :: Dependencies -> Graphing Dependency
+buildGraph Dependencies {..} = run . withLabeling toDependency $ traverse direct depBuildRequires
```
scruffystuffs, line 86: @cnr Is there a better way to run the grapher if you aren't using labels?

cnr: You can:
1. `evalGrapher`, which will give you `m (Graphing RPMDependency)`
2. Use `gmap :: (Ord ty, Ord ty') => (ty -> ty') -> Graphing ty -> Graphing ty'` to change the type of elements in the `Graphing`

cnr: Oh another option here for building the graph: because you just have a list of direct dependencies, `Graphing.fromList :: Ord a => [a] -> Graphing a` is probably what you want

scruffystuffs: I originally tried `evalGrapher`, but I missed `gmap`, so I didn't get it to work.  `fromList` is definitely the move here, although I don't know if that will work if/when we include `Requires` dependencies.

src/Strategy/RPM.hs
```diff
@@ -0,0 +1,137 @@
...
+      }
+  deriving (Eq, Ord, Show, Generic)
+
+discover :: HasDiscover sig m => Path Abs Dir -> m ()
+discover = walk $ \dir _ files ->
+  let specs = find (\f -> ".spec" `isSuffixOf` fileName f) files
+      analzyeFile specFile = runSimpleStrategy "rpm-spec" RPMGroup $ fmap (mkProjectClosure dir) (analyze specFile)
+   in traverse_ analzyeFile specs >> pure WalkSkipAll
+
+analyze :: (Has ReadFS sig m, Has (Error ReadFSErr) sig m) => Path Rel File -> m (Graphing Dependency)
```
scruffystuffs, line 52: @cnr Should I be using diagnostics here?

cnr: Yes -- should just require importing `Control.Effect.Diagnostics`, and changing `Has (Error ReadFSErr) sig m` to `Has Diagnostics sig m`

PR #80, opened 2020-06-17
AOSP graph scan

spatten:
Scan a Ninja build to get the dependency graph and post the graph to Scotland Yard.

I did this by adding a `ninja-graph` command to the `vpscli` binary.

As part of this, I also pulled `validateDir` and `parseUri` out into a Utils.hs module, as they were both being used in multiple places.

This is complete for the happy path -- it works for targets where the first dependency is the input. More work will be incoming to fix some of the cases where this is not true, but this is getting quite large so I thought it should go through a review as soon as possible.

The android dependency file will look something like this:

```
============================================
PLATFORM_VERSION_CODENAME=REL
...
============================================
ninja: no work to do.
...
Starting ninja...
out/target/product/coral/obj/STATIC_LIBRARIES/libgptutils_intermediates/gpt-utils.o: #deps 2, deps mtime 1583962189 (VALID)
    device/google/coral/gpt-utils/gpt-utils.cpp
    external/libcxx/include/stdio.h
    external/libcxx/include/__config

out/target/product/coral/obj/APPS/Settings_intermediates/package.apk: #deps 0, deps mtime 1583991062 (VALID)

out/target/product/coral/obj/JAVA_LIBRARIES/wifi-service_intermediates/dexpreopt.zip: #deps 2, deps mtime 1583991124 (VALID)
    out/soong/host/linux-x86/bin/dex2oatd
    out/soong/host/linux-x86/bin/profman
    out/soong/host/linux-x86/bin/soong_zip

[0;32m#### build completed successfully (20 seconds) ####[00m
```

The interesting part is the part between "Starting ninja..." and "build completed successfully". Lines that start with no spaces are "targets". Lines that start with spaces are dependencies of the preceding target. In most cases the first dependency is pulled out to be the sole input for the target. The ninja-deps file above should result in a dependency graph that looks like this. It has three targets. The first and last target have two dependencies and one input. The middle target has no dependencies or inputs.

```JSON
[
  {
    "path": "out/target/product/coral/obj/STATIC_LIBRARIES/libgptutils_intermediates/gpt-utils.o",
    "inputs": [
      {
        "path": "device/google/coral/gpt-utils/gpt-utils.cpp",
        "componentName": null,
        "hasDependencies": false
      }
    ],
    "dependencies": [
      {
        "path": "external/libcxx/include/stdio.h",
        "componentName": null,
        "hasDependencies": false
      },
      {
        "path": "external/libcxx/include/__config",
        "componentName": null,
        "hasDependencies": false
      }
    ],
    "componentName": null
  },
  {
    "path": "out/target/product/coral/obj/APPS/Settings_intermediates/package.apk",
    "inputs": [],
    "dependencies": [],
    "componentName": null
  },
  {
    "path": "out/target/product/coral/obj/JAVA_LIBRARIES/wifi-service_intermediates/dexpreopt.zip",
    "inputs": [
      {
        "path": "out/soong/host/linux-x86/bin/dex2oatd",
        "componentName": null,
        "hasDependencies": true
      }
    ],
    "dependencies": [
      {
        "path": "out/soong/host/linux-x86/bin/profman",
        "componentName": null,
        "hasDependencies": true
      },
      {
        "path": "out/soong/host/linux-x86/bin/soong_zip",
        "componentName": null,
        "hasDependencies": true
      }
    ],
    "componentName": null
  }
]
```


src/App/VPSScan/Main.hs
```diff
@@ -27,6 +28,13 @@ vpsOpts = VPSOpts <$> runSherlockOpts <*> optional runIPROpts <*> syOpts <*> org
               projectIDOpt = strOption (long "project" <> metavar "String" <> help "Project ID")
               revisionIDOpt = strOption (long "revision" <> metavar "String" <> help "Revision ID")
 
+ninjaGraphOpts :: Parser NinjaGraphOpts
+ninjaGraphOpts = NinjaGraphOpts <$> ninjaDepsOpt <*> lunchTargetOpt <*> scotlandYardUrlOpt
+                 where
+                   ninjaDepsOpt = optional $ strOption (long "ninjadeps" <> metavar "STRING" <> help "Path to a generated ninja_deps file. This is only used to speed up testing and will not normally be used.")
```
zlav, line 34: Where are these comments output? If they are public when running `-h` then it may be worth it to see if we can hide these from public use since the first, `ninjadeps`, is only used for testing.

src/App/VPSScan/NinjaGraph.hs
```diff
@@ -0,0 +1,234 @@
...
+
+readNinjaDepsFile :: (Has Trace sig m, Has ReadFS sig m, Has Diagnostics sig m, MonadIO m) => FilePath -> m ByteString
+readNinjaDepsFile ninjaPath = do
+  trace $ "reading ninja deps from " ++ ninjaPath
+  path <- liftIO $ parseAbsFile ninjaPath
+  readContentsBS path
+
+generateNinjaDeps :: (Has Trace sig m, Has Diagnostics sig m, MonadIO m) => Path Abs Dir -> NinjaGraphOpts -> m BL.ByteString
+generateNinjaDeps baseDir NinjaGraphOpts{..} = do
+  trace $ "Generating ninja deps with this command: " ++ commandString
```
zlav, line 100: Is this `trace` meant to be left in? This seems like it would be helpful to the user if it was. I'm also not sure if `trace` is the ideal way for us to be logging, I don't have an alternative, I just don't recall using this in the past.

src/App/VPSScan/Main.hs
spatten: It looks you can just delete the `<> help "..."` part and nothing is output about that option. I'll make that change

src/App/VPSScan/NinjaGraph.hs
spatten: Yeah, I was planning on leaving it in.

I've been using `trace` all over the place in this code, and I think it was Connor's suggestion to use it. He's approved PRs that use it at the very least :)

PR #83, opened 2020-06-22
Use better types from the CLI handlers to the internals.

scruffystuffs:
Summary:

* Wrap the `Text` of the API key as early as possible, unwrap it as late as possible.
* Abstract building the Authorization header (where we actually need the api key text)
* Collapse project {name, revision, branch} by leveraging the inferred project and an override project, and merging them, using the resulting merge type everywhere else.

### DIFF
--- a/src/App/Fossa/FossaAPIV1.hs
+++ b/src/App/Fossa/FossaAPIV1.hs
@@ -1,7 +1,6 @@
 module App.Fossa.FossaAPIV1
   ( uploadAnalysis
   , UploadResponse(..)
-  , ProjectRevision(..)
   , ProjectMetadata(..)
   , FossaError(..)
   , FossaReq(..)
@@ -21,12 +20,13 @@ module App.Fossa.FossaAPIV1
   , getAttribution
   ) where
 
+import App.Fossa.CliTypes
 import App.Fossa.Analyze.Project
 import qualified App.Fossa.Report.Attribution as Attr
 import Control.Effect.Diagnostics
 import Data.Coerce (coerce)
 import Data.List (isInfixOf)
-import Data.Maybe (fromMaybe)
+import Data.Maybe (catMaybes, fromMaybe)
 import Data.Text (Text)
 import Data.Aeson
 import Prelude
@@ -42,7 +42,6 @@ import Srclib.Converter (toSourceUnit)
 import Srclib.Types
 import Text.URI (URI)
 import qualified Text.URI as URI
-import Data.Maybe (catMaybes)
 
 newtype FossaReq m a = FossaReq { unFossaReq :: m a }
   deriving (Functor, Applicative, Monad, MonadIO, Algebra sig)
@@ -74,6 +73,9 @@ renderLocatorUrl :: Int -> Locator -> Text
 renderLocatorUrl orgId Locator{..} =
   locatorFetcher <> "+" <> T.pack (show orgId) <> "/" <> locatorProject <> "$" <> fromMaybe "" locatorRevision
 
+apiHeader :: ApiKey -> Option scheme
+apiHeader key = header "Authorization" ("token " <> encodeUtf8 (unApiKey key))
+
 data UploadResponse = UploadResponse
   { uploadLocator :: Text
   , uploadError   :: Maybe Text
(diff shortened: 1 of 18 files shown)


src/App/Fossa/FossaAPIV1.hs
```diff
@@ -74,6 +73,9 @@ renderLocatorUrl :: Int -> Locator -> Text
 renderLocatorUrl orgId Locator{..} =
   locatorFetcher <> "+" <> T.pack (show orgId) <> "/" <> locatorProject <> "$" <> fromMaybe "" locatorRevision
 
+apiHeader :: ApiKey -> Option scheme
+apiHeader key = header "Authorization" ("token " <> encodeUtf8 (unApiKey key))
```
scruffystuffs, line 77: This seems like a safe abstraction at this point.

src/App/Fossa/Main.hs
```diff
@@ -27,38 +28,47 @@ appMain = do
 
   maybeApiKey <- checkAPIKey optAPIKey
 
+  let override = OverrideProject { overrideName = optProjectName, overrideRevision = optProjectRevision, overrideBranch = Nothing }
```
scruffystuffs, line 31: I don't like this being on one line.  But I was getting syntax errors trying to move it around. 🤷

src/App/Fossa/Main.hs
```diff
@@ -27,38 +28,47 @@ appMain = do
...
+requireKey :: Maybe ApiKey -> IO ApiKey
 requireKey (Just key) = pure key
 requireKey Nothing = die "A FOSSA API key is required to run this command"
 
 changeDir :: FilePath -> IO ()
 changeDir path = validateDir path >>= setCurrentDir
 
+infixr 1 <$$>
+(<$$>) :: (Functor f, Functor g) => (a -> b) -> f (g a) -> f (g b)
+(<$$>) = fmap . fmap
```
scruffystuffs, line 64: Should this be somewhere else?  Also, should we reconsider making it an operator vs naming it?

src/App/Fossa/ProjectInference.hs
```diff
@@ -24,9 +26,16 @@ import Text.GitConfig.Parser (Section (..), parseConfig)
 import Text.Megaparsec (errorBundlePretty)
 import Effect.ReadFS
 
+mergeOverride :: OverrideProject -> InferredProject -> ProjectRevision
+mergeOverride OverrideProject {..} InferredProject {..} = ProjectRevision name revision branch
+  where
+    name = fromMaybe inferredName overrideName
+    revision = fromMaybe inferredRevision overrideRevision
+    branch = fromMaybe inferredBranch overrideBranch
```
scruffystuffs, line 34: @cnr There are cases where we have the branch, but never actually use it.  Does this break the laziness we had by previously not having access to it at all?

src/App/Fossa/Main.hs
cnr: you can make it multi-line if you indent the `{` (or whatever else you're breaking on) past the `o` in `override`:
```hs
let override = OverrideProject
      { overrideName = optProjectName
      , overrideRevision = optProjectRevision
      , overrideBranch = Nothing
      }
```

and then running ormolu, haha

src/App/Fossa/Main.hs
cnr: This is fine here -- also this is a nit: it should be `infixl 4` like `<$>` (see: `:info (<$>)` in ghci)

src/App/Fossa/Main.hs
```diff
@@ -27,38 +28,47 @@ appMain = do
...
+
 -- | Try to fetch FOSSA_API_KEY from env if not supplied from cmdline
-checkAPIKey :: Maybe Text -> IO (Maybe Text)
-checkAPIKey Nothing = fmap T.pack <$> lookupEnv "FOSSA_API_KEY"
-checkAPIKey key = return key
+checkAPIKey :: Maybe Text -> IO (Maybe ApiKey)
+checkAPIKey key = case key of
+  Just key' -> pure . Just $ ApiKey key'
+  Nothing -> ApiKey <$$> T.pack <$$> lookupEnv "FOSSA_API_KEY"
```
cnr, line 70: this will have to change slightly with `infixl 4`, but uniformity with `<$>` would help readability

scruffystuffs: It took me way too many tries to figure out the left side becomes a pure pipeline applied to the functor.   I originally tried to do this, but I never figured it out.

PR #84, opened 2020-06-23
Deduplicate the build/issue flow

scruffystuffs:
When `fossa report` was created, there was a development time constraint.  To meet this constraint, some of the logic from `App.Fossa.Test` was duplicated to `App.Fossa.Report`.  The two biggest copies were `waitForBuild` and `waitForIssues`, which can be easily abstracted.

Marked as draft until #83 is merged, because we're losing some type refactoring.

### DIFF
--- a/src/App/Fossa/BuildWait.hs
+++ b/src/App/Fossa/BuildWait.hs
@@ -1,0 +1,53 @@
+module App.Fossa.BuildWait (
+    waitForBuild,
+    waitForIssues
+) where
+
+import Prologue
+
+import App.Fossa.CliTypes
+import qualified App.Fossa.FossaAPIV1 as Fossa
+import Control.Carrier.Diagnostics
+import Control.Concurrent (threadDelay)
+import Effect.Logger
+import Text.URI (URI)
+
+pollDelaySeconds :: Int
+pollDelaySeconds = 8
+
+data WaitError = BuildFailed -- ^ we encountered the FAILED status on a build
+  deriving (Show, Generic)
+
+instance ToDiagnostic WaitError where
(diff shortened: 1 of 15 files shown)


src/App/Fossa/BuildWait.hs
```diff
@@ -0,0 +1,53 @@
+module App.Fossa.BuildWait (
```
cnr, line 1: [nit] maybe move this to `App.Fossa.API.BuildWait`?

src/App/Fossa/Report.hs
```diff
@@ -83,48 +85,9 @@ reportMain baseUri apiKey logSeverity timeoutSeconds reportType override = do
...
-timeout seconds act = either id id <$> Async.race (Just <$> act) (threadDelay (seconds * 1000000) *> pure Nothing)
-
-data WaitError
-  = BuildFailed -- ^ we encountered the FAILED status on a build
-  deriving (Show, Generic)
-
-instance ToDiagnostic WaitError where
-  renderDiagnostic BuildFailed = "The build failed. Check the FOSSA webapp for more details."
+-- timeout seconds act = either id id <$> Async.race (Just <$> act) (threadDelay (seconds * 1_000_000) *> pure Nothing)
+timeout seconds act = either id id <$> Async.race (Just <$> act) (threadDelay (seconds * 1_000_000) $> Nothing)
```
cnr, line 93: great call on `-XNumericUnderscores`. and I was questioning whether to use `$>` here when initially writing this; happy to see it here now.

src/App/Fossa/BuildWait.hs
scruffystuffs: Yep, that's better.

PR #85, opened 2020-06-25
Include orphan poms in the global closure graph; match on ".pom" file extensions the same as ".pom.xml"

cnr:
If a pom isn't related to any other poms, it wasn't included in the global closure graph -- and as such, it wouldn't be included as a project

### DIFF
--- a/src/Strategy/Maven/Pom/Closure.hs
+++ b/src/Strategy/Maven/Pom/Closure.hs
@@ -29,9 +29,8 @@ findPomFiles :: (Has ReadFS sig m, MonadIO m) => Path Abs Dir -> m [Path Abs Fil
 findPomFiles dir = do
   relPaths <- execState @[Path Rel File] [] $
     flip walk dir $ \_ _ files -> do
-      case find (\file -> fileName file == "pom.xml" || ".pom" `isSuffixOf` fileName file) files of
-        Just file -> modify @[Path Rel File] (file:)
-        Nothing -> pure ()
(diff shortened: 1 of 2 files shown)


microsoftly:
I tested with the original commit; the results below are for the `6b513e7` commit.

There are 5 files each with the same content that we expect to be hit for decl licenses
* `.pom` [match] (I wouldn't expect this to match, but I guess it's not the end of the world if it did) 
* `pom` [no match] (this didn't match and we don't expect it to)
* `pom.xml` [match]
* `anythingBefore.pom` [match]
* `anythingBefore.pom.xml` [no match]

It looks like we still have 1, maybe 2 edge cases.

src/Strategy/Maven/Pom/Closure.hs
```diff
@@ -29,7 +29,7 @@ findPomFiles :: (Has ReadFS sig m, MonadIO m) => Path Abs Dir -> m [Path Abs Fil
 findPomFiles dir = do
   relPaths <- execState @[Path Rel File] [] $
     flip walk dir $ \_ _ files -> do
-      case find ((== "pom.xml") . fileName) files of
+      case find (\file -> fileName file == "pom.xml" || ".pom" `isSuffixOf` fileName file) files of
```
microsoftly, line 32: I think both checks should use `isSuffixOf`. I think that's what's preventing the `anythingBefore.pom.xml` from being picked up

cnr:
I think `.pom` is fine as a false-positive -- changed the `pom.xml` case to a suffix check

PR #86, opened 2020-06-25
Archive unpacking

cnr:
This adds `--unpack-archives` as a command switch to `fossa analyze`. Discovered `.tar`, `.tar.gz`, `.zip`, `.jar`, and `.rpm` files will be recursively unarchived and analyzed.

This required using absolute filepaths in all strategies, so there are a bunch of "unrelated" changes in this PR. This is required because archives are unpacked in a temporary directory, and the current working directory is process-global -- so reading relative files/directories won't work if we want to run things in parallel.

### DIFF
--- a/src/App/Fossa/Main.hs
+++ b/src/App/Fossa/Main.hs
@@ -31,10 +31,10 @@ appMain = do
     AnalyzeCommand AnalyzeOptions {..} -> do
       changeDir analyzeBaseDir
       if analyzeOutput
-        then analyzeMain logSeverity OutputStdout optProjectName optProjectRevision analyzeBranch
+        then analyzeMain logSeverity OutputStdout optProjectName optProjectRevision analyzeBranch analyzeUnpackArchives
         else do
           key <- requireKey maybeApiKey
-          analyzeMain logSeverity (UploadScan optBaseUrl key analyzeMetadata) optProjectName optProjectRevision analyzeBranch
+          analyzeMain logSeverity (UploadScan optBaseUrl key analyzeMetadata) optProjectName optProjectRevision analyzeBranch analyzeUnpackArchives
     TestCommand TestOptions {..} -> do
       changeDir testBaseDir
       key <- requireKey maybeApiKey
@@ -117,6 +117,7 @@ analyzeOpts :: Parser AnalyzeOptions
 analyzeOpts =
   AnalyzeOptions
     <$> switch (long "output" <> short 'o' <> help "Output results to stdout instead of uploading to fossa")
+    <*> switch (long "unpack-archives" <> help "Recursively unpack and analyze discovered archives")
     <*> optional (strOption (long "branch" <> help "this repository's current branch (default: current VCS branch)"))
     <*> metadataOpts
     <*> baseDirArg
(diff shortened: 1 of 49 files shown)


src/Control/Effect/Finally.hs
```diff
@@ -0,0 +1,18 @@
+module Control.Effect.Finally
```
cnr, line 1: We add a new effect here -- `Finally` -- that has one operation:
```hs
onExit :: Has Finally sig m => m a -> m ()
```

This is used to clean up unpacked archive directories on program exit. We'd normally eagerly clean up these directories after discovering and running strategies, but we don't have a way to tell when the recursively-discovered strategies complete.

src/Strategy/Archive/RPM.hs
```diff
@@ -0,0 +1,41 @@
...
+import qualified Data.CPIO as CPIO
+import Path
+import Prelude
+import qualified Path.IO as PIO
+
+extractRpm :: Has (Lift IO) sig m => Path Abs Dir -> Path Abs File -> m ()
+extractRpm dir rpmFile = do
+  res <-
+    sendIO . runResourceT . runExceptT . runConduit $
+      readRPMEntries rpmFile .| filterC (not . CPIO.isEntryDirectory) .| sinkDir
```
cnr, line 23: This uses a streaming library called `conduit` -- you can think of the `.|` operators as "unix pipes".

As for how to read the conduit type signatures: `ConduitT i o m a` has:

- "streaming input type" `i`
- "streaming output type" `o`
- underlying monad `m` (it's a monad transformer)
- result type `a` (like any other monad)

src/Control/Effect/Finally.hs
scruffystuffs: Bummer.  Staggering those potentially large IO ops could be a lot more performant.  What's the exact reason we don't know when they're done?

cnr: The discover actions don't keep track of the strategies they find -- they just fork the strategy as a new task to be run in parallel and forget about it. It's partially a consequence of how aggressively parallel we are

src/App/Fossa/Main.hs
```diff
@@ -117,6 +117,7 @@ analyzeOpts :: Parser AnalyzeOptions
 analyzeOpts =
   AnalyzeOptions
     <$> switch (long "output" <> short 'o' <> help "Output results to stdout instead of uploading to fossa")
+    <*> switch (long "unpack-archives" <> help "Recursively unpack and analyze discovered archives")
```
zlav, line 120: I like the name `unpack-archives`, I think it is good to be explicit here.

src/Strategy/Archive.hs
```diff
@@ -0,0 +1,104 @@
...
+  (Has (Lift IO) sig m, Has Finally sig m) =>
+  -- | Archive extraction function
+  (Path Abs Dir -> Path Abs File -> m ()) ->
+  -- | Path to archive
+  Path Abs File ->
+  -- | Callback
+  (Path Abs Dir -> m ()) ->
+  m ()
+withArchive extract file go = do
+  tmpDir <- mkTempDir (fileName file)
```
zlav, line 67: Does this not cause a collision between the filename and the newly created temp directory? I would think that we have to strip the file extension before creating the directory?

cnr: The temp directory is created as a _system_-level temp directory (like in `/tmp`), and the name provided serves as a "template" or prefix that'll be used when creating the system temp directory.

If a folder with that name already exists, it'll arbitrarily append random garbage to the end of it

PR #87, opened 2020-06-26
Scope MavenProjectClosure graphs to only include project-local poms

cnr:
MavenProjectClosures used to contain the global pom graph, including pom files unrelated to the specific maven project. 

Pathfinder extracts licenses from every pom in a MavenProjectClosure. Given N projects, every pom from the filetree would appear N times in the output.

This PR removes unrelated poms from individual MavenProjectClosures.


cnr:
(cc'd alex and kit for visibility)

cnr:
Definitely a needs-to-be-done-quickly thing; I'll add tests for this later during the planned tests overhaul

PR #88, opened 2020-06-30
IPR file upload

spatten:
Make `vpscli scan` upload all of the files in `basedir` to S3 or S3 equivalent (i.e. Minio) once IPR and Sherlock scans are done.


src/App/VPSScan/Scan.hs
```diff
@@ -36,37 +37,49 @@ scanMain opts@ScanCmdOpts{..} = do
...
-      trace "[IPR] Failed to scan"
-      trace (show $ renderFailureBundle iprFailure)
+      liftIO $ print $ renderFailureBundle iprFailure
       liftIO exitFailure
     (_, Left sherlockFailure) -> do
-      trace "[Sherlock] Failed to scan"
-      trace (show $ renderFailureBundle sherlockFailure)
+      liftIO $ print $ renderFailureBundle sherlockFailure
       liftIO exitFailure
+    (_, _) ->
```
cnr, line 73: [nit] this is the same as just `_`

src/App/VPSScan/Scan.hs
```diff
@@ -93,4 +106,16 @@ runIPRScan basedir scanId vpsOpts@VPSOpts{..} =
...
-      trace "[IPR] IPR Scan disabled"
\ No newline at end of file
+      trace "[IPR] IPR Scan disabled"
+
+runS3Upload ::
+  ( Has Diagnostics sig m
+  , Has Trace sig m
+  , MonadIO m
+  ) => Path Abs Dir -> Text -> VPSOpts -> String -> m ()
+runS3Upload basedir scanId VPSOpts{..} bucket =
```
cnr, line 116: [nit] can we pass IPROpts in here directly? It seems that the caller already does the case split

src/App/VPSScan/Scan.hs
cnr: also - because `exitFailure` will just exit immediately, we don't have to nest these case expressions - we can use a new top-level case

src/App/VPSScan/Scan/RunS3Upload.hs
```diff
@@ -0,0 +1,109 @@
...
+  deriving (Eq, Ord, Show, Generic, Typeable)
+
+instance ToDiagnostic S3UploadError where
+  renderDiagnostic = \case
+    ErrorRunningS3Upload err -> "Error while uploading to S3: " <> pretty err
+
+-- S3.S3Configuration is defined here:
+-- https://github.com/aristidb/aws/blob/9ccfce8c50723e1b7e02ba59a9d38227e938ac0a/Aws/S3/Core.hs#L69
+s3config :: IPROpts -> S3.S3Configuration Aws.NormalQuery
+s3config IPROpts{..} =
```
cnr, line 39: [nit] it's hard to tell what fields you're using here from IPROpts -- if it's just the `URI`, can we make this call out to a new top-level function that has a simpler type? `s3configFromURI :: URI -> S3.S3Configuration Aws.NormalQuery` or similar

spatten:
We're going to be using a different method to show these files to customers, so I'm just going to close this.

PR #89, opened 2020-06-30
Ninja graph "leading .txt dependencies should not be the input" fix

spatten:
Set the input file correctly for cases where the first N dependencies are .txt files and do not match the basename of the target, and the N+1th dependency is a non-.txt file and matches the basename of the target. In this case, the N+1th dependency is the correct input file

E.g., in this case we want the input to be "system/bpf/bpfloader/BpfLoader.cpp":

```
out/soong/.intermediates/system/bpf/bpfloader/bpfloader/android_arm64_armv8-a_core/obj/system/bpf/bpfloader/BpfLoader.o: #deps 3, deps mtime 1583962500 (VALID)
     external/compiler-rt/lib/cfi/cfi_blacklist.txt
     build/soong/cc/config/integer_overflow_blacklist.txt
     system/bpf/bpfloader/BpfLoader.cpp
     bionic/libc/include/arpa/inet.h
```


src/App/VPSScan/NinjaGraph.hs
```diff
@@ -107,15 +108,48 @@ generateNinjaDeps baseDir NinjaGraphOpts{..} = do
       Nothing -> "cd " ++ show baseDir ++ " && NINJA_ARGS=\"-t deps\" make"
       Just lunch ->  "cd " ++ show baseDir ++ " && source ./build/envsetup.sh && lunch " ++ T.unpack lunch ++ " && NINJA_ARGS=\"-t deps\" make"
 
-addInputsToNinjaDeps :: [DepsTarget] -> [DepsTarget]
-addInputsToNinjaDeps = map addInputToTarget
+correctedTarget :: DepsTarget -> DepsTarget
+correctedTarget target@DepsTarget { targetDependencies = [] } =
```
zlav, line 112: I haven't seen this syntax before, what does this do? `target@DepsTarget`

spatten: It does the same thing as just having everything after the `@` would, but also assigns the argument to a variable named `target`.

spatten:
Set the input file correctly for cases where the first N dependencies are .txt files and do not match the basename of the target, and the N+1th dependency is a non-.txt file and matches the basename of the target. In this case, the N+1th dependency is the correct input file

E.g., in this case we want the input to be "system/bpf/bpfloader/BpfLoader.cpp":

```
out/soong/.intermediates/system/bpf/bpfloader/bpfloader/android_arm64_armv8-a_core/obj/system/bpf/bpfloader/BpfLoader.o: #deps 3, deps mtime 1583962500 (VALID)
     external/compiler-rt/lib/cfi/cfi_blacklist.txt
     build/soong/cc/config/integer_overflow_blacklist.txt
     system/bpf/bpfloader/BpfLoader.cpp
     bionic/libc/include/arpa/inet.h
```

PR #91, opened 2020-07-02
Create draft releases

scruffystuffs:
There are edge cases where we do not want to publish a release, or may
want to create a pre-release instead.  This adds a tiny overhead to
release creation, but prevents a few potential accidents.

Example we could have run into: we needed to push a spot fix, but doing so required including some untested changes.
If we forgot to mark it as a pre-release, sales eng or scripting customers could have pulled an untested
version and potentially broken their builds.  This was very unlikely here, and is even less likely in the future.
That said, safety first.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -134,5 +134,6 @@ jobs:
       uses: softprops/action-gh-release@v1
       with:
         files: release/*
+        draft: true
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


zlav:
I'm not sure I find this necessary, anytime we cut a release the release should have been fully tested already right? I know this is a convention but I don't entirely understand why pre-releases exist. Cutting a release involves explicitly pushing a tag and I'm not sure I see a workflow where my next step wouldn't be unchecking the pre-release box (this is what I used to do with v1).

scruffystuffs:
This doesn't force us to use pre-releases.  It prevents publishing the release until we approve it.  Most of the time, we'll approve it right away.  But, on the odd hotfix case like thursday, we may need to mark it as pre-release.  This prevents an easy, forgettable mistake of not setting the pre-release flag.

zlav:
I'm still not sure I agree with there ever being a reason to mark a release as pre-release for this repository. Approving a PR should be the same as approving a release including that PR in my opinion.

scruffystuffs:
I don't think it makes sense to give the same level of test attention to a single PR as a "monthly" (?) release.  When we can't afford that level of testing, but need to make a change, cutting a pre-release is easy and practical.  Otherwise, we're possibly introducing a bug for anyone who blindly pulls our latest version.  Additionally, the drawback of using drafts is absolutely tiny, and can save us from having to put out a fire later.

zlav:
I don't think I would be as annoying here if this wasn't something I explicitly removed from CLI v1s release process due to the number of times people forgot to unmark releases as pre-releases. This adds another step in the releasing process, albeit a very small one.

> When we can't afford that level of testing, but need to make a change, cutting a pre-release is easy and practical

Why do you need to cut a release in this scenario instead of leaving the code on master?

scruffystuffs:
To be crystal clear: We are not marking everything as a pre-release.  Nothing about this forces us to use pre-releases which are promoted to normal releases.

As for why we cut a release, master does not provide pre-built binaries.

scruffystuffs:
Also, consider the post-mortems of each side of the argument here.
Without drafts: we released something we shouldn't have released, and now we have to **put out a fire**.
With drafts: we forgot to release something we intended to release, and now we have to **click a button**.

In terms of safety, there's a really clear winner here.

zlav:
I believe that master does have binaries after the tests finish. This page wasn't immediately clear to me but is there something that this list of artifacts doesn't contain that you're looking for? https://github.com/fossas/spectrometer/actions/runs/159928135

I think I'm misunderstanding the point about this not forcing us to use pre-releases. With this change, every new release will be initially listed as a pre-release correct?

PR #94, opened 2020-07-07
Implement contributor tracking

scruffystuffs:
Basic outline:

- If upload analysis has been successful (meaning we successfully called the API, not that we didn't receive any errors), try to upload contributor tracking data.
- Data takes the shape of `{"<authorEmail>": "<latest contribution date: YYYY-MM-DD>", ...}`
- Once contributor tracking phase is started, we do not fail the analysis or process for any reason within our control, including errors.
- Errors are reported as debug messages, but otherwise ignored.  This includes all three instances of `IO`:
  - Retrieving the current date
  - Invoking and reading the `git log ...` output
  - Uploading the contribution data to the core endpoint
- Malformed output lines from `git log ...` are ignored, since git should be taking care of that.  We can add protection here later if needed.
- No response is needed from the API, but we may want to check for an error message.  We can add this once the core endpoint has been implemented.

### DIFF
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -11,15 +11,22 @@ import qualified Control.Carrier.Diagnostics as Diag
 import Control.Carrier.Output.IO
 import Control.Concurrent
 
+import App.Fossa.CliTypes
+import App.Fossa.FossaAPIV1 (ProjectMetadata, uploadAnalysis, UploadResponse(..), Contributors(..), uploadContributors)
 import App.Fossa.Analyze.Project (Project, mkProjects)
-import App.Fossa.FossaAPIV1 (ProjectMetadata, uploadAnalysis, UploadResponse(..))
 import App.Fossa.ProjectInference (mergeOverride, inferProject)
 import App.Types
 import Control.Carrier.Finally
 import Control.Carrier.TaskPool
+import Data.ByteString.Lazy (toStrict)
+import qualified Data.Map as M
+import qualified Data.Text as T
 import Data.Text.Lazy.Encoding (decodeUtf8)
 import Data.Text.Prettyprint.Doc
 import Data.Text.Prettyprint.Doc.Render.Terminal
+import Data.Time
+import Data.Time.Format.ISO8601 (iso8601Show)
+import Effect.Exec
 import Effect.Logger
 import Network.HTTP.Types (urlEncode)
 import qualified Srclib.Converter as Srclib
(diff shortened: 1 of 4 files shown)


src/App/Fossa/Analyze.hs
```diff
@@ -123,6 +140,60 @@ analyze basedir destination override unpackArchives = runFinally $ do
...
+fetchGitContributors :: 
+  ( Has Diag.Diagnostics sig m
+  , Has Exec sig m
+  , MonadIO m
+  ) => Path x Dir -> m Contributors
+fetchGitContributors basedir = do
+  now <- liftIO getCurrentTime
+  rawContrib <- execThrow basedir $ gitLogCmd now
+  textContrib <- Diag.fromEitherShow . TE.decodeUtf8' $ toStrict rawContrib
+  pure . Contributors . M.map (T.pack . iso8601Show) $ foldr updateMap M.empty $ T.lines textContrib
```
cnr, line 171: This is a really good imperative way of thinking about the problem. I think we can use these functions to replace that `foldr` make it a little more functional (and also eliminate hacks like `minimumDay`):
```hs
M.fromListWith :: Ord k => (a -> a -> a) -> [(k,a)] -> Map k a
-- from Data.Maybe
mapMaybe :: (a -> Maybe b) -> [a] -> [b]
```

src/App/Fossa/Analyze.hs
```diff
@@ -123,6 +141,53 @@ analyze basedir destination override unpackArchives = runFinally $ do
...
+  ) => Path x Dir -> m Contributors
+fetchGitContributors basedir = do
+  now <- liftIO getCurrentTime
+  rawContrib <- execThrow basedir $ gitLogCmd now
+  textContrib <- Diag.fromEitherShow . TE.decodeUtf8' $ toStrict rawContrib
+  pure . Contributors
+    . M.map (T.pack . iso8601Show)
+    . M.fromListWith max
+    . mapMaybe readLine
+    $ T.lines textContrib
```
scruffystuffs, line 176: This line got too long for me.  ormolu suggested this style, so I went with it.

src/App/Fossa/Analyze.hs
```diff
@@ -123,6 +141,53 @@ analyze basedir destination override unpackArchives = runFinally $ do
...
+  -> URI
+  -> ApiKey
+  -> Text -- ^ Locator
+  -> m ()
+tryUploadContributors baseDir baseUrl apiKey locator = do
+  contributors <- fetchGitContributors baseDir
+  uploadContributors baseUrl apiKey locator contributors
+
+
+fetchGitContributors ::
```
cnr, line 163: [nit] can we move this to another module? (along with `gitLogCmd`, of course)

scruffystuffs: Have a name/existing module in mind?

cnr: I'd be okay with a new module named `VCS.Git` or similar

scruffystuffs:
@cnr would you mind rechecking?  I fixed some merge conflicts and I'd like another set of eyes making sure I didn't miss something obvious.

PR #96, opened 2020-07-08
Abort TaskPool when an exception is thrown in any task

cnr:
Tested with the following snippet:

```hs
example :: Has (Lift IO) sig m => m ()
example = withTaskPool 5 (sendIO . print) $ do
  forkTask . sendIO $ fail "oh no"
  forkTask . sendIO $ threadDelay maxBound
  forkTask go
    where
      go = do
        sendIO $ threadDelay 2000000
        forkTask go
```
i.e. forking three threads:
- one that throws an exception
- one that sleeps forever
- one that loops and creates a new task every two seconds (not really necessary to prove correctness here)

### DIFF
--- a/src/Control/Carrier/TaskPool.hs
+++ b/src/Control/Carrier/TaskPool.hs
@@ -13,20 +13,22 @@ import Control.Carrier.Reader
 import Control.Carrier.Threaded
 import Control.Concurrent.STM
 import Prologue
+import Control.Effect.Lift (sendIO)
 
-withTaskPool :: (Has (Lift IO) sig m, MonadIO m)
+withTaskPool :: Has (Lift IO) sig m
   => Int -- number of workers
   -> (Progress -> m ()) -- get progress updates
   -> TaskPoolC m a
   -> m ()
 withTaskPool numWorkers reportProgress act = do
-  state <- liftIO $
+  state <- sendIO $
     State <$> newTVarIO []
           <*> newTVarIO 0
           <*> newTVarIO 0
+          <*> newEmptyTMVarIO
 
-  let enqueue action = liftIO $ atomically $ modifyTVar (stQueued state) (action:)
-  liftIO $ (enqueue (void (runTaskPool (stQueued state) act)))
+  let enqueue action = sendIO $ atomically $ modifyTVar (stQueued state) (action:)
+  sendIO $ (enqueue (void (runTaskPool (stQueued state) act)))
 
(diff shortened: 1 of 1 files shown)


src/Control/Carrier/TaskPool.hs
```diff
@@ -13,20 +13,22 @@ import Control.Carrier.Reader
 import Control.Carrier.Threaded
 import Control.Concurrent.STM
 import Prologue
+import Control.Effect.Lift (sendIO)
 
-withTaskPool :: (Has (Lift IO) sig m, MonadIO m)
+withTaskPool :: Has (Lift IO) sig m
```
cnr, line 18: switched all of the uses of `MonadIO`/`liftIO` here to `Has (Lift IO)`/`sendIO`, because MonadIO is redundant with Lift IO

src/Control/Carrier/TaskPool.hs
```diff
@@ -101,16 +114,17 @@ data Progress = Progress
...
-      liftIO $ atomically $ modifyTVar' var (runReader var (runTaskPoolC (void (hdl (m <$ ctx)))):)
+      sendIO $ atomically $ modifyTVar' var (runReader var (runTaskPoolC (void (hdl (m <$ ctx)))):)
       pure ctx
     R other -> alg (runTaskPoolC . hdl) (R other) ctx
 
 data State m = State
   { stQueued    :: TVar [m ()]
   , stRunning   :: TVar Int
   , stCompleted :: TVar Int
+  , stError     :: TMVar SomeException
```
cnr, line 129: This is the trick: we have a mutable variable that's either empty or it contains a single exception. tasks insert an exception when one is thrown

src/Control/Carrier/TaskPool.hs
```diff
@@ -69,25 +76,31 @@ stopWhenDone State{..} act = do
...
     queued <- readTVar stQueued
     case queued of
       [] -> retry
       (x:xs) -> do
         writeTVar stQueued xs
         addRunning
-        -- FIXME: async exceptions should kill execution
-        pure $ x `finally` (complete *> loop)
+        pure $
+          mask $ \restore -> do
```
cnr, line 91: `mask` is confusing, but it essentially masks (ignores) exceptions in the inner block, which is useful for cleanup actions like filling the mutable variable. `restore` is used to "unmask"  exceptions for the actual task

PR #97, opened 2020-07-09
Include stdout in command failures

cnr:
We previously only included stderr in the error message


src/Effect/Exec.hs
```diff
@@ -72,18 +73,16 @@ data Exec m k where
   -- - a description of the command failure
   Exec :: Path x Dir -> Command -> Exec m (Either CmdFailure Stdout)
 
--- TODO: include info about CmdFailures as appropriate
 data ExecErr
   = -- | Command execution failed, usually from a non-zero exit. command, stderr
```
scruffystuffs, line 77: [nit] Comment should be updated here.  Maybe removed, it's pretty clear to me what this is w/o the comment.

PR #99, opened 2020-07-10
Chunked IPR uploads

spatten:
Upload IPR scan results in chunks of 1000 files (we're only uploading files with licenses), and post to the "IPR scan complete" endpoint once all IPR results have been posted

NOTE: This is dependent on the changes to SY in https://github.com/fossas/basis/pull/819

closes fossas/issues#190

### DIFF
--- a/src/App/VPSScan/Scan/ScotlandYard.hs
+++ b/src/App/VPSScan/Scan/ScotlandYard.hs
@@ -180,7 +180,7 @@ chunkJSON (Object obj) key chunkSize = do
     _ -> Nothing
   let chunked = chunksOf chunkSize $ V.toList arr
   let chunker :: [Value] -> Value
-      chunker v = object ["Files" .= v]
+      chunker v = object [key .= v]
   Just $ map chunker chunked
 
 chunkJSON _ _ _ = Nothing
--- a/src/App/VPSScan/Scan/ScotlandYard.hs
+++ b/src/App/VPSScan/Scan/ScotlandYard.hs
@@ -118,7 +118,7 @@ createBuildGraphEndpoint baseurl projectId scanId = coreProxyPrefix baseurl /: "
 
 -- /projects/{projectID}/scans/{scanID}/build-graphs/{buildGraphID}/rules
 uploadBuildGraphChunkEndpoint :: Url 'Https -> Text -> Text -> Text ->  Url 'Https
-uploadBuildGraphChunkEndpoint baseurl projectId scanId buildGraphId = coreProxyPrefix baseurl /: "projects" /: projectId /: "scans" /: scanId /: "build-graphs" /: buildGraphId /: "rules"
+uploadBuildGraphChunkEndpoint baseurl projectId scanId buildGraphId = coreProxyPrefix baseurl /: "projects" /: projectId /: "scans" /: scanId /: "build-graphs" /: buildGraphId /: "rules" /: "partial"
 
 -- /projects/{projectID}/scans/{scanID}/build-graphs/{buildGraphID}/rules/complete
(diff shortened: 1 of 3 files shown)


src/App/VPSScan/Scan/ScotlandYard.hs
```diff
@@ -82,17 +89,27 @@ createScotlandYardScan ScotlandYardOpts {..} = runHTTP $ do
   pure (responseBody resp)
 
 -- Given the results from a run of IPR, a scan ID and a URL for Scotland Yard,
--- post the IPR result to the "Upload Scan Data" endpoint on Scotland Yard
--- POST /scans/{scanID}/discovered_licenses
-uploadIPRResults :: (ToJSON a, Has (Lift IO) sig m, Has Diagnostics sig m) => Text -> a -> ScotlandYardOpts -> m ()
+-- post the IPR result in chunks of 1000 entries to the "Upload IPR Data" endpoint on Scotland Yard.
```
kitified, line 92: I'd love if this change allowed us to no longer need customers to reconfigure their ingresses for large bodies.

My vague thoughts are:
* Would it be possible to have a "smart upload window" that resizes smaller if we get a 413 error?
* Would it be possible to split based on upload body size instead of number of files?

If neither of these are simple to accomplish, let's just keep going as is and make a ticket to revisit this later- either way this is better than what we have now for sure!

kitified (approved):

src/App/VPSScan/Scan/ScotlandYard.hs
spatten: That's a really good idea. Let me play around with this, timeboxed to an hour or so, and get back to you.

spatten:
@kitified I've tested this with a large IPR result, and it chunks the result up by size. I think this is good to go.

I tested by grabbing the results of an IPR scan from a subset of an AOSP scan (specifically the systems directory) which is around 27 MB in size. I replaced `vendor/ramjet-cli-ipr` with this:

```bash
#!/usr/bin/env bash

cat /path/to/ramjet-cli-ipr-result.json
```

and ran it. It chunked it up into a bunch of requests and all went smoothly.

I did the same for the build-graph chunked uploads by testing with the ninja_deps file output from building the Seastar project, and it worked as well.

src/App/VPSScan/Scan/ScotlandYard.hs
spatten: I have something. It's really inefficient in that it calculates the length of the chunks many times, but it works. I think we can ship this now and refactor to make it a bit less gross later.

kitified: Sounds great! 👍 thanks so much.

src/App/VPSScan/Scan/ScotlandYard.hs
```diff
@@ -48,8 +51,12 @@ createScanEndpoint :: Url 'Https -> Text -> Url 'Https
 createScanEndpoint baseurl projectId = coreProxyPrefix baseurl /: "projects" /: projectId /: "scans"
 
 -- /projects/{projectID}/scans/{scanID}/discovered_licenses
```
kitified, line 53: [nit]
```suggestion
-- /projects/{projectID}/scans/{scanID}/discovered_licenses/partial
```

src/App/VPSScan/Scan/ScotlandYard.hs
kitified, line 92:
```suggestion
-- post the IPR result in ~1MB chunks to the "Upload IPR Data" endpoint on Scotland Yard.
```

kitified (approved):

PR #100, opened 2020-07-13
Bump dependency versions

cnr:
This removes our dependency on the prerelease `fused-effects` / my fork of `fused-effects-exceptions`. There are a few propagating changes, but most suspiciously, `WalkSkipSome` is now causing compile errors where it wasn't before, but should have.

### DIFF
--- a/src/Discovery/Walk.hs
+++ b/src/Discovery/Walk.hs
@@ -1,22 +1,29 @@
 module Discovery.Walk
   ( -- * Walking the filetree
-    walk
-  , WalkStep(..)
+    walk,
+    WalkStep (..),
+    dirName,
+    fileName,
+  )
+where
 
-  , dirName
-  , fileName
-  ) where
-
-import Prologue
-
+import Control.Monad.IO.Class (MonadIO)
+import Data.Text (Text)
+import qualified Data.Text as T
+import Path
+import Prelude
 import Path.IO
 
 data WalkStep
-  = WalkContinue -- ^ Continue walking subdirectories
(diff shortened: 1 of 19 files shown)


src/Discovery/Walk.hs
```diff
@@ -1,46 +1,53 @@
...
 walk f = walkDir $ \dir subdirs files -> do
   -- normally, subdirs and files are _relative to dir_. We want them to be
   -- relative to the filetree walk
   step <- f dir subdirs files
   case step of
-    WalkContinue -> pure (WalkExclude [])
-    WalkSkipSome dirs -> pure (WalkExclude dirs)
-    WalkSkipAll -> pure (WalkExclude subdirs)
+    WalkContinue -> pure $ WalkExclude []
+    WalkSkipSome dirs ->
```
zlav, line 45: Does this solve the `WalkSkipSome` compilation errors you mentioned?

src/Effect/ReadFS.hs
```diff
@@ -130,6 +132,13 @@ readContentsJson file = do
     Left err -> fatal (FileParseError (toFilePath file) (T.pack err))
     Right a -> pure a
 
+readContentsToml :: (Has ReadFS sig m, Has Diagnostics sig m) => Toml.TomlCodec a -> Path b File -> m a
```
zlav, line 135: Nice refactor

src/Discovery/Walk.hs
cnr: It does -- I'm genuinely surprised there weren't compilation errors before

PR #101, opened 2020-07-20
Pass filter option to `ramjet-cli-ipr` and `sherlock-cli`

kitified:
* Accept filter options as a text blob, but specify in the option help text that the value provided should be a JSON encoded string array.
* Pass filter options blob to `ramjet-cli-ipr` and `sherlock-cli` as is.

Note: won't be merged until support for this has been merged into downstream CLIs.


cnr:
LGTM too 👍 

In general, we try to avoid `String` in haskell in favor of `Text`, but you obviously couldn't do that here because of existing code that uses `String`. I opened a PR that would allow using `Text` here as well: https://github.com/fossas/spectrometer/pull/103

kitified:
@spatten / @cnr updated with Connor's feedback, wanted to get another sanity check before merging if you don't mind.

PR #103, opened 2020-07-20
Use Text for Effect.Exec arguments

cnr:
We previously had to use `String` in a bunch of places, because our `Effect.Exec` operations took `String`s as arguments instead of `Text`. This migrates `Effect.Exec` and its callsites


src/App/VPSScan/Scan/RunIPR.hs
```diff
@@ -14,13 +14,14 @@ import qualified Data.HashMap.Strict as HM
 import qualified Data.Vector as V
 import Effect.Exec
 import Data.Aeson
+import Data.Text
```
spatten, line 17: We should do `import qualified Data.Text as T` here and then `T.pack` further down in the file

PR #105, opened 2020-07-28
Update to use reverse proxy to communicate with SY

kitified:
This PR updates `vpscli` to use the Core reverse proxy to communicate with SY instead of communicating directly.

_Note: this will not be merged until fossas/FOSSA#4858 is accepted._

### DIFF
--- a/src/App/VPSScan/Scan/ScotlandYard.hs
+++ b/src/App/VPSScan/Scan/ScotlandYard.hs
@@ -22,7 +22,7 @@ coreProxyPrefix :: Url 'Https -> Url 'Https
 coreProxyPrefix baseurl = baseurl /: "api" /: "proxy" /: "scotland-yard"
 
 authHeader :: Text -> Option scheme
-authHeader apiKey = header "Authorization" (encodeUtf8 ("token " <> apiKey))
+authHeader apiKey = header "Authorization" (encodeUtf8 ("Bearer " <> apiKey))
 
 -- /projects/{projectID}/scans
 createScanEndpoint :: Url 'Https -> Text -> Url 'Https
--- a/src/App/VPSScan/Scan/ScotlandYard.hs
+++ b/src/App/VPSScan/Scan/ScotlandYard.hs
@@ -9,9 +9,9 @@ where
 
 import App.VPSScan.Types
 import Control.Effect.Diagnostics
-import Control.Monad.IO.Class
+import Control.Monad.IO.Class (MonadIO)
 import Data.Aeson
-import Data.Text
+import Data.Text (Text, append, pack)
 import Network.HTTP.Req
 import Prelude
 import App.Util (parseUri)
@@ -26,7 +26,7 @@ authHeader apiKey = header "Authorization" (encodeUtf8 (append (pack "token ") a
 
 -- /projects/{projectID}/scans
 createScanEndpoint :: Url 'Https -> Text -> Url 'Https
-createScanEndpoint baseurl projectId = (coreProxyPrefix baseurl) /: "projects" /: projectId /: "scans"
+createScanEndpoint baseurl projectId = coreProxyPrefix baseurl /: "projects" /: projectId /: "scans"
 
 -- /projects/{projectID}/scans/{scanID}/discovered_licenses
 scanDataEndpoint :: Url 'Https -> Text -> Text -> Url 'Https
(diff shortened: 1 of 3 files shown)


src/App/VPSScan/Scan/ScotlandYard.hs
```diff
@@ -36,20 +44,22 @@ instance FromJSON ScanResponse where
 createScotlandYardScan :: (MonadIO m, Has Diagnostics sig m) => VPSOpts -> m ScanResponse
 createScotlandYardScan VPSOpts {..} = runHTTP $ do
   let body = object ["organizationId" .= organizationID, "revisionId" .= revisionID, "projectId" .= projectID]
-      ScotlandYardOpts {..} = vpsScotlandYard
+      CoreServerOpts {..} = coreInstance
+  let auth = authHeader fossaApiKey
```
spatten, line 48: nit: you can just continue the `let` block on line 48

src/App/VPSScan/Scan/ScotlandYard.hs
```diff
@@ -9,16 +9,24 @@ where
 
 import App.VPSScan.Types
 import Control.Effect.Diagnostics
-import Control.Monad.IO.Class (MonadIO)
+import Control.Monad.IO.Class
```
spatten, line 12: nit: Unless Connor disagrees, I think we should leave these lines as is. I think the goal is to constrict imports as much as possible so that we document which parts of the import we're actually using.

src/App/VPSScan/Scan/ScotlandYard.hs
```diff
@@ -9,16 +9,24 @@ where
...
+coreProxyPrefix :: Url 'Https -> Url 'Https
+coreProxyPrefix baseurl = baseurl /: "api" /: "proxy" /: "scotland-yard"
+
+authHeader :: Text -> Option scheme
+authHeader apiKey = header "Authorization" (encodeUtf8 (append (pack "token ") apiKey))
 
 -- /projects/{projectID}/scans
 createScanEndpoint :: Url 'Https -> Text -> Url 'Https
-createScanEndpoint baseurl projectId = baseurl /: "projects" /: projectId /: "scans"
+createScanEndpoint baseurl projectId = (coreProxyPrefix baseurl) /: "projects" /: projectId /: "scans"
```
spatten, line 29: one more minor nit: `hlint` says that the parens on this line are redundant :)

src/App/VPSScan/Scan/ScotlandYard.hs
kitified: Fair enough. I removed them because whatever linter is set up was claiming they were redundant.

src/App/VPSScan/Scan/ScotlandYard.hs
```diff
@@ -11,14 +11,22 @@ import App.VPSScan.Types
...
 import Prelude
 import App.Util (parseUri)
+import Data.Text.Encoding
+
+-- Prefix for Core's reverse proxy to SY
+coreProxyPrefix :: Url 'Https -> Url 'Https
+coreProxyPrefix baseurl = baseurl /: "api" /: "proxy" /: "scotland-yard"
+
+authHeader :: Text -> Option scheme
+authHeader apiKey = header "Authorization" (encodeUtf8 (append (pack "token ") apiKey))
```
cnr, line 25: [nit]
- `append` on text is also called `<>`
- `pack "token "` is the same as `"token "` with the `OverloadedStrings` extension enabled
so:
```suggestion
authHeader apiKey = header "Authorization" (encodeUtf8 ("token " <> apiKey))
```

src/App/VPSScan/Types.hs
```diff
@@ -17,8 +17,8 @@ import Text.URI (URI)
 import Network.HTTP.Req
 import Data.Text.Prettyprint.Doc (viaShow)
 
-data ScotlandYardOpts = ScotlandYardOpts
-  {scotlandYardUrl :: URI}
+data CoreServerOpts = CoreServerOpts
+  {fossaUrl :: URI, fossaApiKey :: Text}
```
cnr, line 21: [nit] we should make the naming here consistent, either:
1. renaming `CoreServerOpts` to `FossaServerOpts`
or
2. `fossaUrl` -> `coreUrl`; `fossaApiKey` -> `coreApiKey`

src/App/VPSScan/Scan/ScotlandYard.hs
```diff
@@ -15,10 +15,18 @@ import Data.Text (Text)
...
 import Prelude
 import App.Util (parseUri)
+import Data.Text.Encoding
+
+-- Prefix for Core's reverse proxy to SY
+coreProxyPrefix :: Url 'Https -> Url 'Https
+coreProxyPrefix baseurl = baseurl /: "api" /: "proxy" /: "scotland-yard"
+
+authHeader :: Text -> Option scheme
+authHeader apiKey = header "Authorization" (encodeUtf8 ("token " <> apiKey))
```
dchenk, line 25:
```suggestion
authHeader apiKey = header "Authorization" (encodeUtf8 ("Bearer " <> apiKey))
```
As far as I know, we made up the "token XXXXX" pattern for the Authorization header, but some tools like Swagger won't allow it, so to get [swagger.io](https://swagger.io/) to work for the Core API we needed to add support for the Bearer scheme: https://github.com/fossas/FOSSA/pull/4450

kitified: Oh, nice. Thanks.

PR #106, opened 2020-07-29
Catch IOExceptions in the Effect.Exec handler

cnr:
Previously, errors like "command doesn't exist" would be an uncaught exception.

### DIFF
--- a/src/Effect/Exec.hs
+++ b/src/Effect/Exec.hs
@@ -17,6 +17,7 @@ where
 import Control.Applicative (Alternative)
 import Control.Algebra as X
 import Control.Effect.Diagnostics
+import Control.Exception (IOException, try)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import qualified Data.ByteString.Lazy as BL
 import Data.Aeson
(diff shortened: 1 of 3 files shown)


src/Effect/Exec.hs
```diff
@@ -129,17 +132,29 @@ instance (Algebra sig m, MonadIO m) => Algebra (Exec :+: sig) (ExecIOC m) where
...
+                  (_, Never) -> Left $ mkFailure exitcode stdout stderr
+                  (_, NonEmptyStdout) ->
+                    if BL.null stdout
+                      then Left $ mkFailure exitcode stdout stderr
+                      else Right stdout
+                  (_, Always) -> Right stdout
+
+
+      let result :: Either CmdFailure Stdout
+          result = first ioExceptionToCmdFailure processResult >>= mangleResult
```
cnr, line 158: `first` comes from `Bifunctor`, which is like `Functor` but allows mapping over two positional arguments:
```haskell
class Bifunctor (p :: Type -> Type -> Type) where
  bimap :: (a -> b) -> (c -> d) -> p a c -> p b d

  first :: (a -> b) -> p a c -> p b c
  first f = bimap f id

  second :: (b -> c) -> p a b -> p a c
  second g = bimap id g
```

in this case, `first` is monomorphized to:
```haskell
first ::
  (IOException -> CmdFailure) ->
  Either IOException Stdout ->
  Either CmdFailure Stdout
```

scruffystuffs:
I think this new behavior could be unit-tested, seems pretty simple.  Would it be simple to do so?  If not, let's discuss why (so I can understand how this could be tested in the future) and then I'll approve this PR w/o tests.  LGTM otherwise, but it may be worth testing some of these effects, since we can't type-assert against uncaught exceptions like this.

PR #108, opened 2020-07-30
Ninja scans in main vpscli scan

spatten:
We need to run dependency graph scans as part of a normal `vpscli` scan, so get rid of the `vpscli ninja-graph` command and add some params to `vpscli scan`

I think this will create some merge conflicts with #105


spatten:
I'm closing this as the requirement is wrong -- you can run multiple ninja scans per vpscli scan, so we should leave it as is.

PR #109, opened 2020-07-30
Only run req.txt strategy when we find files

cnr:


src/Strategy/Python/ReqTxt.hs
```diff
@@ -23,7 +23,8 @@ discover = walk $ \dir _ files -> do
   let txtFiles = filter (\f -> "req" `isInfixOf` fileName f
                             && ".txt" `isSuffixOf` fileName f) files
 
-  runSimpleStrategy "python-requirements" PythonGroup $ analyze dir txtFiles
+  unless (null txtFiles) $
+    runSimpleStrategy "python-requirements" PythonGroup $ analyze dir txtFiles
```
scruffystuffs, line 27: What is the cost of running when there are no files?  Where is it actually triggering anything?

cnr: This is related to the spurious errors we saw during the call -- this strategy runs (and fails) for each directory in the project, regardless of whether we find `*req*.txt` files

scruffystuffs: Ah, if there's no successes, line 36 is screaming at us.

PR #112, opened 2020-08-04
Initiate and complete scans from CI

kitified:
This PR adds the ability for `vpscli` to create scans from CI and mark them complete in the FOSSA server.
I also attempted to refactor some of the code to better suit the control flow of the program and other conventions within the codebase.


src/App/VPSScan/Scan/Core.hs
```diff
@@ -0,0 +1,95 @@
...
+  deriving (Eq, Ord, Show)
+
+instance FromJSON SherlockInfo where
+  parseJSON = withObject "" $ \obj -> do
+    auth <- obj .: "auth"
+    sherlockUrl <- obj .: "url"
+    sherlockOrgId <- obj .: "orgId"
+    sherlockClientId <- auth .: "clientId"
+    sherlockClientToken <- auth .: "clientToken"
+    return (SherlockInfo { sherlockUrl = sherlockUrl, sherlockClientId = sherlockClientId, sherlockClientToken = sherlockClientToken, sherlockOrgId = sherlockOrgId })
```
cnr, line 39: This is good! And we can make it a little better:

As we know, `Functor` allows lifting a function `a -> b` into `f a -> f b`:
```hs
class Functor f where
  fmap :: (a -> b) -> (f a -> f b)
```

which we use elsewhere to "lift" newtype wrappers (like `FilterExpressions`) when we parse command line options.

there's an abstraction, `Applicative` that generalizes this to lifting `N`-ary functions:
```hs
-- this is the same as fmap
liftA1 :: Applicative f => (a -> b) -> (f a -> f b)
liftA1 f a = f <$> a

liftA2 :: Applicative f => (a -> b -> c) -> (f a -> f b -> f c)
liftA2 f a b = f <$> a <*> b

liftA3 :: Applicative f => (a -> b -> c -> d) -> (f a -> f b -> f c -> f d)
liftA3 f a b c = f <$> a <*> b <*> c

liftA4 :: Applicative f => (a -> b -> c -> d -> e) -> (f a -> f b -> f c -> f d -> f e)
liftA4 f a b c d = f <$> a <*> b <*> c <*> d
```

 (the right-side parentheses are redundant, but that's not really important)

we notice a pattern here: for each new argument, we chain another `<*> val`

---

how does this apply here? well, it turns out that the `SherlockInfo` data constructor is a function:
```hs
SherlockInfo :: Text -> Text -> Text -> Text -> SherlockInfo
```

and we have a bunch of aeson `Parser`s for the fields:
```hs
obj .: "url" :: Parser Text
obj .: "orgId" :: Parser Text
...
```

and we're building a `Parser SherlockInfo`. squinting at `liftA4` again:
```hs
liftA4 :: (Text -> Text -> Text -> Text -> SherlockInfo)
       -> Parser Text
       -> Parser Text
       -> Parser Text
       -> Parser Text
       -> Parser SherlockInfo
```

or applying it directly:
```hs
liftA4 SherlockInfo (obj .: "url") (obj .: "orgId") (auth .: "clientId") (auth .: "clientToken")
```

and inlining the definition of `liftA4`:
```hs
SherlockInfo <$> obj .: "url"
             <*> obj .: "orgId"
             <*> auth .: "clientId"
             <*> auth .: "clientToken"
```

---

bringing it all together:
```hs
instance FromJSON SherlockInfo where
  parseJSON = withObject "SherlockInfo" $ \obj -> do
    auth <- obj .: "auth"

    SherlockInfo
      <$> obj .: "url"
      <*> obj .: "orgId"
      <*> auth .: "clientId"
      <*> auth .: "clientToken"
```

---

it may not be immediately clear why this refactor matters. the primary motivator is that using `Applicative` gives us _less_ power, so there's substantially less mental overhead in determining what this does, whether it's correct, and making modifications to it later. we can discuss this later if you'd like, with examples

src/App/VPSScan/Scan/Core.hs
```diff
@@ -0,0 +1,95 @@
...
+coreAuthHeader apiKey = header "Authorization" (encodeUtf8 ("Bearer " <> apiKey))
+
+
+buildRevision :: (MonadIO m) => Text -> m Text
+buildRevision "" = do
+  posixTime <- liftIO getPOSIXTime
+  pure (pack $ show $ (floor $ toRational posixTime :: Int))
+buildRevision userProvidedRevision = pure (userProvidedRevision)
+
+createLocator :: Text -> Text -> Text
```
cnr, line 51: I'm curious if our `Locator` type in `Srclib.Types` could be reused here

src/App/VPSScan/Scan/ScotlandYard.hs
```diff
@@ -41,24 +45,28 @@ instance FromJSON ScanResponse where
...
 
-createScotlandYardScan :: (MonadIO m, Has Diagnostics sig m) => VPSOpts -> m ScanResponse
-createScotlandYardScan VPSOpts {..} = runHTTP $ do
-  let body = object ["organizationId" .= organizationID, "revisionId" .= revisionID, "projectId" .= projectID]; FossaOpts {..} = fossaInstance
-  let auth = authHeader fossaApiKey
+createScotlandYardScan :: (MonadIO m, Has Diagnostics sig m) => ScotlandYardOpts -> m ScanResponse
+createScotlandYardScan ScotlandYardOpts {..} = runHTTP $ do
+  let VPSOpts{..} = syVpsOpts
+  let FossaOpts{..} = fossa
+  let parsedOrgId = read $ unpack organizationId :: Int
```
cnr, line 52: can we decode this to an `Int` upstream somewhere? `read` is dangerous, and passing around an invalid organization ID is awkward. we should parse it as soon as possible

src/App/VPSScan/Types.hs
```diff
@@ -10,38 +10,33 @@ module App.VPSScan.Types
...
 import Control.Carrier.Diagnostics
 import Prologue
 import Text.URI (URI)
 import Network.HTTP.Req
 import Data.Text.Prettyprint.Doc (viaShow)
 
-data FossaOpts = FossaOpts
-  {fossaUrl :: URI, fossaApiKey :: Text}
-  deriving (Generic)
+newtype FilterExpressions = FilterExpressions String
```
cnr, line 19: Let's use `Text`

src/App/VPSScan/Types.hs
```diff
@@ -10,38 +10,33 @@ module App.VPSScan.Types
...
+
+instance Show FilterExpressions where
+  show (FilterExpressions x) = x :: String
```
cnr, line 22: rather than using `Show`, the usual pattern is to add a record field:
```hs
newtype FilterExpressions = FilterExpressions { unFilterExpressions :: Text }
```

you can use `unFilterExpressions` to extract the underlying `Text` (and kill this `Show` instance, though `deriving Show` could still be useful)

src/App/VPSScan/Scan/Core.hs
kitified: It doesn't include explicit support for `orgId`, we'd have to manually build it in to the `locatorProject` string.
But I do agree we shouldn't use `Text`- `newtype`'d it.

src/App/VPSScan/Main.hs
```diff
@@ -22,68 +21,38 @@ commands :: Parser (IO ())
...
-vpsOpts = VPSOpts <$> runSherlockOpts <*> optional runIPROpts <*> fossaOpts <*> organizationIDOpt <*> projectIDOpt <*> revisionIDOpt <*> filterOpt
-            where
-              organizationIDOpt = option auto (long "organization" <> metavar "orgID" <> help "Organization ID")
-              projectIDOpt = strOption (long "project" <> metavar "String" <> help "Project ID")
-              revisionIDOpt = strOption (long "revision" <> metavar "String" <> help "Revision ID")
+-- vpsOpts = VPSOpts <$> fossaOpts <*> filterOpt <*> projectNameOpt <*> skipIprScanOpt
+vpsOpts = VPSOpts <$> fossaOpts <*> projectNameOpt <*> projectRevision <*> skipIprScanOpt <*> filterOpt
+  where
+    projectNameOpt = strOption (long "project-name" <> metavar "STRING" <> help "The name of the project to create in FOSSA")
+    projectRevision = strOption (long "project-revision" <> metavar "STRING" <> help "The revision of the project to create in FOSSA. If not specified, uses the current Unix timestamp." <> value "")
```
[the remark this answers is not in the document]
kitified: I don't see a way to do this and still allow the flag to be optional, although I'm probably misreading.

src/App/VPSScan/Main.hs
```diff
@@ -22,68 +21,38 @@ commands :: Parser (IO ())
...
+    urlOpt = uriOption (long "fossa-url" <> metavar "STRING" <> help "URL for FOSSA service")
+    apiKeyOpt =  strOption (long "fossa-api-key" <> metavar "STRING" <> help "API key for FOSSA service")
 
 basedirOpt :: Parser FilePath
 basedirOpt = strOption (long "basedir" <> short 'd' <> metavar "DIR" <> help "Base directory for scanning" <> value ".")
 
-filterOpt :: Parser RunIPR.FilterExpressions
-filterOpt = RunIPR.FilterExpressions <$> strOption (long "ignore-file-regex" <> short 'i' <> metavar "REGEXPS" <> help "JSON encoded array of regular expressions used to filter scanned paths" <> value "[]")
+filterOpt :: Parser FilterExpressions
+filterOpt = FilterExpressions <$> strOption (long "ignore-file-regex" <> short 'i' <> metavar "REGEXPS" <> help "JSON encoded array of regular expressions used to filter scanned paths" <> value "[]")
```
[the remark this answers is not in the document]
kitified: I'd be OK to implement that. I'd rather save that for a separate PR though since that's not directly related to this PR.

src/App/VPSScan/Main.hs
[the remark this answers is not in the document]
cnr: We use `optional :: Alternative f => f a -> f (Maybe a)` elsewhere

(monomorphized: `optional :: Parser a -> Parser (Maybe a)`)

kitified:
@cnr I still have some iteration to do on the build action, but wanted to go ahead and put the code up for re-review :)

spectrometer.cabal
```diff
@@ -58,6 +59,7 @@ common deps
     , conduit-extra                ^>=1.3.5
     , containers                   ^>=0.6.0
     , cpio-conduit                 ^>=0.7.0
+    , directory                    ^>=1.3
```
cnr, line 62: [nit] let's kill this `directory` dependency

src/App/VPSScan/EmbeddedBinary.hs
```diff
@@ -0,0 +1,73 @@
...
+          go
+  where
+    go tmpDir = do
+      let binaryPath = fromAbsDir tmpDir </> "sherlock-cli"
+      liftIO (writeFile binaryPath embeddedBinarySherlockCli)
+      liftIO (makeExecutable binaryPath)
+      act binaryPath
+
+data IPRBinaryPaths = IPRBinaryPaths
+  { ramjetBinaryPath :: FilePath
```
cnr, line 31: [nit] [not blocking] I'd prefer these to be typed paths -- `Path Abs File`, probably. this helps reinforce that it's safe to `exec` this script from _any_ directory, because it's not a relative path to a script

I know I don't do it in the maven plugin module, but I don't know what past-me was thinking

src/App/VPSScan/Scan/ScotlandYard.hs
```diff
@@ -1,29 +1,33 @@
...
+  , ScotlandYardOpts (..)
   )
 where
 
 import App.VPSScan.Types
+import App.VPSScan.Scan.Core
 import Control.Effect.Diagnostics
 import Control.Monad.IO.Class (MonadIO)
 import Data.Aeson
-import Data.Text (Text)
```
cnr, line 14: [nit] let's keep this import as-is -- if you need functions from `Data.Text`, e.g., `pack` also add:
```hs
import qualified Data.Text as T
```
and use `T.pack`

there are a bunch of overlapping function names in `Data.Text` -- so it's meant to be imported qualified

kitified:
These builds appear to be failing due to whatever is going on at Github right now (network operations seem to be failing across at least two of our repos). I'd like to go ahead and merge despite failing tests. 

When 983a768 succeeded I downloaded it and used it to test against FOSSA Core, so I'm confident the changes are good to go. @cnr agree, or prefer I wait for green builds?

cnr:
Feel free to merge 👍

PR #113, opened 2020-08-05
SPEC file parsing with focus on requires dependencies

zlav:
After customer feedback we have determined that focusing on the `requires` lines in the rpm spec file is more important than the `buildrequires` lines. This will be mitigated in the future by introducing package tagging, but at the moment the best solution is only uploading `requires` deps.

### DIFF
--- a/test/RPM/SpecFileSpec.hs
+++ b/test/RPM/SpecFileSpec.hs
@@ -80,11 +80,11 @@ spec = do
       getTypeFromLine "Requires: xz = 1" `Test.shouldBe` Just (RuntimeRequires $ mkVerDep "xz" $ CEq "1")
   --
   Test.describe "rpm dependency grapher" $
-    Test.it "should produce flat BuildRequires" $ do
+    Test.it "should produce flat RuntimeRequires" $ do
       let deps = Dependencies [boostDep, xzDep] [tacpDep, jsoncppDep]
           gr = buildGraph deps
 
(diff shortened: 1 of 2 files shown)


test/RPM/SpecFileSpec.hs
```diff
@@ -80,11 +80,11 @@ spec = do
...
   Test.describe "rpm dependency grapher" $
-    Test.it "should produce flat BuildRequires" $ do
+    Test.it "should produce flat RuntimeRequires" $ do
       let deps = Dependencies [boostDep, xzDep] [tacpDep, jsoncppDep]
           gr = buildGraph deps
 
-      expectDeps (map toDependency [boostDep, xzDep]) gr
-      expectDirect (map toDependency [boostDep, xzDep]) gr
+      expectDeps [toDependency tacpDep] gr
+      expectDirect [toDependency tacp] gr
```
scruffystuffs, line 88: This should include both of the runtime deps.

on [line 84](https://github.com/fossas/spectrometer/pull/113/files#diff-8f235742309df0fca599b6da5b06d997R84), we have the creation of the base structure as `Dependencies buildRequiresList runtimeRequiresList`.

This should produce a flat graph (planar?) equal to the list passed in the constructor.

I would consider reverting this part of the change and just changing the variables to the other list of deps:

```suggestion
      expectDeps (map toDependency [tacpDep, jsoncppDep]) gr
      expectDirect (map toDependency [tacpDep, jsoncppDep]) gr
```

PR #116, opened 2020-08-19
ghc 8.10

cnr:
This PR includes changes required to build with ghc 8.10

unrelated: this adds fallback keys to the `cache` action, which improves how often we get a cache hit

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,3 +1,8 @@
 packages: .
 tests: True
 optimization: False
+
+source-repository-package
+  type: git
+  location: https://github.com/fossas/yarn-lock.git
+  tag: master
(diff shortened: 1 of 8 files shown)


cabal.project
```diff
@@ -1,3 +1,8 @@
 packages: .
 tests: True
 optimization: False
+
+source-repository-package
+  type: git
+  location: https://github.com/fossas/yarn-lock.git
```
cnr, line 7: had to fork yarn-lock to widen its dependency version bounds

src/App/VPSScan/NinjaGraph.hs
```diff
@@ -74,9 +74,8 @@ getNinjaDeps baseDir opts@NinjaGraphOpts{..} =
     Nothing -> BL.toStrict <$> generateNinjaDeps baseDir opts
     Just ninjaPath -> readNinjaDepsFile ninjaPath
 
-scanNinjaDeps :: (Has Diagnostics sig m) => NinjaGraphOpts -> ByteString -> m [DepsTarget]
-scanNinjaDeps NinjaGraphOpts{..} ninjaDepsContents =
```
cnr, line 78: turns out the `NinjaGraphOpts` argument here was unused -- and ghc 8.10 picked this up 🎉

spatten: Nice!

PR #119, opened 2020-08-24
Integration testing: prerequisites

cnr:
For integration testing, we want to use a different interpreter of the `Exec` effect. As currently written, `discover` functions use `runStrategy` to internally interpret the `Exec` effect, which prevents us from doing so.

Instead, these changes force the `Has Exec`/`Has ReadFS` constraints upward, through the `discover`, so we can interpret `Exec` with a different interpreter in our tests.

---

This also fixes a few issues discovered while building integration tests:
- `rebar3tree` would treat dependencies as top-level projects
- the `setup.py` parser would horribly break if you even looked at it the wrong way (didn't allow double-quotes, was whitespace-sensitive, didn't allow trailing commas in the dependency list)

### DIFF
--- a/src/Effect/Exec.hs
+++ b/src/Effect/Exec.hs
@@ -4,6 +4,7 @@ module Effect.Exec
     exec,
     execThrow,
     Command (..),
+    CmdFailure (..),
     AllowErr (..),
     execParser,
     execJson,
@@ -19,8 +20,10 @@ import Control.Algebra as X
 import Control.Effect.Diagnostics
 import Control.Exception (IOException, try)
 import Control.Monad.IO.Class (MonadIO, liftIO)
-import qualified Data.ByteString.Lazy as BL
 import Data.Aeson
+import Data.Bifunctor (first)
+import qualified Data.ByteString.Lazy as BL
+import Data.String (fromString)
 import Data.Text (Text)
 import qualified Data.Text as T
 import qualified Data.Text.Lazy as TL
@@ -34,8 +37,6 @@ import System.Process.Typed
 import Text.Megaparsec (Parsec, runParser)
 import Text.Megaparsec.Error (errorBundlePretty)
 import Prelude
-import Data.Bifunctor (first)
-import Data.String (fromString)
 
 data Command = Command
   { -- | Command name to use. E.g., "pip", "pip3", "./gradlew".
@@ -145,14 +146,13 @@ instance (Algebra sig m, MonadIO m) => Algebra (Exec :+: sig) (ExecIOC m) where
       let mangleResult :: (ExitCode, Stdout, Stderr) -> Either CmdFailure Stdout
           mangleResult (exitcode, stdout, stderr) =
             case (exitcode, cmdAllowErr cmd) of
-                  (ExitSuccess, _) -> Right stdout
-                  (_, Never) -> Left $ mkFailure exitcode stdout stderr
-                  (_, NonEmptyStdout) ->
-                    if BL.null stdout
-                      then Left $ mkFailure exitcode stdout stderr
-                      else Right stdout
-                  (_, Always) -> Right stdout
-
+              (ExitSuccess, _) -> Right stdout
+              (_, Never) -> Left $ mkFailure exitcode stdout stderr
+              (_, NonEmptyStdout) ->
+                if BL.null stdout
+                  then Left $ mkFailure exitcode stdout stderr
+                  else Right stdout
+              (_, Always) -> Right stdout
 
       let result :: Either CmdFailure Stdout
           result = first ioExceptionToCmdFailure processResult >>= mangleResult
--- a/src/Strategy/Erlang/Rebar3Tree.hs
+++ b/src/Strategy/Erlang/Rebar3Tree.hs
@@ -25,10 +25,10 @@ import Types
 discover :: HasDiscover sig m => Path Abs Dir -> m ()
 discover = walk $ \dir _ files -> do
   case find (\f -> (fileName f) == "rebar.config") files of
-    Nothing -> pure ()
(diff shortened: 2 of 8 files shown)


src/Effect/Exec.hs
```diff
@@ -145,14 +146,13 @@ instance (Algebra sig m, MonadIO m) => Algebra (Exec :+: sig) (ExecIOC m) where
...
             case (exitcode, cmdAllowErr cmd) of
-                  (ExitSuccess, _) -> Right stdout
-                  (_, Never) -> Left $ mkFailure exitcode stdout stderr
-                  (_, NonEmptyStdout) ->
-                    if BL.null stdout
-                      then Left $ mkFailure exitcode stdout stderr
-                      else Right stdout
-                  (_, Always) -> Right stdout
-
+              (ExitSuccess, _) -> Right stdout
```
cnr, line 149: a lot of unrelated formatting changes in this commit; continuing to do formatting with `ormolu` piecemeal as I edit files

src/Strategy/Erlang/Rebar3Tree.hs
```diff
@@ -25,10 +25,10 @@ import Types
...
 discover = walk $ \dir _ files -> do
   case find (\f -> (fileName f) == "rebar.config") files of
-    Nothing -> pure ()
-    Just _  -> runSimpleStrategy "erlang-rebar3tree" ErlangGroup $ analyze dir
-
-  pure WalkContinue
+    Nothing -> pure WalkContinue
+    Just _  -> do
+      runSimpleStrategy "erlang-rebar3tree" ErlangGroup $ analyze dir
+      pure WalkSkipAll
```
cnr, line 31: This is the fix for the rebar3tree issue

src/Strategy/Python/SetupPy.hs
```diff
@@ -45,8 +46,11 @@ type Parser = Parsec Void Text
 installRequiresParser :: Parser [Req]
 installRequiresParser = prefix *> entries <* end
   where
-  prefix  = skipManyTill anySingle (string "install_requires=[")
-  entries = between quote quote requirementParser `sepBy` char ','
-  end     = char ']'
+  prefix  = skipManyTill anySingle (symbol "install_requires") *> symbol "=" *> symbol "["
```
cnr, line 49: These are the `setup.py` parser fixes (I didn't know about `symbol`/lexemes when I first wrote this)

src/Types.hs
```diff
@@ -70,15 +68,15 @@ runStrategy name strategyGroup act = forkTask $ do
         let (bodies, ()) = resultValue result
          in traverse_ (output . toProjectClosure strategyGroup name) bodies -- TODO: warnings
 
-type TaskC m = ExecIOC (ReadFSIOC (DiagnosticsC m))
```
cnr, line 73: Rather than using `ExecIOC` and `ReadFSIOC` as part of running the task, we send the constraints upward (see the below additions to `HasDiscover`)

src/Strategy/Python/SetupPy.hs
```diff
@@ -45,8 +46,11 @@ type Parser = Parsec Void Text
...
+  entries = (between quote quote requirementParser) `sepEndBy` symbol ","
```
scruffystuffs, line 50: How would this `between quote quote ...` handle mismatched quoting styles like `install_requires = ["requests']`?
Open question: Should it be permissive and accept mismatches, or should it just search for matching quotes?

cnr: You're right; fixed: https://github.com/fossas/spectrometer/pull/119/commits/f778e9fd4125cc50ea936ff622e368bc0648015f

src/Strategy/Erlang/Rebar3Tree.hs
zlav: Why does this fix the issue? Shouldn't we continue walking to find other `rebar.config` files?

src/Strategy/Python/SetupPy.hs
zlav: Do you have an example of what this fixes?

src/Strategy/Erlang/Rebar3Tree.hs
cnr: Specifically, the issue is that dependencies are downloaded into a subdirectory, and those dependencies _also_ contain `rebar.config`s, e.g., (`.rebar/deps` isn't accurate here; I forget what the directories are actually called)
```
myproject/rebar.config
myproject/.rebar/deps/packageA/rebar.config
myproject/.rebar/deps/packageB/rebar.config
```

..so we end up treating those as direct dependencies.

Worse though, anytime `rebar3 tree` is called on one of those dependencies, it then downloads _another level_ of dependencies under it, which also contain `rebar.config`s:

```
myproject/rebar.config
myproject/.rebar/deps/packageA/rebar.config
myproject/.rebar/deps/packageA/.rebar/deps/packageC/rebar.config
myproject/.rebar/deps/packageB/rebar.config
myproject/.rebar/deps/packageB/.rebar/deps/packageD/rebar.config
```

and this expands an additional level each time we run analysis

---

I don't know if the correct option is to ignore `rebar.config` in subdirectories, or just to blacklist the `.rebar/deps` (or whatever it's actually called)

src/Strategy/Python/SetupPy.hs
cnr: - whitespace between things, e.g., `install_requires  =     \n[ 'foo']`
- trailing commas `install_requires=['foo',]`
- double-quotes used instead of single-quotes `install_requires=["foo"]`

src/Strategy/Erlang/Rebar3Tree.hs
zlav: Ah, this makes sense. I like your change for now and we can revert or modify in the future. Thanks for the very detailed response.

PR #120, opened 2020-08-24
Create internal text util w/ tests

scruffystuffs:
_scheduling notes: Working on Cabal analyzer, thought I would need this utility.   After some refactoring, didn't need it anymore, so I broke it out into a separate PR.  This is a simple, small, tech debt PR._

This PR adds the `splitOnceOn` and `splitOnceOnEnd` utilities, which are both of the type `Text -> Text -> (Text, Text)`


scruffystuffs:
That what they're referred to as in the `Data.Text` source code.  I also couldn't think of anything better.

PR #121, opened 2020-08-25
Composer analyzer support

zlav:
This PR creates support for the Composer package manager used most commonly by PHP developers.

Composer support in this analyzer relies on parsing the `composer.lock` file. This analyzer is capable of constructing an accurate dependency graph with pinned versions and differentiating between production and development dependencies.

Future work might include support for parsing `composer` command line output or the `composer.json` manifest file to accurately determine which dependencies are direct dependencies. This is unnecessary for the initial support and can be added once more work has. been done to make merging strategies feasible.

### DIFF
--- a/src/Strategy/Composer.hs
+++ b/src/Strategy/Composer.hs
@@ -1,17 +1,14 @@
-{-# language TemplateHaskell #-}
+{-# LANGUAGE TemplateHaskell #-}
 
 module Strategy.Composer
-  ( discover
-  , analyze
-  , buildGraph
-
-  , ComposerLock(..)
-  , CompDep(..)
-  , Source(..)
+  ( discover,
+    analyze,
+    buildGraph,
+    ComposerLock (..),
+    CompDep (..),
+    Source (..),
   )
-  where
-
-import Prologue
+where
 
 import Control.Effect.Diagnostics
 import qualified Data.Map.Strict as M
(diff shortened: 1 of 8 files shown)


scruffystuffs:
I strongly suggest rolling back the merge to master commit and then rebasing this on master.

scruffystuffs:
Is it possible to simplify the `testdata/composer.lock` file down to its minimal parts?  I'm seeing a lot of info that's probably irrelevant for the tests.

src/Strategy/Composer.hs
```diff
@@ -0,0 +1,133 @@
...
+  where
+    dependencies =
+      ProjectDependencies
+        { dependenciesGraph = buildGraph lock,
+          dependenciesOptimal = Optimal,
+          dependenciesComplete = Complete
+        }
+
+newtype CompPkg = CompPkg {pkgName :: Text}
+  deriving (Eq, Ord, Show, Generic)
```
scruffystuffs, line 92: No need for this `Generic`, we've phased this out across the codebase in #124.
This will also require a `{-# LANGUAGE GeneralizedNewtypeDeriving #-}` declaration at the top of the module.

```suggestion
  deriving (Eq, Ord, Show)
```

src/Strategy/Composer.hs
```diff
@@ -0,0 +1,133 @@
+{-# LANGUAGE TemplateHaskell #-}
```
scruffystuffs, line 1: I don't think this is necessary, we can probably remove this.

test/Composer/ComposerLockSpec.hs
```diff
@@ -0,0 +1,75 @@
+{-# language TemplateHaskell #-}
+
+module Composer.ComposerLockSpec
+  ( spec
+  ) where
+
+import Prologue
```
scruffystuffs, line 7: `Prologue` no longer exists, and this module isn't using `TemplateHaskell`, so we can remove both of these.
```suggestion
module Composer.ComposerLockSpec
  ( spec
  ) where
```

src/Strategy/Composer.hs
```diff
@@ -0,0 +1,133 @@
...
+          dependenciesComplete = Complete
+        }
+
+newtype CompPkg = CompPkg {pkgName :: Text}
+  deriving (Eq, Ord, Show, Generic)
+
+type CompGrapher = LabeledGrapher CompPkg CompLabel
+
+data CompLabel
+  = CompVersion Text
```
scruffystuffs, line 97: Is this the version of composer that created the lock file, or the dependency version?

src/Strategy/Composer.hs
```diff
@@ -0,0 +1,133 @@
...
+    Nothing -> pure ()
+    Just file -> runSimpleStrategy "php-composerlock" PHPGroup $ analyze file
+
+  pure WalkContinue
+
+data ComposerLock = ComposerLock
+  { packages :: [CompDep],
+    packagesDev :: [CompDep]
+  }
+  deriving (Eq, Ord, Show, Generic)
```
scruffystuffs, line 35: We can drop this derived `Generic`.

src/Strategy/Composer.hs
```diff
@@ -0,0 +1,133 @@
...
+
+data CompDep = CompDep
+  { name :: Text,
+    version :: Text,
+    source :: Source,
+    -- | name to version spec
+    require :: Maybe (Map Text Text),
+    requireDev :: Maybe (Map Text Text)
+  }
+  deriving (Eq, Ord, Show, Generic)
```
scruffystuffs, line 45: We can drop this derived `Generic`.

src/Strategy/Composer.hs
```diff
@@ -0,0 +1,133 @@
...
+
+data Source = Source
+  { sourceType :: Text,
+    url :: Text,
+    reference :: Text
+  }
+  deriving (Eq, Ord, Show, Generic)
```
scruffystuffs, line 52: We can drop this derived `Generic`.

src/Strategy/Composer.hs
```diff
@@ -0,0 +1,133 @@
...
+where
+
+import Control.Effect.Diagnostics
+import qualified Data.Map.Strict as M
+import DepTypes
+import Discovery.Walk
+import Effect.Grapher
+import Effect.ReadFS
+import Graphing (Graphing)
+import Prologue
```
scruffystuffs, line 20: We done killed the prologue in #124.  You'll definitely need to add a few module to cover the missing ones.  Here's a few you definitely need:

- `Data.Aeson.Types`
- `Data.Foldable (find)`
- `Data.Text (Text)`

src/Strategy/Composer.hs
```diff
@@ -0,0 +1,133 @@
...
+
+newtype CompPkg = CompPkg {pkgName :: Text}
+  deriving (Eq, Ord, Show, Generic)
+
+type CompGrapher = LabeledGrapher CompPkg CompLabel
+
+data CompLabel
+  = CompVersion Text
+  | CompEnv DepEnvironment
+  deriving (Eq, Ord, Show, Generic)
```
scruffystuffs, line 99: We can drop this derived `Generic`.

zlav:
@scruffystuffs I can trim down the sample `composer.lock` file a bit more but I do like to leave in the unimportant fields at least once so we have reference to them and can ensure they don't fail the parser.

scruffystuffs:
With JSON, failing the parser from extra fields is not really a valid failure mode.  With custom formats, I would definitely agree.   As far as the future reference for fields we aren't using, I personally follow the YAGNI rule (You Ain't Gonna Need It) before everything else.  I wouldn't block a PR on this, but it might be a team discussion when we do our next spring cleaning of the project.

src/Strategy/Composer.hs
zlav: I fixed this to make it more clear. It is the dependency version.

src/Strategy/Composer.hs
```diff
@@ -0,0 +1,139 @@
...
+
+buildGraph :: ComposerLock -> Graphing Dependency
+buildGraph composerLock = run . withLabeling toDependency $ do
+  traverse_ (addDeps EnvProduction) $ lockPackages composerLock
+  traverse_ (addDeps EnvDevelopment) $ lockPackagesDev composerLock
+  where
+    addDeps :: Has CompGrapher sig m => DepEnvironment -> CompDep -> m ()
+    addDeps env dep = do
+      let pkg = CompPkg (depName dep)
+      _ <- M.traverseWithKey (addEdge pkg) (fromMaybe (M.fromList []) $ depRequire dep)
```
scruffystuffs, line 115: [nit]: We can use `Map.empty` here instead.

```suggestion
      _ <- M.traverseWithKey (addEdge pkg) (fromMaybe M.empty $ depRequire dep)
```

src/Strategy/Composer.hs
```diff
@@ -0,0 +1,139 @@
...
+
+    toDependency :: CompPkg -> Set CompLabel -> Dependency
+    toDependency pkg = foldr addLabel (start pkg)
+
+    addLabel :: CompLabel -> Dependency -> Dependency
+    addLabel (DepVersion ver) dep = dep {dependencyVersion = Just (CEq ver)}
+    addLabel (CompEnv env) dep = dep {dependencyEnvironments = env : dependencyEnvironments dep}
+
+    start :: CompPkg -> Dependency
+    start pkg =
```
scruffystuffs, line 131: [nit]: Not sure about the name `start`.  Since `toDependency` is already taken I think this could be merged into `toDependency`:

``` haskell
toDependency pkg = foldr addLabel $
  Dependency
    {  ...
    }
```

PR #122, opened 2020-08-25
Add Cabal analyzer

scruffystuffs:
Check comments for specific notes.  Other than those comments, this is pretty straightforward.

### DIFF
--- a/src/Strategy/Haskell/Cabal.hs
+++ b/src/Strategy/Haskell/Cabal.hs
@@ -153,7 +153,7 @@ shouldExclude :: InstallPlan -> Bool
 shouldExclude plan = not $ isDirectDep plan || planType plan == PreExisting
 
 ignorePlanId :: PlanId -> InstallPlan -> InstallPlan
-ignorePlanId _ b = b
+ignorePlanId = flip const
 
 buildGraph :: Has Diagnostics sig m => BuildPlan -> m (Graphing Dependency)
 buildGraph plan = do
@@ -176,5 +176,5 @@ toDependency plan =
 analyze :: (Has Exec sig m, Has ReadFS sig m, Has Diagnostics sig m) => Path Abs Dir -> m (Graphing Dependency)
 analyze dir = do
   _ <- execThrow dir cabalGenPlanCmd
-  plans <- readContentsJson@BuildPlan (dir </> cabalPlanFilePath)
+  plans <- readContentsJson @BuildPlan (dir </> cabalPlanFilePath)
   buildGraph plans
--- a/src/Strategy/Haskell/Cabal.hs
+++ b/src/Strategy/Haskell/Cabal.hs
@@ -63,8 +63,8 @@ instance FromJSON InstallPlan where
       <*> obj .: "id"
       <*> obj .: "pkg-name"
       <*> obj .: "pkg-version"
-      <*> (fromMaybe mempty <$> obj .:? "depends")
-      <*> (obj .:? "style" >>= parsePlanStyle)
+      <*> (obj .:? "depends" .!= S.empty)
+      <*> (obj .:? "style" >>= traverse parsePlanStyle)
       <*> (obj .:? "components" >>= mergeComponents)
 
 mergeComponents :: Maybe (Map Text (Map Text (Set Text))) -> Parser (Set PlanId)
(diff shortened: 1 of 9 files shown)


src/Strategy/Haskell/Cabal.hs
```diff
@@ -0,0 +1,180 @@
...
+    getDepends mapB = pure . S.map PlanId . fromMaybe mempty $ M.lookup "depends" mapB
+
+installPlanDepends :: InstallPlan -> Set PlanId
+installPlanDepends InstallPlan {..} =
+  if planType == PreExisting
+    then mempty
+    else planComponents <> planDepends
+
+isDirectDep :: InstallPlan -> Bool
+isDirectDep InstallPlan {..} = planStyle == Just Local && planType == Configured
```
scruffystuffs, line 86: Not sure on the naming here.  When running against spectrometer, this is all of the plans created by the `spectrometer.cabal` file.  Examples are the spectrometer lib, the 3 binaries, and the unit-test binary.

src/Strategy/Haskell/Cabal.hs
```diff
@@ -0,0 +1,180 @@
...
+  pure $ S.unions nested
+  where
+    getDepends :: Map Text (Set Text) -> Parser (Set PlanId)
+    getDepends mapB = pure . S.map PlanId . fromMaybe mempty $ M.lookup "depends" mapB
+
+installPlanDepends :: InstallPlan -> Set PlanId
+installPlanDepends InstallPlan {..} =
+  if planType == PreExisting
+    then mempty
+    else planComponents <> planDepends
```
scruffystuffs, line 83: We only care about a packages' deps if that package is not a base package.

src/Strategy/Haskell/Cabal.hs
```diff
@@ -0,0 +1,180 @@
...
+      <*> (obj .:? "components" >>= mergeComponents)
+
+mergeComponents :: Maybe (Map Text (Map Text (Set Text))) -> Parser (Set PlanId)
+mergeComponents Nothing = pure mempty
+mergeComponents (Just mapA) = do
+  nested <- traverse getDepends $ M.elems mapA
+  pure $ S.unions nested
+  where
+    getDepends :: Map Text (Set Text) -> Parser (Set PlanId)
+    getDepends mapB = pure . S.map PlanId . fromMaybe mempty $ M.lookup "depends" mapB
```
scruffystuffs, line 77: See [this example](https://github.com/fossas/spectrometer/blob/7268c12841d4f9ae8b0bf119650ea771a5ed7bfb/test/Haskell/testdata/cabal-plan.json#L33-L51) for why we need this Parser/Transformer.

src/Strategy/Haskell/Cabal.hs
```diff
@@ -0,0 +1,180 @@
...
+    BuildPlan <$> obj .: "install-plan"
+
+instance FromJSON InstallPlan where
+  parseJSON = withObject "InstallPlan" $ \obj ->
+    InstallPlan <$> (obj .: "type" >>= parsePlanType)
+      <*> obj .: "id"
+      <*> obj .: "pkg-name"
+      <*> obj .: "pkg-version"
+      <*> (fromMaybe mempty <$> obj .:? "depends")
+      <*> (obj .:? "style" >>= parsePlanStyle)
```
cnr, line 67: this is good. we can improve it slightly with `traverse`:
```suggestion
      <*> (obj .:? "style" >>= traverse parsePlanStyle)
```

..where traverse has this type:
```hs
traverse :: (Text -> Parser PlanStyle) -> Maybe Text -> Parser (Maybe PlanStyle)
```
..which simplifies `parsePlanStyle`:

```hs
parsePlanStyle :: Text -> Parser PlanStyle
parsePlanStyle "global" = pure Global
parsePlanStyle "local" = pure Local
parsePlanStyle style = fail "unknown install plan style: " ++ T.unpack style
```

turns out `parsePlanStyle` has a more-general type, which GHC will tell you if you remove the type signature:
```hs
parsePlanStyle :: MonadFail m => Text -> m PlanStyle
```

src/Strategy/Haskell/Cabal.hs
```diff
@@ -0,0 +1,180 @@
...
+isDirectDep InstallPlan {..} = planStyle == Just Local && planType == Configured
+
+parsePlanStyle :: Maybe Text -> Parser (Maybe PlanStyle)
+parsePlanStyle Nothing = pure Nothing
+parsePlanStyle (Just style) = case T.toLower style of
+  "global" -> pure $ Just Global
+  "local" -> pure $ Just Local
+  _ -> fail $ "unknown install plan style" ++ T.unpack style
+
+parsePlanType :: Text -> Parser PlanType
```
cnr, line 95: [nit] this has a more-general type, `parsePlanType :: MonadFail m => Text -> m PlanType`. this makes it easier to see at a glance what the function does

src/Strategy/Haskell/Cabal.hs
```diff
@@ -0,0 +1,180 @@
...
+    { cmdName = "cabal",
+      cmdArgs = ["v2-build", "--dry-run"],
+      cmdAllowErr = Never
+    }
+
+cabalPlanFilePath :: Path Rel File
+cabalPlanFilePath = $(mkRelFile "dist-newstyle/cache/plan.json")
+
+isCabalFile :: Path Abs File -> Bool
+isCabalFile file = is_v1 || is_v2
```
cnr, line 113: [nit] `is_v1`/`is_v2` might be confusing names to future-us; perhaps `isDotCabal` or `isCabalProject` or similar?

src/Strategy/Haskell/Cabal.hs
```diff
@@ -0,0 +1,180 @@
...
+doGraph plan = do
+  let parentId = planId plan
+  mapping parentId plan
+  for_ (installPlanDepends plan) $ \dep -> do
+    edge parentId dep
+    if isDirectDep plan
+      then direct dep
+      else pure ()
+
+shouldExclude :: InstallPlan -> Bool
```
cnr, line 152: `shouldExclude` seems to behave like a `shouldInclude` -- either rename or extract the `not` into the callsite below:

```hs
G.filter (not . shouldExclude) gr
```

src/Strategy/Haskell/Cabal.hs
```diff
@@ -0,0 +1,180 @@
...
+          dependenciesComplete = Complete
+        }
+
+doGraph :: Has (MappedGrapher PlanId InstallPlan) sig m => InstallPlan -> m ()
+doGraph plan = do
+  let parentId = planId plan
+  mapping parentId plan
+  for_ (installPlanDepends plan) $ \dep -> do
+    edge parentId dep
+    if isDirectDep plan
```
cnr, line 148: this is also called `when` from `Control.Monad`:
```hs
when (isDirectDep plan) $ direct dep
```

```hs
when :: Applicative f => Bool -> f () -> f ()
when True f = f
when False _ = pure ()
```

src/Strategy/Haskell/Cabal.hs
scruffystuffs: Yeah good point.  This was left over from a previous filtering attempt.

PR #124, opened 2020-08-25
Remove the Prologue module; remove default-extensions; use `sendIO` instead of `liftIO`

cnr:
As useful as `Prologue` is, it makes it difficult for new (and existing) contributors to figure out where things are coming from -- in particular hurting our ability to look up things in docs. With IDE integrations making it easier than ever to import things from modules, `Prologue` has outlived its usefulness.

---

Similarly, while having a bunch of `default-extensions` enabled, like `OverloadedStrings` and `LambdaCase`, is helpful to experienced folks, it's also harmful:

- figuring out how the language extensions interact + what language extensions are impacting a piece of code is sometimes non-trivial
- type inference becomes worse
- I'm more encouraged to write complicated haskell with a bunch of default extensions enabled

This also highlights some of the more-complicated parts of the codebase. It gives us a path to a common shared understanding of how the more-complicated parts work, through learning the language extensions involved

GHC/haskell-language-server will generally recommend safe language extensions required to compile.

---

There are two ways we can lift IO actions:
```hs
foo :: MonadIO m => m ()
foo = liftIO $ putStrLn "ok"
```

```hs
foo :: Has (Lift IO) sig m => m ()
foo = sendIO $ putStrLn "ok"
```

We generally prefer the `Lift IO` route, because it composes better with our other effects. This changes each use of `MonadIO` to `Has (Lift IO)`. I'll add an hlint rule soon to reflect this.


scruffystuffs:
Is there a reason we're combining the `MonadIO` to `Has LiftIO` with the removal of the prologue?

cnr:
The `liftIO` -> `sendIO` changes also modify imports of a ton of modules. I had to touch the imports of pretty much every module anyway, so it felt like something that could be done simultaneously, and with fewer merge conflicts this way

zlav:
I fully support this. Imports through Prologue were something that made the Haskell transition pretty difficult for me in the beginning so I am happy to see this being modified.

cnr:
Good call on enabling `OverloadedStrings` by default. Pushed a new commit with that change

PR #125, opened 2020-08-26
Log the best strategy name + dependency count for projects we find

cnr:
The output ends up looking like this:

```
Found python-requirements project at /Users/connor/Desktop/tmp5/home-assistant/ with 668 dependencies
Found python-requirements project at /Users/connor/Desktop/tmp5/home-assistant/homeassistant/ with 0 dependencies
Found python-requirements project at /Users/connor/Desktop/tmp5/home-assistant/homeassistant/components/device_tracker/ with 0 dependencies
Found python-requirements project at /Users/connor/Desktop/tmp5/imagededup/ with 0 dependencies
Found python-pipenv project at /Users/connor/Desktop/tmp5/streamlit/lib/ with 106 dependencies
Found python-pipenv project at /Users/connor/Desktop/tmp5/theHarvester/ with 46 dependencies
Found python-requirements project at /Users/connor/Desktop/tmp5/youtube-dl/ with 0 dependencies
```

We could definitely use more color in our output 🤔

---

minor refactor along the way (the first commit): projects always contain a non-empty list of related strategies. Using the proper type here -- `NonEmpty` from `Data.List.NonEmpty`, rather than the normal list type -- has a couple of advantages:

1. Things like sourceunit conversion are simplified (`toSourceUnit` no longer produces a `Maybe` type), meaning we can just `map` instead of `traverse`
2. Projects are much more "correct by construction" -- i.e. we can now expect that _every_ project has _at least_ one associated strategy

### DIFF
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -7,7 +7,7 @@ module App.Fossa.Analyze
   , ScanDestination(..)
   ) where
 
-import App.Fossa.Analyze.Project (Project(..), mkProjects)
+import App.Fossa.Analyze.Project (Project(..), ProjectStrategy(..), mkProjects)
 import App.Fossa.FossaAPIV1 (ProjectMetadata, UploadResponse (..), uploadAnalysis, uploadContributors)
 import App.Fossa.ProjectInference (inferProject, mergeOverride)
 import App.Types
@@ -22,7 +22,8 @@ import Control.Monad.IO.Class (MonadIO)
 import Data.Aeson ((.=))
 import qualified Data.Aeson as Aeson
 import Data.ByteString (ByteString)
-import Data.Foldable (traverse_)
+import Data.Foldable (for_, traverse_)
+import qualified Data.List.NonEmpty as NE
 import Data.Maybe (fromMaybe)
 import Data.Text (Text)
 import qualified Data.Text.Encoding as TE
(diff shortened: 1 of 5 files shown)


src/App/Fossa/Analyze.hs
```diff
@@ -108,7 +110,25 @@ analyze basedir destination override unpackArchives = runFinally $ do
...
+        bestStrategyDepCount :: Int
+        bestStrategyDepCount = Graphing.size . projStrategyGraph $ bestStrategy
+
+    logInfo $
+      "Found " <> pretty bestStrategyName
+        <> " project at "
+        <> pretty (fromAbsDir (projectPath project))
+        <> " with "
+        <> pretty bestStrategyDepCount
+        <> " dependencies"
```
scruffystuffs, line 130: I'd prefer to see this functionality a little more restructured so we can improve the color in the output later.  Some things I think would make this better long term are:

* `map` the projects into some data structure that represents the data we want to output.
* have some `render` function that turns that data structure into the properly-typed argument for `logInfo` (if there's already a typeclass, let's go with that)
* moving all of that into a separate function for readability.  Specifically, I think that `analyze` should only have 1 new line of code.

cnr:
@scruffystuffs how does the latest commit look

scruffystuffs:
@cnr 
giphub

PR #126, opened 2020-08-26
Add stack analyzer, fix Deptypes

scruffystuffs:
Very similar to #122 (Cabal analyzer), since stack wraps a lot of cabal-the-library functionality.

This ends up being much simpler than the associated analyzer in v1.x, because of the availability of `stack ls dependencies json`.

### DIFF
--- a/src/Strategy/Haskell/Stack.hs
+++ b/src/Strategy/Haskell/Stack.hs
@@ -1,0 +1,127 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Strategy.Haskell.Stack
+  ( discover,
+    -- * Testing
+    buildGraph,
+    PackageName (..),
+    StackDep (..),
+    StackLocationType (..),
+  )
+where
+
+import Control.Effect.Diagnostics
+import Data.Aeson.Types
+import Data.Foldable (for_, find)
+import qualified Data.Map.Strict as M
+import Data.Text (Text)
+import qualified Data.Text as T
+import Discovery.Walk
+import Effect.Exec
+import Effect.Grapher
+import qualified Graphing as G
+import Path
+import Types
+import Prelude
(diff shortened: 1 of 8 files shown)


src/Strategy/Haskell/Stack.hs
```diff
@@ -0,0 +1,127 @@
...
+  | Remote
+  | BuiltIn
+  deriving (Eq, Ord, Show)
+
+instance FromJSON StackDep where
+  parseJSON = withObject "StackDep" $ \obj ->
+    StackDep <$> obj .: "name"
+      <*> obj .: "version"
+      <*> obj .:? "dependencies" .!= []
+      <*> fmap unStackLocation (obj .:? "location" .!= StackLocation BuiltIn)
```
cnr, line 51: rather than using a `StackLocation` wrapper around `StackLocationType`, directly define a `FromJSON` instance on `StackLocationType`
```suggestion
      <*> obj .:? "location" .!= BuiltIn
```

src/Strategy/Haskell/Stack.hs
```diff
@@ -0,0 +1,127 @@
...
+      dependencyName = unPackageName $ stackName dep,
+      dependencyVersion = Just $ CEq $ stackVersion dep,
+      dependencyLocations = [],
+      dependencyEnvironments = [],
+      dependencyTags = M.empty
+    }
+
+buildGraph :: Has Diagnostics sig m => [StackDep] -> m (G.Graphing Dependency)
+buildGraph deps = do
+  result <- withMapping ignorePackageName $ traverse doGraph deps
```
cnr, line 121: I just remembered we have `fromEither :: (ToDiagnostic err, Has Diagnostics sig m) => Either err a -> m a` in the Diagnostics module:

```hs
result <- fromEither =<< withMapping ignorePackageName (traverse doGraph deps)
pure . G.gmap toDependency . G.filter shouldInclude $ result
```

unsure if this is too clever

PR #128, opened 2020-08-29
Add basic doc files

scruffystuffs:
* cargo
* carthage
* cocoapods
* node

Not sure if we're looking for more detail here, but this is probably a decent starting point to get a review.  We can decide that this needs to go a lot further.


cnr:
Merged these changes into the `docs/userguide` branch 🙏

PR #130, opened 2020-09-01
Carthage and cocoapods docs.

scruffystuffs:
Also, one additional change to RPM docs after a minor change earlier.

### DIFF
--- a/docs/strategies/ios.md
+++ b/docs/strategies/ios.md
@@ -1,0 +1,12 @@
+# iOS Analysis
+
+The iOS buildtool ecosystem consists of two major toolchains: `Carthage` and `Cocoapods`
+
+| Strategy | Direct Deps | Deep Deps | Edges | Tags |
+| --- | --- | --- | --- | --- |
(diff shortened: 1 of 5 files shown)


docs/strategies/ios.md
```diff
@@ -0,0 +1,12 @@
+# iOS Analysis
```
cnr, line 1: I'm conflicted on the naming here. Maybe `swift/objective-c` would be better here? In general we need to rethink strategy doc grouping, but that's out of scope for this PR in particular

docs/strategies/ios.md
```diff
@@ -0,0 +1,12 @@
...
+
+The iOS buildtool ecosystem consists of two major toolchains: `Carthage` and `Cocoapods`
+
+| Strategy | Direct Deps | Deep Deps | Edges | Tags |
```
cnr, line 5: I've been removing the `Tags` column from our other docs, as it's ill-defined. Let's remove it here too

PR #132, opened 2020-09-03
Implement erlang config DSL parser

scruffystuffs:
Syntax is JSON-like, including recursion, but without strongly-typed formatting guarantees.
Since Erlang terminates all statements with a `.`, we just treat a config file as a `.`-terminated list of tuples.


src/Strategy/Erlang/ConfigParser.hs
```diff
@@ -0,0 +1,51 @@
...
+  cc '{'
+  (Atom atom) <- parseAtom
+  cc ','
+  value <- parseErlValue
+  cc '}'
+  pure $ Pair atom value
+
+-- Char consumer (also consumes space)
+cc :: Char -> Parser ()
+cc = void <$> lexeme . char
```
cnr, line 45: what's the monomorphized type of `<$>` here? the answer may surprise you.

in ghci:
```hs
>>> :t void `_something` lexeme . char :: Char -> Parser ()
```

is there anything more familiar that has that type?

src/Strategy/Erlang/ConfigParser.hs
```diff
@@ -0,0 +1,51 @@
...
+parsePair = do
+  cc '{'
+  (Atom atom) <- parseAtom
+  cc ','
+  value <- parseErlValue
+  cc '}'
+  pure $ Pair atom value
+
+-- Char consumer (also consumes space)
+cc :: Char -> Parser ()
```
cnr, line 44: typically we use `symbol` from the lexer module for things like this:

```hs
symbol :: Text -> Parser Text
symbol = L.symbol scn
```

src/Strategy/Erlang/ConfigParser.hs
```diff
@@ -0,0 +1,51 @@
...
+
+data ErlValue
+  = Atom Text
+  | Array [ErlValue]
+  | Pair Text ErlValue
+
+parseConfig :: Parser [ErlValue]
+parseConfig = do
+  scn
+  many (parsePair <* char '.')
```
cnr, line 23: this probably isn't correct -- see https://stackoverflow.com/questions/1110601/in-erlang-when-do-i-use-or-or for what `.` and `,` mean in erlang. also: `sepBy` might help here

src/Strategy/Erlang/ConfigParser.hs
```diff
@@ -0,0 +1,51 @@
...
+parseErlValue = parseAtom <|> parseArray <|> parsePair
+
+parseAtom :: Parser ErlValue
+parseAtom = Atom <$> lexeme (takeWhile1P (Just "atom") (== ','))
+
+parseArray :: Parser ErlValue
+parseArray = lexeme (Array <$> between (char '[') (char ']') (many parseErlValue))
+
+parsePair :: Parser ErlValue
+parsePair = do
```
cnr, line 35: this is great, and you probably thought about using the applicative syntax to write this (`<*>`), but realized that you couldn't do so because of the punctuation you want to ignore

this is where a couple of other operators come in:

```hs
(<*) :: Applicative f => f a -> f b -> f a
(<*) = error "TODO: implement me using <$> and <*>"

(*>) :: Applicative f => f a -> f b -> f b
(*>) = error "TODO: implement me using <$> and <*>"
```

the "bird operators" have their beak pointing toward the result we want to keep, discarding the other one. they're left-associative (just like <*> and <$>) -- so you can pretend that there are parentheses wrapping the left side each time: `(((fa <* fb) *> fc) <*> fd)`.

when `f ~ Parser`, we can use these operators to run two parsers in sequence, discarding the result of one of them. for example, here:

```hs
parsePair :: Parser ErlValue
parsePair = pure Pair <* cc "{" <*> parseAtom <* cc "," <*> parseErlValue <* cc '}'
```

going from left-to-right, because the operators are left-associative:
```hs
pure Pair :: Parser (ErlValue -> ErlValue -> ErlValue)

-- ignoring the {, just keeping the Pair constructor we have so far
pure Pair <* cc "{" :: Parser (ErlValue -> ErlValue -> ErlValue)

-- applying the Pair constructor to the result of parseAtom
pure Pair <* cc "{" <*> parseAtom :: Parser (ErlValue -> ErlValue)

-- ignoring the ,
pure Pair <* cc "{" <*> parseAtom <* cc ',' :: Parser (ErlValue -> ErlValue)

-- applying the remaining Pair constructor to the result of parseErlValue
pure Pair <* cc "{" <*> parseAtom <* cc ',' <*> parseErlValue :: Parser ErlValue

-- ignoring the }
pure Pair <* cc "{" <*> parseAtom <* cc ',' <*> parseErlValue <* cc '}' :: Parser ErlValue
```

---

and if that wasn't enough, `pure anything <*>` has a name:

```hs
-- note the similarity between the types of <$ and <*
(<$) :: Functor f => a -> f b -> f a
x <$ f = error "TODO: implement me using <$>"
```

where `<$` again is kind of "pointing at the result we want to keep"

so our final parser:

```hs
parsePair :: Parser ErlValue
parsePair = Pair <$ cc "{" <*> parseAtom <* cc "," <*> parseErlValue <* cc '}'
```

src/Strategy/Erlang/ConfigParser.hs
```diff
@@ -0,0 +1,51 @@
...
+import Text.Megaparsec
+import Text.Megaparsec.Char
+import qualified Text.Megaparsec.Char.Lexer as L
+
+type Parser = Parsec Void Text
+
+data ErlValue
+  = Atom Text
+  | Array [ErlValue]
+  | Pair Text ErlValue
```
cnr, line 18: it looks like you can have arbitrary-length tuples in the dependencies input, not just length-2 text/value pairs

src/Strategy/Erlang/ConfigParser.hs
scruffystuffs: That was the most clickbait PR comment I've ever received.

src/Strategy/Erlang/ConfigParser.hs
scruffystuffs: Can you show or reference an example?  My understanding of erlang config syntax is that a curly-brace item is always a key-value pair.

cnr: oh whoops, I meant to link this: https://www.rebar3.org/docs/dependencies/ . they have a bunch of git dependencies with 3- or 4-tuples. tuples are the [curly brace syntax](https://www.tutorialspoint.com/erlang/erlang_tuples.htm)

src/Strategy/Erlang/ConfigParser.hs
scruffystuffs: Also, if you're specifically pointing me somewhere, I'm not seeing it.  I'm sure there's more than one way to do this, but I can't think of any way which is more readable, convenient, or general.  What am I missing here?

src/Strategy/Erlang/ConfigParser.hs
scruffystuffs: What's the benefit of doing this exact parser that way?  At first glance, that's less readable to me, but mostly because of the `<*>` operator.  I think the spacing helps with readability here.  Frankly, though, I could be convinced otherwise.

src/Strategy/Erlang/ConfigParser.hs
scruffystuffs: This does seem correct to me.  Every rebar example I can find terminates with a `.`.  Also, the link you posted is erlang syntax, rather than erlang config syntax, which seems to be a slightly altered subset of erlang syntax.  Functions, for example, are not allowed in erlang config.

src/Strategy/Erlang/ConfigParser.hs
cnr: The general benefit of using applicative syntax over monadic syntax is that the applicative syntax is _more restrictive_, so is more obviously-correct at a glance.

For example, if we're binding out values monadically, we can use those values to _change the behavior_ of our computation:

```hs
something = do
  foo <- bar
  case foo of
    ...
```

..so when I see a `do` block, I _have to_ carefully inspect it. I also have to inspect it for possible mistakes with binding variables and using the wrong variables later

With applicative, I can't do that. Applicative only allows us to "thread some structure" through a bunch of computations, without inspecting the values.

You can also play around with the spacing a bit for readability, e.g.,
```hs
parsePair :: Parser ErlValue
parsePair = Pair
  <$ cc "{"
  <*> parseAtom
  <* cc ","
  <*> parseErlValue
  <* cc '}'
```

src/Strategy/Erlang/ConfigParser.hs
cnr: oh you're right here -- I misunderstood what was happening. This is definitely real erlang syntax though, even if some things aren't allowed

scruffystuffs:
@zlav @cnr This is ready for review.  Test and build works locally, CI error doesn't seem related.

src/Strategy/Erlang/ConfigParser.hs
```diff
@@ -0,0 +1,159 @@
...
+
+type Parser = Parsec Void Text
+
+newtype AtomText = AtomText {unAtomText :: Text} deriving (Eq, Ord, Show, ToJSON)
+
+data ErlValue
+  = ErlAtom AtomText
+  | ErlString Text
+  | ErlInt Int
+  | ErlFloat Float
```
cnr, line 42: [nit] use `Double` here

src/Strategy/Erlang/ConfigParser.hs
```diff
@@ -0,0 +1,159 @@
...
+
+parseAtom :: Parser ErlValue
+parseAtom = ErlAtom <$> parseAtomText
+
+parseAtomText :: Parser AtomText
+parseAtomText = AtomText <$> lexeme (rawAtom <|> quotedAtom)
+  where
+    quotedAtom = enclosed "'" "'" $ takeWhile1P (Just "quoted Atom") (/= '\'')
+    --
+    rawAtom = T.cons <$> firstChar <*> rawAtomTail
```
cnr, line 132: [nit] some type annotations would help here

cnr (approved):
This is great. It's really cool that we can parse this entire format, and this should make the rebar-package-name-alias task really easy 🙏

PR #139, opened 2020-09-25
Merge rebar aliases into analyzer

scruffystuffs:
Minor changes required for the config parser, including a newtype for safety.

### DIFF
--- a/src/Strategy/Erlang/ConfigParser.hs
+++ b/src/Strategy/Erlang/ConfigParser.hs
@@ -13,8 +13,10 @@ module Strategy.Erlang.ConfigParser
     parseCharNum,
     parseIntLiteral,
     parseFloatLiteral,
+    atom,
     AtomText (..),
     ErlValue (..),
+    ConfigValues (..),
     intLiteralInBase,
     alphaNumToInt,
   )
@@ -34,6 +36,7 @@ import qualified Text.Megaparsec.Char.Lexer as L
 type Parser = Parsec Void Text
 
 newtype AtomText = AtomText {unAtomText :: Text} deriving (Eq, Ord, Show, ToJSON)
+newtype ConfigValues = ConfigValues {unConfigValues :: [ErlValue]} deriving (Eq, Ord, Show)
 
 data ErlValue
   = ErlAtom AtomText
@@ -57,8 +60,11 @@ instance ToJSON ErlValue where
 alphaNumSeq :: [Char]
 alphaNumSeq = ['0' .. '9'] <> ['A' .. 'Z']
 
-parseConfig :: Parser [ErlValue]
-parseConfig = scn *> parseTuple `endBy1` symbol "."
+atom :: Text -> ErlValue
+atom = ErlAtom . AtomText
+
+parseConfig :: Parser ConfigValues
+parseConfig = ConfigValues <$ scn <*> parseTuple `endBy1` symbol "."
 
 parseErlValue :: Parser ErlValue
 parseErlValue = parseErlArray <|> parseTuple <|> parseNumber <|> parseErlString <|> parseAtom
--- a/src/Strategy/Erlang/Rebar3Tree.hs
+++ b/src/Strategy/Erlang/Rebar3Tree.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE TemplateHaskell #-}
 
 module Strategy.Erlang.Rebar3Tree
   ( discover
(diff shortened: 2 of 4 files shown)


src/Strategy/Erlang/Rebar3Tree.hs
```diff
@@ -40,8 +44,37 @@ rebar3TreeCmd = Command
...
+    extract val aliasMap = aliasMap <> M.fromList (mapMaybe getAlias packages)
+      where
+        packages :: [ErlValue]
+        packages = case val of
+          ErlTuple [ErlAtom (AtomText "deps"), ErlArray deplist] -> deplist
+          _ -> []
+        
+        getAlias :: ErlValue -> Maybe (Text, Text)
+        getAlias erl = case erl of
+          ErlTuple [ErlAtom (AtomText realname), ErlString _, ErlTuple [ErlAtom (AtomText "pkg"), ErlAtom (AtomText alias)]] -> Just (realname, alias)
```
scruffystuffs, line 68: This pattern match is a beast, but it actually seems like the most idiomatic thing.  It's kind of a weird structure to find, functionally, so expressing it as-is seems to make sense.

src/Strategy/Erlang/ConfigParser.hs
```diff
@@ -57,8 +60,11 @@ instance ToJSON ErlValue where
 alphaNumSeq :: [Char]
 alphaNumSeq = ['0' .. '9'] <> ['A' .. 'Z']
 
-parseConfig :: Parser [ErlValue]
-parseConfig = scn *> parseTuple `endBy1` symbol "."
+atom :: Text -> ErlValue
```
zlav, line 63: [nit] What is an atom?
I ask because here `atom` sounds more like a type and something like `atomize` or `atomFromText` sounds a bit more descriptive.

test/Erlang/ConfigParserSpec.hs
```diff
@@ -104,19 +101,20 @@ spec = do
...
-          ErlInt 1234, -- Literal
-          ErlFloat 3.14159,
-          ErlInt 120, -- '$x'
-          ErlInt 35338, -- 'wes' in base 33
-          ErlArray [atom "arr1"],
-          ErlArray [ErlTuple [atom "key", ErlString "value"]],
-          ErlTuple [atom "number", ErlInt 5678] -- Literal
-        ]]
+        ConfigValues 
+          [ErlTuple [
```
zlav, line 105: [nit] Would it make sense to have multiple `ErlTuple` values in the `ConfigValues` array for the test now that we made that possible with the `ConfigValues` type?

src/Strategy/Erlang/Rebar3Tree.hs
```diff
@@ -68,13 +101,19 @@ buildGraph deps = unfold deps subDeps toDependency
                  , dependencyTags = M.empty
                  }
 
+lookupName :: Text -> M.Map Text Text -> Text
```
zlav, line 104: [nit] This function feels unnecessary to me but I don't feel strongly about it. If you do decide to keep it, you should move it up closer to where its used and maybe put it in a `where` block.

src/Strategy/Erlang/Rebar3Tree.hs
```diff
@@ -68,13 +101,19 @@ buildGraph deps = unfold deps subDeps toDependency
...
+
 data Rebar3Dep = Rebar3Dep
   { depName     :: Text
   , depVersion  :: Text
   , depLocation :: Text
   , subDeps     :: [Rebar3Dep]
   } deriving (Eq, Ord, Show)
 
+changeName :: Text -> Rebar3Dep -> Rebar3Dep
+changeName name dep = dep { depName = name }
```
zlav, line 115: [nit] I have the same opinions about this as with `lookupName`

zlav (approved):
Looks great! I have a few small nits but other than that the alias logic looks good.

test/Erlang/ConfigParserSpec.hs
scruffystuffs: Not sure I understand.  It already supports multiple `ErlValues`.

src/Strategy/Erlang/ConfigParser.hs
scruffystuffs: It's working as a smart constructor for an Atom.  In my mind it makes sense to name it like a constructor.  If you feel strongly I can change it to a more function-like name.

scruffystuffs:
Merging without resolving nits.  We can follow these up later @zlav

src/Strategy/Erlang/ConfigParser.hs
zlav: No that makes sense, I don't feel strongly about this at all.

test/Erlang/ConfigParserSpec.hs
zlav: Yea that makes sense, I was asking if we should modify the test values to contain multiple `ErlValues` so that we can test it

scruffystuffs: Ah, the real-life test already checks that.  It fails if the parser doesn't sit just right.  [Check here](https://github.com/fossas/spectrometer/pull/139/files#diff-502443b22c0ddb7b11e6fe0f961f459cR24)

PR #140, opened 2020-09-28
Analysis-target filtering

cnr:
This is a big one.

## Summary

See the [design doc](https://docs.google.com/document/d/1IgBwIKSQBphgXKCp6LJhsZkw_NH8c1q9W9mDJy7R5lc/edit#) for additional context.

- This PR adds a project-level analysis-target for all project types
- This PR adds support for subproject analysis-targets **for gradle only**. Further subproject support will come in the future.

### `fossa analyze --filter ANALYSIS-TARGET`

This allows for project-wide filtering or analysis-target-specific filtering. When more than one filter is specified, an analysis-target that passes any of the filters will be analyzed

### `fossa list-targets`

This command lists valid analysis-targets that can be used with `--filter`.

Example output:
```
Found project: maven@examples/example-gauth/
Found project: maven@examples/example-tls/
Found project: maven@examples/
Found project: gradle@./
Found target: gradle@./::grpc-all
Found target: gradle@./::grpc-alts
Found target: gradle@./::grpc-api
Found target: gradle@./::grpc-auth
Found target: gradle@./::grpc-benchmarks
Found target: gradle@./::grpc-bom
Found target: gradle@./::grpc-compiler
Found target: gradle@./::grpc-context
Found target: gradle@./::grpc-core
Found target: gradle@./::grpc-gae-interop-testing-jdk8
Found target: gradle@./::grpc-grpclb
Found target: gradle@./::grpc-interop-testing
Found target: gradle@./::grpc-netty
Found target: gradle@./::grpc-netty-shaded
Found target: gradle@./::grpc-okhttp
Found target: gradle@./::grpc-protobuf
Found target: gradle@./::grpc-protobuf-lite
Found target: gradle@./::grpc-services
Found target: gradle@./::grpc-stub
Found target: gradle@./::grpc-testing
Found target: gradle@./::grpc-testing-proto
```

### strategy-focused pipeline

Historically, we've had a very "strategy"-focused analysis pipeline, where we have a bunch of possibly-related strategies, that don't know about each other, that all go try to find and analyze projects independently.

For example, a yarn-lock strategy that only does static analysis on yarn.lock would be run independently of a yarn-command-running strategy. We'd pick the "best" strategy's dependency graph when deduplicating projects.

This has a couple of downsides:
- There are cases when we'd want to _merge_ dependency graphs between strategies, e.g., a setup.py and a requirements.txt strategy would want their dependency graphs unioned.
- This makes it harder to reason about _which_ dependency graph will be used

As part of these changes, we move toward a more project-focused analysis (e.g., "yarn"). This is critical for allowing users to select projects as analysis-targets

### subprojects

Subprojects are not a well-supported notion in our current datatypes, leading to the hacky workaround of a `SubprojectType` dependency type. We need to reify subprojects as a first-class construct, to both eliminate that tech debt, but also to allow users to select subprojects as analysis-targets.

These changes reify "BuildTargets" as part of the project, which are typically subprojects -- but in the future, things like bazel build targets will fit there too.

## Bonus improvements

- [x] combine setup.py and requirements.txt dependency graphs
- [x] use maven-pom project discovery for maven plugin strategy. this wildly improves discovery for maven-plugin-applicable projects
- [x] fix bug in gradle analyzer: "./gradlew" doesn't work as an executable, because exec on posix is relative to the working directory of the parent process

## Future tickets

Features
- analysis-target enumeration for subprojects
  - maven
  - ninja
  - nuget
  - cargo
  - sbt
- explain _why_ we found a project in `list-targets`
- add a subproject graph to dependency graph output && eliminate SubprojectType


src/App/Fossa/Analyze.hs
```diff
@@ -81,9 +62,52 @@ data ScanDestination
...
+  ( Has (Lift IO) sig m,
+    MonadIO m,
+    Has ReadFS sig m,
+    Has Exec sig m,
+    Has Logger sig m,
+    Has Diag.Diagnostics sig m
+  ) =>
+  -- | Discover functions
+  [Path Abs Dir -> m [NewProject]]
+discoverFuncs =
```
cnr, line 79: Not all of the analysis types have been migrated. That'll happen in the next day or two as I have time to cherry-pick out the commits

src/Control/Carrier/Output/Interpret.hs
```diff
@@ -0,0 +1,38 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
+
+module Control.Carrier.Output.Interpret
```
cnr, line 10: I think this is actually unused. Going to kill this

cnr:
A couple of additional notes:

- There's currently a ton of dead code, so this PR should become more balanced when I'm able to cull that (namely, after adding commits to migrate remaining strategies to the new structure && update pathfinder to use NewProject)
- naming of `NewProject` isn't permanent; same with `discover'`

src/Control/Effect/Diagnostics.hs
```diff
@@ -65,6 +71,14 @@ fromEither = either fatal pure
 fromEitherShow :: (Show err, Has Diagnostics sig m) => Either err a -> m a
 fromEitherShow = either (fatal . T.pack . show) pure
 
+-- | Lift a Maybe result into Diagnostics, with the given diagnostic thrown for @Nothing@
+fromMaybe :: (ToDiagnostic err, Has Diagnostics sig m) => err -> Maybe a -> m a
+fromMaybe msg = maybe (fatal msg) pure
+
+-- | Lift a Maybe result into Diagnostics, with a Text error message for @Nothing@
+fromMaybeText :: Has Diagnostics sig m => Text -> Maybe a -> m a
+fromMaybeText = fromMaybe
```
scruffystuffs, line 80: Why is this needed?

src/Data/Text/Extra.hs
```diff
@@ -1,12 +1,16 @@
...
+    strippedPrefix,
   )
 where
 
 import Data.Text (Text)
 import qualified Data.Text as T
 
+strippedPrefix :: Text -> Text -> Text
+strippedPrefix pre txt = maybe txt id (T.stripPrefix pre txt)
+
```
scruffystuffs, line 13: Doc would be helpful here, not immediately obvious what this function does.

src/Discovery/Filters.hs
```diff
@@ -0,0 +1,84 @@
...
+  | tool == tool',
+    dir == dir' =
+    Just targets
+  | otherwise = Nothing
+applyFilter (TargetFilter tool dir target) tool' dir' targets
+  | tool == tool',
+    dir == dir',
+    S.member target targets =
+    Just $ S.singleton target
+  | otherwise = Nothing
```
scruffystuffs, line 61: Has this been `ormulu`'d?  Looks a little weird to me.

src/Discovery/Filters.hs
```diff
@@ -0,0 +1,84 @@
...
+
+type Parser = Parsec Void Text
+
+filterParser :: Parser BuildTargetFilter
+filterParser = (try targetFilter <|> projectFilter) <* eof
+  where
+    targetFilter =
+      TargetFilter <$> buildtool <* char '@' <*> path <* char ':' <*> target
+    projectFilter =
+      ProjectFilter <$> buildtool <* char '@' <*> path
```
scruffystuffs, line 71: Not necessary for this PR, but we definitely need at least two more filters:
* `BuildtoolFilter`: `--filter maven` -- *This should be part of quickstart, since it takes little cognitive effort to add and understand.*
* `DirectoryFilter`: `--filter @src` -- *Multi-language, single-scan projects will probably prefer this.*

Alternatively, we can have more explicit, more easily parsed formats:

- `--filter path:path/to/src`
- `--filter tool:maven,gradle`
- `--filter target:maven@src:foo`
- `--filter project:maven@src

@zlav comments on this?

src/Srclib/Converter.hs
```diff
@@ -36,11 +35,10 @@ toSourceUnit Project {..} =
           }
     }
   where
-    bestStrategy = NE.head projectStrategies
-    renderedPath = Text.pack (toFilePath projectPath) <> "/" <> projStrategyName bestStrategy
+    renderedPath = Text.pack (toFilePath projectResultPath) <> "||" <> projectResultType
 
     graph :: Graphing Dependency
-    graph = projStrategyGraph bestStrategy
+    graph = projectResultGraph
```
scruffystuffs, line 41: Maybe remove this function?

src/Control/Effect/Diagnostics.hs
cnr: normally, because of `OverloadedStrings`, `fromMaybe "foo"` will have an ambiguous `IsString` constraint. We can force it by using a type application `fromMaybe @Text "foo"` or we can add this version that's monomorphic for `Text`

src/Discovery/Filters.hs
cnr: Yeah, `ormolu` caused this mess. This is not how I'd like to format this at all 😕

scruffystuffs: Bummer, looks like we'll have to get used to it.

src/Discovery/Filters.hs
cnr: I generally agree with wanting `maven` or `@src` filters, though I think it's outside the scope of these changes. Given the use-cases we have now, the project-level and analysis-target-level filters are enough, and we can add more later as necessary

scruffystuffs: Like I said, not necessary for this PR, but definitely needed later for usability.  This gets us to MVP with low-level control.  Once we've nailed that, we can wrap high-level wrappers later.

scruffystuffs (commented):
Overall approach and methodology looks good to me.  I do have quite a few comments though.  2 or 3 major, most of the rest are minor.

test/Discovery/FiltersSpec.hs
```diff
@@ -0,0 +1,62 @@
...
+import Test.Hspec.Megaparsec
+import Text.Megaparsec
+import Types (BuildTarget (..))
+
+spec :: Spec
+spec = do
+  describe "filterParser" $ do
+    it "should parse both types of projects" $ do
+      runParser filterParser "" "foo@bar" `shouldParse` (ProjectFilter "foo" $(mkRelDir "bar"))
+      runParser filterParser "" "foo@bar:baz" `shouldParse` (TargetFilter "foo" $(mkRelDir "bar") (BuildTarget "baz"))
```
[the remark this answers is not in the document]
scruffystuffs: Also, I'm not sure about the string format of `tool@path/to/project:buildtarget`.  I'd like to visit alternatives, or at least rule some obvious other choices out before sticking to this.  My main reasoning is that it has no semantic meaning unless you learn the format.  In which case, we must document the format.  Probably in the help of the `list-targets` and `analyze` commands.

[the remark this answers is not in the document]
cnr: short of using a structured format (e.g. json), I'm not sure we can come up with an intuitive format.

But I'm not sure it needs to be intuitive: the expected flow is to run `fossa list-targets`, then copy/paste those filters as arguments to `--filter`

---

there's a sort of mnemonic: "maven project at directory" (`maven@directory`); "maven project @ directory, specifically: buildtarget" (`maven@directory:buildtarget`)

as for why I chose those symbols:
- `@` isn't a valid character in analyzer names, which are alphanumeric, so using it as a delimiter between the analyzer name and project path is safe.
- `:` isn't a valid character in paths on windows or posix, so using it as a delimiter between the project path and the buildtarget name is safe

thus: `maven@foo@bar:some:project` is unambiguous: `maven`, `foo@bar`, `some:project`

src/Types.hs
```diff
@@ -119,6 +122,18 @@ toProjectClosure strategyGroup name body = ProjectClosure
   , closureLicenses = bodyLicenses body
   }
 
+-- TODO: results should be within a graph of build targets && eliminate SubprojectType
+data NewProject = NewProject
+  { projectType :: Text,
+    projectPath :: Path Abs Dir,
+    projectBuildTargets :: Set BuildTarget,
+    projectDependencyGraph :: Set BuildTarget -> DiagnosticsC IO (Graphing Dependency),
+    projectLicenses :: DiagnosticsC IO [LicenseResult]
```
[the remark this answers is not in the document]
cnr: We use licenses for pathfinder

src/Strategy/Python/Setuptools.hs
```diff
@@ -0,0 +1,73 @@
...
+discover' dir = map mkProject <$> findProjects dir
+
+findProjects :: MonadIO m => Path Abs Dir -> m [SetuptoolsProject]
+findProjects = walk' $ \dir _ files -> do
+  let reqTxtFiles =
+        filter
+          ( \f ->
+              "req" `isInfixOf` fileName f
+                && ".txt" `isSuffixOf` fileName f
+          )
```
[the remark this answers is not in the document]
cnr: Ormolu did this, because I inserted a line break (I thought the line was too long). The spacing is odd here, though I'm not sure let-bindings improve readability 😕. I'll see if I can improve the spacing somehow

src/Strategy/Maven.hs
```diff
@@ -0,0 +1,48 @@
...
+discover' dir = map mkProject <$> Pom.findProjects dir
+
+mkProject :: Pom.MavenProjectClosure -> NewProject
+mkProject closure = 
+  NewProject
+    { projectType = "maven",
+      projectPath = parent $ Pom.closurePath closure,
+      projectBuildTargets = mempty,
+      projectDependencyGraph = const . runReadFSIO . runExecIO $ getDeps closure,
+      projectLicenses = pure [] -- FIXME
```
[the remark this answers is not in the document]
cnr: I didn't re-add the pathfinder/license logic yet. I'll resolve this

src/Strategy/Cocoapods.hs
```diff
@@ -0,0 +1,62 @@
...
+import qualified Strategy.Cocoapods.PodfileLock as PodfileLock
+import Types
+
+discover' :: MonadIO m => Path Abs Dir -> m [NewProject]
+discover' dir = map mkProject <$> findProjects dir
+
+findProjects :: MonadIO m => Path Abs Dir -> m [CocoapodsProject]
+findProjects = walk' $ \dir _ files -> do
+  let podfile = find (\f -> fileName f == "Podfile") files
+  let podfileLock = find (\f -> fileName f == "Podfile.lock") files
```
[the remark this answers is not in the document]
cnr: yeah I'd be fine with exporting `findFileNamed name = find (\f -> fileName f == name)` from `Discovery.Walk`

test/Discovery/FiltersSpec.hs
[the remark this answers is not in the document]
scruffystuffs: `mkdir 'foo:bar'` works on linux.  I'm not sure about mac (based on BSD, not linux), but it definitely works on my system.  The only truly invalid path characters are `\0` for c-string termination and `/` for path separation.  Even invalid UTF-* characters are known to work against the kernel/fs.  POSIX rejects the bad characters by convention, but has no automatic rejection of these characters.

The point about copying from the previous command is quite a good point.  I'll spend a little more time thinking about it, but I'm certainly less skeptical from that explanation.

src/Discovery/Walk.hs
```diff
@@ -48,5 +54,21 @@ walk f = walkDir $ \dir subdirs files -> do
     WalkSkipAll -> pure $ WalkExclude subdirs
     WalkStop -> pure WalkFinish
 
+walk' ::
+  forall m o.
+  (MonadIO m, Monoid o) =>
+  (Path Abs Dir -> [Path Abs Dir] -> [Path Abs File] -> m (o, WalkStep)) ->
```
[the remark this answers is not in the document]
cnr: The type here is awkward for sure.

The difficulty is that we need to keep the existing `walk`, because it's still useful for more-complex discovery (see: Maven.Pom), so modifying the existing `WalkStep` datatype to add another parameter is something we probably shouldn't do -- and confusingly adding an additional `DiscoveryStep` type that's only used for one type of file walking is something I'd like to avoid. (I do like the smart constructors there though! That would be a good design if we went down that path)

I do think that the type signature of `walk'` is informative though, specifically the `Monoid` constraint:

```hs
walk' ::
  (MonadIO m, Monoid o) =>
  (Path Abs Dir -> [Path Abs Dir] -> [Path Abs File] -> m (o, WalkStep)) ->
  Path Abs Dir ->
  m o
```

notably, `walk'` returns `m o`, and the only thing we know about `o` is that it has a `Monoid` instance. so in the degenerate case, this is just `pure mempty`.

in the interesting case, because the "walkstep function" returns `o`s, the only interesting thing we can do with the `o`s is to smash them together (`<>`), and return the result of walking the filetree, smashing together the `o`s (`m o`)

src/Discovery/Projects.hs
```diff
@@ -0,0 +1,35 @@
...
+  [Path Abs Dir -> Diag.DiagnosticsC m [NewProject]] ->
+  -- | whether to unpack archives
+  Bool ->
+  Path Abs Dir ->
+  (NewProject -> m ()) ->
+  m ()
+withDiscoveredProjects discoverFuncs unpackArchives basedir f = do
+  for_ discoverFuncs $ \discover -> forkTask $ do
+    projectsResult <- Diag.runDiagnosticsIO (discover basedir)
+    Diag.withResult SevError projectsResult (traverse_ (forkTask . f))
```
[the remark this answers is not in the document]
cnr: this forks once to run the discovery function to produce projects (`projectsResult :: [DiscoveredProject]`), and then the supplied function (usually, dependency analysis) is run on each of the projects in parallel (`traverse_ (forkTask . f)`)

test/Discovery/FiltersSpec.hs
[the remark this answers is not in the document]
cnr: hrm. I could've sworn I tried `:` before, but it clearly works. that's annoying.

src/Strategy/Maven.hs
```diff
@@ -0,0 +1,48 @@
...
+import Control.Effect.Diagnostics
+import Control.Effect.Lift
+import Control.Monad.IO.Class
+import Effect.Exec
+import Effect.ReadFS
+import Graphing (Graphing)
+import Path
+import qualified Strategy.Maven.PluginStrategy as Plugin
+import qualified Strategy.Maven.Pom as Pom
+import qualified Strategy.Maven.Pom.Closure as Pom
```
scruffystuffs, line 17: These imports have the same name, we could probably rename the second to `PomClosure`.

src/Discovery/Walk.hs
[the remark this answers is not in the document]
scruffystuffs: If you're against adding a new `DiscoveryStep` type, then let's create smart constructors for the existing `(o, WalkStep)` tuple.  This allows us a smoother API into the admittedly, yet unavoidably, awkward type.

```haskell
discoveryContinue :: Monoid o => o -> (o, WalkStep)
discoveryContinue o = (o, WalkContinue)
```

src/Strategy/Cocoapods.hs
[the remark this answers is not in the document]
cnr: haskell-language-server has a new "fold definition" / "unfold definition" feature, which I couldn't quite get to work, but it uses `retrie` under the hood:

```sh
retrie --adhoc "forall name files. find (\f -> fileName f == name) files = findFileNamed name files"
```

turns out making that change across the entire codebase was really easy. the less easy part was going through each module to kill the "unused import" warning

test/Discovery/FiltersSpec.hs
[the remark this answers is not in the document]
scruffystuffs: I no longer think we need alternatives to this string format.  I do still think we need a few more tests here.

src/App/Fossa/Analyze.hs
scruffystuffs: Have the analyses all been migrated?  Are there any pending (not yet pushed to this PR) that you know of?

src/App/Fossa/Analyze.hs
```diff
@@ -81,9 +71,62 @@ data ScanDestination
...
+    Maven.discover',
+    Leiningen.discover',
+    Composer.discover',
+    Cabal.discover',
+    Stack.discover',
+    Yarn.discover',
+    Npm.discover',
+    Scala.discover',
+    RPM.discover',
+    RepoManifest.discover'
```
scruffystuffs, line 107: Have we decided on a name?  Are we sticking with `discover'`?

src/App/Fossa/Analyze.hs
```diff
@@ -81,9 +71,62 @@ data ScanDestination
...
+    Nothing -> logInfo $ "Skipping " <> pretty (projectType project) <> " project at " <> viaShow (projectPath project) <> ": no filters matched"
+    Just targets -> do
+      logInfo $ "Analyzing " <> pretty (projectType project) <> " project at " <> pretty (toFilePath (projectPath project))
+      graphResult <- sendIO . Diag.runDiagnosticsIO $ projectDependencyGraph project targets
+      Diag.withResult SevWarn graphResult (output . mkResult project)
+
+applyFiltersToProject :: Path Abs Dir -> [BuildTargetFilter] -> NewProject -> Maybe (Set BuildTarget)
+applyFiltersToProject basedir filters NewProject{..} =
+  case makeRelative basedir projectPath of
+    Nothing -> Just projectBuildTargets -- FIXME: this is required for --unpack-archives to continue to work
```
scruffystuffs, line 128: What is this FIXME referring to?  What is wrong that needs to be fixed?

src/Strategy/RPM.hs
```diff
@@ -58,6 +60,33 @@ discover = walk $ \dir _ files ->
...
+
+data RpmProject = RpmProject
+  { rpmFile :: Path Abs File
+  }
+  deriving (Eq, Ord, Show)
+
+mkProject :: RpmProject -> NewProject
+mkProject project =
+  NewProject
+    { projectType = "rpm3",
```
scruffystuffs, line 80: Why `rpm3`?

scruffystuffs (changes requested):
Needs some items addressed as already discussed, but overall PR is in good shape.  Last remaining bits after those changes is to confirm that all strategies have been migrated, check pathfinder still works (with licenses), check the fork safety issue you mentioned, and do a VERY short write up of how we tackle the easy discovery tests (in a later follow-up PR on the test repo, of course).

Additionally, this PR should *probably* update the GH actions workflow to fix our licensing issues.  Since the cabal strategy has been already migrated, there should be no blocker there.

src/App/Fossa/Analyze.hs
cnr: They're all migrated now

src/App/Fossa/Analyze.hs
cnr: Just renamed to `discover`

src/App/Fossa/Analyze.hs
cnr: Added more context here. Archives are unpacked to a temporary directory, so projectPath will be fail this `makeRelative`

src/Strategy/RPM.hs
cnr: whoops

src/Discovery/Archive.hs
```diff
@@ -1,58 +1,47 @@
...
 import Discovery.Walk
 import Path
 import qualified Path.IO as PIO
-import Strategy.Archive.RPM (extractRpm)
-import Types
 import Prelude hiding (zip)
 
--- Given a discover function to run over unarchived contents, discover projects
-discover :: HasDiscover sig m => (Path Abs Dir -> m ()) -> Path Abs Dir -> m ()
+-- FIXME: exception safety of forkTask? what happens when an exception is thrown in withArchive?
```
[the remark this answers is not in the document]
cnr: just manually tested this -- it turns out this behaves fine and has the exception-safety semantics I expected, but I'm glad I checked

PR #141, opened 2020-09-30
Infer detached head

scruffystuffs:
## Base Issue
We do not currently support git repositories in "detached HEAD" mode.  We require that a branch is present at all times, but there are several states where that would not be the case, including during dogfood builds.

## This PR
This PR is the "more correct" way of solving the problem, by populating the lack of branch all the way out to FOSSA Core.
The relevant types have been changed, and all breaking areas have been fixed.  Analysis Upload API call has been updated with the relevant information.  There are several known issues, which will be outlined in the comments, each matching the overall theme: **"Can we execute our entire process without a branch name?"**

## Alternatives
~#142 - The "less correct", but much simpler way.  Note that this PR needs some questions answered, and it may preclude the possibility of this method entirely.  The other PR is working and ready.~
^ closed since this one worked and is more correct.

### DIFF
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -158,17 +158,24 @@ tryUploadContributors baseDir baseUrl apiKey locator = do
   contributors <- fetchGitContributors baseDir
   uploadContributors baseUrl apiKey locator contributors
 
+-- This url can have a two forms (Core may allow more, but we don't care here):
+--    https://<fossa host>/projects/<project>/
+--    https://<fossa host>/projects/<project>/refs/branch/<branch>/<revision>
 fossaProjectUrl :: URI -> Text -> ProjectRevision -> Text
-fossaProjectUrl baseUrl rawLocator revision = URI.render baseUrl <> "projects/" <> encodedProject <> "/refs/branch/" <> branch <> "/" <> encodedRevision
+fossaProjectUrl baseUrl rawLocator revision = URI.render baseUrl <> "projects/" <> encodedProject <> buildSelector
   where
     Locator{locatorFetcher, locatorProject, locatorRevision} = parseLocator rawLocator
 
     underBS :: (ByteString -> ByteString) -> Text -> Text
     underBS f = TE.decodeUtf8 . f . TE.encodeUtf8
+    urlEncode' = underBS (urlEncode True)
 
-    branch = underBS (urlEncode True) (fromMaybe undefined (projectBranch revision))
-    encodedProject = underBS (urlEncode True) (locatorFetcher <> "+" <> locatorProject)
-    encodedRevision = underBS (urlEncode True) (fromMaybe "" locatorRevision)
+    encodedProject = urlEncode' (locatorFetcher <> "+" <> locatorProject)
+    encodedRevision = urlEncode' (fromMaybe "" locatorRevision)
+    -- | buildSelector is empty string unless we have a real branch to work with.
+    buildSelector = fromMaybe "" $ do
+      branch <- projectBranch revision
+      Just $ "/refs/branch/" <> urlEncode' branch <> "/" <> encodedRevision
 
 buildResult :: [Project] -> [ProjectFailure] -> Aeson.Value
 buildResult projects failures = Aeson.object
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -132,7 +132,7 @@ analyze basedir destination override unpackArchives = runFinally $ do
             [ "============================================================"
             , ""
             , "    View FOSSA Report:"
-            , "    " <> pretty (fossaProjectUrl baseurl (uploadLocator resp) (projectBranch revision))
+            , "    " <> pretty (fossaProjectUrl baseurl (uploadLocator resp) revision)
             , ""
             , "============================================================"
             ]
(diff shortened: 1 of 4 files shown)


src/App/Fossa/Analyze.hs
```diff
@@ -158,14 +158,15 @@ tryUploadContributors baseDir baseUrl apiKey locator = do
...
-fossaProjectUrl baseUrl rawLocator branch = URI.render baseUrl <> "projects/" <> encodedProject <> "/refs/branch/" <> branch <> "/" <> encodedRevision
+fossaProjectUrl :: URI -> Text -> ProjectRevision -> Text
+fossaProjectUrl baseUrl rawLocator revision = URI.render baseUrl <> "projects/" <> encodedProject <> "/refs/branch/" <> branch <> "/" <> encodedRevision
   where
     Locator{locatorFetcher, locatorProject, locatorRevision} = parseLocator rawLocator
 
     underBS :: (ByteString -> ByteString) -> Text -> Text
     underBS f = TE.decodeUtf8 . f . TE.encodeUtf8
 
+    branch = underBS (urlEncode True) (fromMaybe undefined (projectBranch revision))
```
scruffystuffs, line 169: How do we resolve this undefined?  The URL here requires a branch.

src/App/Fossa/ProjectInference.hs
```diff
@@ -78,7 +79,7 @@ inferSVN dir = do
               . dropPrefix rootRelativeToUrl
               . dropPrefix "^" $ relUrl
 
-        pure $ InferredProject root revision (if T.null trimmedRelative then "trunk" else trimmedRelative)
+        pure $ InferredProject root revision (Just $ if T.null trimmedRelative then "trunk" else trimmedRelative)
```
scruffystuffs, line 82: We hard-code `trunk` here for SVN.  Should we also remove this logic?

src/App/Fossa/ProjectInference.hs
```diff
@@ -104,7 +105,7 @@ inferDefault dir = sendIO $ do
   tmp <- getTempDir
   writeFile (fromAbsDir tmp FP.</> ".fossa.revision") (show time)
 
-  pure (InferredProject (T.pack name) (T.pack (show time)) "master")
+  pure (InferredProject (T.pack name) (T.pack (show time)) Nothing)
```
scruffystuffs, line 108: We previously hard-coded `master` here.  Is this safe to remove?

scruffystuffs:
How should this message be handled?  Remove it when branch is `Nothing`, or report detached HEAD state?
https://github.com/fossas/spectrometer/blob/a934c3497028f430cec4cc3a4a53fc923b68f042/src/App/Fossa/Analyze.hs#L124

src/App/Fossa/Analyze.hs
cnr: Good question. I wonder if there's a URL we can use that doesn't require a branch. If there isn't, we'll have to hardcode what core uses (`master`) as the default 😞

scruffystuffs: In which case, we should just resolve the `Maybe` branch name earlier in the chain, to take advantage of that default everywhere.

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -154,9 +154,11 @@ uploadAnalysis rootDir baseUri key ProjectRevision{..} metadata projects = fossa
           <> "v" =: cliVersion
           <> "managedBuild" =: True
           <> "title" =: fromMaybe projectName (projectTitle metadata)
-          <> "branch" =: projectBranch
           <> apiHeader key
           <> mkMetadataOpts metadata
+          <> case projectBranch of
```
cnr, line 159: it's annoying that the `Maybe` instance doesn't behave this way.

[nit] let's use `maybe :: b -> (a -> b) -> Maybe a -> b` here:
```suggestion
          <> maybe mempty ("branch" =:) projectBranch
```

scruffystuffs: I didn't like this code either, thanks for the suggestion.  I forgot to bring it up as a comment.

src/App/Fossa/ProjectInference.hs
cnr: Yes 👍

src/App/Fossa/ProjectInference.hs
cnr: Yes 👍

src/App/Fossa/Analyze.hs
```diff
@@ -158,16 +158,24 @@ tryUploadContributors baseDir baseUrl apiKey locator = do
   contributors <- fetchGitContributors baseDir
   uploadContributors baseUrl apiKey locator contributors
 
-fossaProjectUrl :: URI -> Text -> Text -> Text
-fossaProjectUrl baseUrl rawLocator branch = URI.render baseUrl <> "projects/" <> encodedProject <> "/refs/branch/" <> branch <> "/" <> encodedRevision
+-- This url can have a two forms (Core may allow more, but we don't care here):
+--    https://<fossa host>/projects/<project>/
+--    https://<fossa host>/projects/<project>/refs/branch/<branch>/<revision>
```
scruffystuffs, line 163: @cnr This is the alternative we were looking for.

src/App/Fossa/Analyze.hs
```diff
@@ -158,16 +158,24 @@ tryUploadContributors baseDir baseUrl apiKey locator = do
...
-    encodedProject = underBS (urlEncode True) (locatorFetcher <> "+" <> locatorProject)
-    encodedRevision = underBS (urlEncode True) (fromMaybe "" locatorRevision)
+    urlEncode' = underBS (urlEncode True)
+
+    encodedProject = urlEncode' (locatorFetcher <> "+" <> locatorProject)
+    encodedRevision = urlEncode' (fromMaybe "" locatorRevision)
+    -- | buildSelector is empty string unless we have a real branch to work with.
+    buildSelector = fromMaybe "" $ do
+      branch <- projectBranch revision
+      Just $ "/refs/branch/" <> urlEncode' branch <> "/" <> encodedRevision
```
scruffystuffs, line 179: I'm a little iffy on this `do` notation here.  Is there a better way to write this, or are the other ways more confusing?

cnr: I think this way is perfect. It's immediately clear to me what this is doing. the other common way is to use `maybe :: b -> (a -> b) -> Maybe a -> b`:
```hs
buildSelector = maybe ""
  (\branch -> "/refs/branch/" <> urlEncode' branch <> "/" <> encodedRevision)
  $ projectBranch revision
```

cnr (approved):
this is great 🙌

PR #148, opened 2020-10-19
Merge vpscli into fossa

scruffystuffs:
I would have liked to split this into multiple commits to make this easier, but due to the moving of files which changed significantly, that quickly became unreasonable.  So instead I will outline the major changes and where they'll be found (locations are omitted where obvious, or where module is mentioned):

* Moving `App.VPSScan.*` modules to `App.Fossa.VPS.*`.
* Altering the entry points of `App.Fossa.VPS.NinjaGraph` and `App.Fossa.VPS.Scan` to mesh with the `App.Fossa.Main` CLI handler.
  * Note that there is a pre-existing issue in the VPS modules where a type intended for deserializing CLI arguments is passed through the entire command like a god object.  For the CLI, we should separate Parsed types and Validated types, and use the individual subcommand entry points to do the `Parsed -> IO Validated` conversion (please don't use those names for real).  This is a tech debt item and not in scope for this change.
* Fixing a weird assumption about `upx` being installed to compress the vendored binaries.  `upx` is now optional and warns if it fails for any reason. (`vendor_download.sh`)
* Updating `Effect.Logger` to run as a two-stage fused-effect, like most other effects we have.  This is required for the `concurrently` call in `App.Fossa.VPS.Scan`.
* Removed `Trace` effect from `App.Fossa.VPS.*`.  Replaced all trace calls with `logDebug`, as well as removing uses of `String` types in those calls where possible.
  * Since I don't actually know how the original `Trace` actually worked, @kitified may need to confirm the expected logging/tracing behavior.  The main reason is that by default, `logDebug` will NOT be shown to the user at all.
* Removed the build for `vpscli` binary.  ~GH workflows may need to be adjusted to match this.  If so, that is in scope and should be handled in this PR.~  (`spectrometer.cabal` && `app/vpscli/Main.hs`)
  * GH Workflows have been adjusted to match the missing binary.
* Added aliases to general CLI fields that had VPS analogues. (`App.Fossa.Main`)
* Extended the `App.Fossa.Main` CLI handler to handle `vpscli` functionality under the `fossa vps` subcommand. (@kitified should we alias `analyze` to `scan`?  Or should we drop backwards compat to move to blessed 1.0 style?  Easy change if needed.)


kitified:
> Since I don't actually know how the original Trace actually worked, @kitified may need to confirm the expected logging/tracing behavior. The main reason is that by default, logDebug will NOT be shown to the user at all

Does the user have an easily-discoverable way to show this output?

> should we alias analyze to scan? Or should we drop backwards compat to move to blessed 1.0 style? Easy change if needed.

I wouldn't worry about keeping backwards compat; we're already breaking it with the use of `fossa vps` instead of `vpscli`. Are you proposing we rename `fossa vps analyze` to `fossa vps scan`?

scruffystuffs:
> Does the user have an easily-discoverable way to show this output?

The `--debug` flag is used for `fossa *` commands, and is expected to be the same for the vps commands.
> Are you proposing we rename `fossa vps analyze` to `fossa vps scan`?

I was suggesting that `fossa vps scan` redirect to the functionality of `fossa vps analyze`.  The commands would be identical.

kitified (approved):

kitified:
> I was suggesting that fossa vps scan redirect to the functionality of fossa vps analyze. The commands would be identical.

Got it. I think it's fine to just leave it as `fossa vps analyze`.

spatten (approved):
I agree with Kit about not being worried about backwards compatibility.

I'm fine with having the debug lines not showing up by default too.

The changes look good to me. Thanks for making my code better :)

src/App/Fossa/VPS/Scan.hs
```diff
@@ -0,0 +1,140 @@
...
+import Text.URI (URI)
+
+newtype SkipIPRScan = SkipIPRScan {unSkipIPRScan :: Bool}
+
+scanMain :: URI -> BaseDir -> ApiKey -> Severity -> OverrideProject -> FilterExpressions -> SkipIPRScan ->  IO ()
+scanMain baseuri basedir apikey logSeverity overrideProject fileFilters skipIprScan = do
+  let fossaOpts = FossaOpts baseuri $ unApiKey apikey
+      partVpsOpts = PartialVPSOpts fossaOpts (unSkipIPRScan skipIprScan) fileFilters
+
+  result <- runDiagnostics $ bracket extractEmbeddedBinaries cleanupExtractedBinaries $ vpsScan basedir logSeverity overrideProject partVpsOpts
```
cnr, line 36: while we're here, can we extract `bracket embeddedBinaries cleanupExtractedBinaries` into a function we export from `App.Fossa.VPS.EmbeddedBinary`?

cnr (approved):
looks good!

PR #154, opened 2020-11-02
Add support for gzip-compressed RPM files

cnr:
`codec-rpm`'s `payloadContentsC` function only supports `lzma`-compressed rpm files. This adds support for `gzip`-compressed rpm files as well.

### DIFF
--- a/src/Discovery/Archive/RPM.hs
+++ b/src/Discovery/Archive/RPM.hs
@@ -6,6 +6,7 @@ module Discovery.Archive.RPM
 where
 
 import qualified Codec.RPM.Conduit as RPM
+import qualified Codec.RPM.Types as RPMTypes
 import Conduit
 import Control.Effect.Lift
 import Control.Exception (throwIO)
@@ -13,6 +14,8 @@ import Control.Monad.Except
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Char8 as BS8
 import qualified Data.ByteString.Lazy as BL
+import qualified Data.Conduit.Lzma as Lzma
+import qualified Data.Conduit.Zlib as Zlib
 import qualified Data.CPIO as CPIO
 import Path
 import Prelude
@@ -29,7 +32,7 @@ extractRpm dir rpmFile = do
   where
     -- morally: :: Path b File -> ConduitT i CPIO.Entry m ()
     -- (has a couple of additional constraints, e.g., MonadError for attoparsec ParseError)
-    readRPMEntries file = sourceFileBS (toFilePath file) .| RPM.parseRPMC .| RPM.payloadContentsC
+    readRPMEntries file = sourceFileBS (toFilePath file) .| RPM.parseRPMC .| rpmPayload
     --
     sinkDir :: MonadIO m => ConduitT CPIO.Entry o m ()
     sinkDir = mapM_C $ \entry -> do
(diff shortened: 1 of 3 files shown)


src/Discovery/Archive/RPM.hs
```diff
@@ -14,30 +16,59 @@ import qualified Data.ByteString as BS
...
+    sendIO . runResourceT . runExceptT . runConduit $ do
+      sourceFileBS (toFilePath rpmFile) .| RPM.parseRPMC .| awaitForever (extractEntries dir)
+
   case res of
     Left parseErr -> sendIO . throwIO $ parseErr
     Right () -> pure ()
+
+-- | Extract RPM entries to a directory
+extractEntries :: (PrimMonad m, MonadThrow m, MonadIO m) => Path Abs Dir -> RPMTypes.RPM -> ConduitT i o m ()
+extractEntries dir rpm = yield rpm .| RPM.payloadC .| decompressorFor rpm .| CPIO.readCPIO .| filterC (not . CPIO.isEntryDirectory) .| sinkDir dir
```
cnr, line 38: We're using conduit here, which allows us to do _streaming_ decompression on the RPM entries. Some notes about reading conduit code:

Whenever you see a type like `ConduitT i o m a`, the type variables mean:
- `i` - the type of elements coming from the upstream
- `o` - the type of elements we're passing downstream
- `m` - the underlying monad (ConduitT is a monad transformer)
- `a` - the arbitrary result type of this computation (like any other monad)

The way to read `.|` is to think of it as a unix pipe. Each function in between is doing a transformation on things moving down the pipe

src/Discovery/Archive/RPM.hs
```diff
@@ -14,30 +16,59 @@ import qualified Data.ByteString as BS
...
 import qualified Path.IO as PIO
+import Prelude
 
 extractRpm :: Has (Lift IO) sig m => Path Abs Dir -> Path Abs File -> m ()
 extractRpm dir rpmFile = do
   res <-
-    sendIO . runResourceT . runExceptT . runConduit $
-      readRPMEntries rpmFile .| filterC (not . CPIO.isEntryDirectory) .| sinkDir
+    sendIO . runResourceT . runExceptT . runConduit $ do
+      sourceFileBS (toFilePath rpmFile) .| RPM.parseRPMC .| awaitForever (extractEntries dir)
```
zlav, line 30: Is this function in any danger of hanging with the use of `awaitForever`? I have seen us come across unsupported file formats when doing RPM extraction in the RPM fetcher that caused extraction to hang. I'm not aware of a better option here, but is this a valid concern? If so, it may be an acceptable failure mode for now.

zlav (approved):
This looks good, I don't have any large concerns, just one small question. Tests would be good to add in this repo or the integration tests repo as well.

src/Discovery/Archive/RPM.hs
cnr: `awaitForever` is a bit of a misnomer: it just repeatedly tries to `await` input from the upstream pipe, passing it to the provided function `(extractEntries dir)` until no more input is available.

`await` in conduit doesn't actually make a thread wait or anything; everything in a conduit happens in the same thread. `await` is just a way to pass control to the upstream until it can give you back an element, at which point control flow moves back to you

PR #156, opened 2020-11-09
fix: Embed vendored binaries before they are packaged for release

liftM:
The CI pipeline uploads artifacts in the `release` folder. However, the artifacts _in_ that folder were the ones generated _before_ embedding vendored binaries, not _after_ embedding vendored binaries. Therefore, the uploaded artifacts incorrectly vendored zero-byte binaries, causing the "no input" error.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,20 +87,6 @@ jobs:
       run: |
         cabal test --project-file=${{ matrix.project-file }} all
 
-    - name: Find and move binaries (windows)
-      if: ${{ contains(matrix.os, 'windows') }}
-      run: |
-        mkdir release
-        find . -type f -path '*/fossa/fossa.exe' -exec cp {} release \;
-        find . -type f -path '*/pathfinder/pathfinder.exe' -exec cp {} release \;
-  
-    - name: Find and move binaries (non-windows)
-      if: ${{ !contains(matrix.os, 'windows') }}
-      run: |
-        mkdir release
-        find . -type f -path '*/fossa/fossa' -exec cp {} release \;
-        find . -type f -path '*/pathfinder/pathfinder' -exec cp {} release \;
-
     - name: Update vendored binaries
       if: ${{ !contains(matrix.os, 'windows') }}
       run: ./vendor_download.sh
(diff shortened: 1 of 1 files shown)


scruffystuffs:
So, basically, this is just an order-of-operations bug after all?

liftM:
> So, basically, this is just an order-of-operations bug after all?

Yes. The original, incorrect order of operations:

1. Build CLI.
2. Copy CLI into `release`.
3. Build CLI with embedded vendored binaries.
4. Upload files in `release` folder.

The new, correct order of operations:

1. Build CLI.
2. Build CLI with embedded vendored binaries.
3. Copy CLI into `release`.
4. Upload files in `release` folder.

We have swapped (2) and (3).

kitified (approved):
Ah, nice catch!

liftM:
To acceptance test, check out the artifacts uploaded by the Actions run of this PR:

- [Linux binaries](https://github.com/fossas/spectrometer/suites/1476192822/artifacts/25463241)
- [MacOS binaries](https://github.com/fossas/spectrometer/suites/1476192822/artifacts/25463242)

I've validated that the Linux one works, although I'm seeing some really weird garbled output. @kitified, have you seen this before? IMO this is a different issue and not a blocker to merging this PR, but I'd like to know what you think before I merge.

kitified:
Acceptance test LGTM on MacOS.

I think generally the upload progress printout is a little strange- it outputs even when not running with `--debug`, and behaves oddly on various platforms, so that's probably the source of the output issue. I agree it's not a merge blocker and is pretty low priority in general: the VPS team is considering taking over the upload progress with a VPS CLI anyway.

scruffystuffs:
@liftM We are already aware of that IPR logging output issue.  We're still not sure why it's happening.  We think it's related to the terminal and not the app, since I can reproduce it in vscode but not alacritty

liftM:
I can reproduce it in Terminator. I think it's an issue with the actual CLI output - I bet something is going wrong with the progress indicator and the ANSI escape codes it's emitting.

liftM:
Look at the garbled output - stuff like `0c[m0pml` _looks_ like an ANSI escape sequence (which usually looks like `\033[XXXm`) - I wonder if output is being interleaved by parallel processes somewhere.

PR #158, opened 2020-11-17
Improve maven pom property interpolation

cnr:
Our maven static analysis strategy does naive property interpolation for dependency resolution, e.g.,

```xml
<dependency>
      <version>${my.dep.version}</version>
      [...]
</dependency>
```

would have `${my.dep.version}` replaced with the value of the `my.dep.version` property.

---

Our property interpolation breaks in a couple of common ways:

1. We require that the property be the entire string, e.g., `${property}` would be correctly interpolated, but `foo${property}` would not
2. There are commonly-used builtin properties used for dependency resolution, e.g., `project.groupId` or `project.version` that aren't supported by our interpolation method

This PR resolves both of these issues, and is tested against the user project in https://fossa.atlassian.net/browse/FC-2529

### DIFF
--- a/src/Strategy/Maven/Pom.hs
+++ b/src/Strategy/Maven/Pom.hs
@@ -3,6 +3,7 @@
 module Strategy.Maven.Pom
   ( analyze',
     getLicenses,
+    interpolateProperties,
   )
 where
 
@@ -116,7 +117,7 @@ buildProjectGraph closure = run . withLabeling toDependency $ do
 
         addDep :: Has MavenGrapher sig m => (Group, Artifact) -> MvnDepBody -> m ()
         addDep (group, artifact) body = do
-          let interpolatedVersion = classify . naiveInterpolate (pomProperties completePom) <$> depVersion body
+          let interpolatedVersion = classify . interpolateProperties completePom <$> depVersion body
               -- maven classifiers are appended to the end of versions, e.g., 3.0.0 with a classifier
               -- of "sources" would result in "3.0.0-sources"
               classify version = case depClassifier body of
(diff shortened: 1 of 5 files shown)


src/Strategy/Maven/Pom.hs
```diff
@@ -142,16 +143,39 @@ reifyDeps pom = M.mapWithKey overlayDepManagement (pomDependencies pom)
...
+computeBuiltinProperties pom = M.fromList
+  [ ("project.groupId", coordGroup (pomCoord pom))
+  , ("project.artifactId", coordArtifact (pomCoord pom))
+  , ("project.version", coordVersion (pomCoord pom))
+  ]
+
+interpolate :: Map Text Text -> Text -> Text
+interpolate properties text
+  | Just (beforeBegin, afterBegin) <- breakOnAndRemove "${" text
+  , Just (property, afterEnd) <- breakOnAndRemove "}" afterBegin =
```
scruffystuffs, line 163: Not sure I understand this syntax here.  Can you explain this?

cnr: so normally guards are used for boolean values:

```
foo thing
  | someCondition thing
  , andSomeOtherCondition thing = ...
  | otherwise = ...
```

but there's also something called a "pattern guard", which is essentially just pattern matching as a guard. in this case, I'm pattern matching that the results of both of these `breakOnAndRemove` are `Just ...`. If that condition fails for either one, it'll fall through to the next guard (the `| otherwise =`)

scruffystuffs: I think this might be more readable as a function on its own:
```haskell
splitPropertyString :: Text -> Maybe (Text, Text, Text)
splitPropertyString tx = _
```

where `splitPropertyString "a${b}c"` would return `Just ("a", "b", "c")`.

src/Strategy/Maven/Pom.hs
```diff
@@ -142,16 +143,39 @@ reifyDeps pom = M.mapWithKey overlayDepManagement (pomDependencies pom)
...
+-- 1. This removes the text that was broken on, e.g., `Text.breakOn "foo" "foobar" == ("", "foobar")` `breakOnAndRemove "foo" "foobar" == ("", "bar")`
+-- 2. This returns a `Maybe` value if the substring wasn't able to be found
+--
+-- >>> breakOnAndRemove "foo" "bazfoobar"
+-- Just ("baz","bar")
+--
+-- >>> breakOnAndRemove "foo" "bar"
+-- Nothing
+breakOnAndRemove :: Text -> Text -> Maybe (Text, Text)
+breakOnAndRemove needle haystack
```
scruffystuffs, line 178: I think this would fit better in `Data.Text.Extra`.  I could see myself re-using this.

test/Maven/PomStrategySpec.hs
```diff
@@ -0,0 +1,29 @@
...
+    it "should prefer user-specified properties over computed ones" $ do
+      let pom' = pom { pomProperties = M.singleton "project.groupId" "OTHERGROUP" }
+      interpolateProperties pom' "${project.groupId}" `shouldBe` "OTHERGROUP"
+
+    it "should work in the middle of strings" $ do
+      interpolateProperties pom "foo${project.groupId}bar" `shouldBe` "fooMYGROUPbar"
+
+    it "should interpolate multiple properties" $ do
+      interpolateProperties pom "${project.groupId}${project.artifactId}" `shouldBe` "MYGROUPMYARTIFACT"
+
```
scruffystuffs, line 29: Not sure if it's possible to add a test for missing properties, but if we can, we should.

scruffystuffs (approved):
Looks OK to merge as-is, but some small issues that would be good to handle if we can.

PR #161, opened 2020-11-19
Optional version field in Pipfile.lock

zlav:
This PR addresses an issue we have seen in the `Pipfile.lock` parser that did not handle an optional version field. This PR also includes expanded testing for pipenv that now uses an example file.

Example of the missing `version` field under `pkgFour`:
```
{
    "default": {
        "pkgThree": {
            "hashes": [
                "sha256:fc323ffcaeaed0e0a02bf4d117757b98aed530d9ed4531e3e15460124c106691",
                "sha256:84ab92ed1c4d4f16916e05906b6b75a6c0fb5db821cc65e70cbd64a3e2a5eaae"
            ],
            "version": "==3.0.0"
        },
        "pkgFour": {
            "hashes": [
                "sha256:2ccb79b01769d99115aa600d7eed99f524bf752bba8f041dc1c184853514655a",
                "sha256:0f2d585d22050e90c7d293b6451c83db097df77871974d90efd5a30dc12fcde3"
            ]
        }
    }
} 
```

### DIFF
--- a/src/Strategy/Python/Pipenv.hs
+++ b/src/Strategy/Python/Pipenv.hs
@@ -128,7 +128,7 @@ buildNodes PipfileLock{..} = do
                -> PipfileDep
                -> m ()
   addWithEnv env sourcesMap depName dep = do
-    let pkg = PipPkg depName $ T.drop 2 <$> fileDepVersion dep
+    let pkg = PipPkg depName (T.drop 2 <$> fileDepVersion dep)
     -- TODO: reachable instead of direct
     direct pkg
     label pkg (PipEnvironment env)
--- a/src/Strategy/Python/Pipenv.hs
+++ b/src/Strategy/Python/Pipenv.hs
@@ -91,7 +91,9 @@ buildGraph lock maybeDeps = run . withLabeling toDependency $ do
     start = Dependency
       { dependencyType = PipType
       , dependencyName = pipPkgName pkg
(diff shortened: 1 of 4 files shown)


src/Strategy/Python/Pipenv.hs
```diff
@@ -128,7 +128,7 @@ buildNodes PipfileLock{..} = do
                -> PipfileDep
                -> m ()
   addWithEnv env sourcesMap depName dep = do
-    let pkg = PipPkg depName (T.drop 2 (fileDepVersion dep))
+    let pkg = PipPkg depName $ T.drop 2 <$> fileDepVersion dep
```
scruffystuffs, line 131: [nit]: Not sure this `$` is the most readable due to the `<$>` operator.  I'll let you decide, but I think it might be more readable as this.
```suggestion
    let pkg = PipPkg depName (T.drop 2 <$> fileDepVersion dep)
```
That would prevent confusion about which operator applies where, which was a little confusing to me at first.  Normally, I prefer to use `.` and `$` liberally.  However, I think it's less readable in the presence of other infix operators.

scruffystuffs (approved):
Looks great, one optional readability nit.

src/Strategy/Python/Pipenv.hs
zlav: Yea I see what you're saying here, I agree. Thanks.

PR #162, opened 2020-11-23
Use real value in sourceunit type field

cnr:
Rather than using `"dummy-cli"` as the sourceunit type, use the real value from the analyzer name (e.g., `maven` or `gradle` or `setuptools`)

This also makes the `manifest` field a valid directory path

### DIFF
--- a/src/Srclib/Converter.hs
+++ b/src/Srclib/Converter.hs
@@ -24,7 +24,7 @@ toSourceUnit :: ProjectResult -> SourceUnit
 toSourceUnit ProjectResult{..} =
   SourceUnit
     { sourceUnitName = renderedPath,
-      sourceUnitType = SourceUnitTypeDummyCLI,
+      sourceUnitType = projectResultType,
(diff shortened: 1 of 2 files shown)


src/Srclib/Converter.hs
```diff
@@ -35,7 +35,7 @@ toSourceUnit ProjectResult{..} =
           }
     }
   where
-    renderedPath = Text.pack (toFilePath projectResultPath) <> "||" <> projectResultType
+    renderedPath = Text.pack (toFilePath projectResultPath)
```
scruffystuffs, line 38: Does this invalidate or fix the `TODO` on line 28:
```haskell
sourceUnitManifest = renderedPath, -- TODO: use a real value here instead of renderedPath?
```

cnr: It does 👍

zlav (approved):
Awesome! I didn't realize this would solve the confusion around `path||strategy` as well.

PR #164, opened 2020-12-01
Ignore Cargo dep source

scruffystuffs:
I can't actually reproduce this error, but apparently we don't even use the source field, so we can just remove that.

Fixes [this issue](https://github.com/fossas/spectrometer-tests/blob/959ef98c91fb539430be703fc29885fe61dd9bea/test/IntegrationSpec.hs#L185-L186)

### DIFF
--- a/src/Strategy/Cargo.hs
+++ b/src/Strategy/Cargo.hs
@@ -38,7 +38,7 @@ data PackageId = PackageId
 
 data PackageDependency = PackageDependency
   { pkgDepName :: T.Text
-  , pkgDepSource :: T.Text
+  , pkgDepSource :: Maybe T.Text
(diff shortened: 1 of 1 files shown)


cnr (approved):
🙌

liftM:
@scruffystuffs can you clarify "can't reproduce"? As in, the integration test at https://github.com/fossas/spectrometer-tests/blob/master/test/IntegrationSpec.hs#L186 succeeds for you?

scruffystuffs:
When I ran the older version of the code, I didn't hit the error.  When I read directly from the output of `cargo metadata` I didn't see the reported issue.  There may be some nondeterminism coming from cargo metadata.

PR #165, opened 2020-12-02
Switch to Wiggins for VPS Scans

kitified:
Switches to using the Wiggins CLI as an "unmanaged plugin", in other words a binary that is simply executed by `fossa` with the same arguments it is provided.

Removes all logic related to interacting with FOSSA services for VPS scan initiation or completion, instead all that is now handled by Wiggins.

Note: wiggins isn't _quite_ done yet, but won't significantly change at this point.


src/App/Fossa/VPS/Scan.hs
```diff
@@ -43,95 +39,22 @@ scanMain basedir apiOpts metadata logSeverity overrideProject fileFilters skipIp
...
-  logDebug "[All] Creating project in FOSSA"
-  _ <- context "creating project in FOSSA" $ createCoreProject vpsProjectName projectRevision metadata apiOpts
-
-  -- Create scan in SY
-  logDebug "[All] Creating scan in Scotland Yard"
-  let syOpts = ScotlandYardOpts locator projectRevision sherlockOrgId vpsOpts'
-  response <- context "creating scan ID" $ createScotlandYardScan apiOpts syOpts
-  let scanId = createScanResponseId response
+  ) => BaseDir -> Severity -> OverrideProject -> Flag SkipIPRScan -> Flag LicenseOnlyScan -> FilterExpressions -> ApiOpts -> ProjectMetadata -> BinaryPaths -> m ()
+vpsScan (BaseDir basedir) logSeverity overrideProject skipIprFlag licenseOnlyScan fileFilters apiOpts metadata binaryPaths = withLogQueue logSeverity $ \queue -> runLogger queue $ do
```
scruffystuffs, line 43: We can simplify the logging call here, since we're no longer using `concurrently`.

```suggestion
vpsScan (BaseDir basedir) logSeverity overrideProject skipIprFlag licenseOnlyScan fileFilters apiOpts metadata binaryPaths = withLogger logSeverity $ do
```

src/App/Fossa/VPS/Scan.hs
```diff
@@ -43,95 +39,22 @@ scanMain basedir apiOpts metadata logSeverity overrideProject fileFilters skipIp
...
-  case (iprResult, sherlockResult) of
-    (Right _, Right _) -> logDebug "[All] Scans complete"
-    (Left iprFailure, _) -> do
-      logDebug "[IPR] Failed to scan"
-      logDebug $ renderFailureBundle iprFailure
-      sendIO exitFailure
-    (_, Left sherlockFailure) -> do
-      logDebug "[Sherlock] Failed to scan"
-      logDebug $ renderFailureBundle sherlockFailure
+  wigginsResult <- sendIO $ runIt (runWiggins binaryPaths wigginsOpts)
```
scruffystuffs, line 50: Since we aren't using `concurrently` anymore, we can just use the current effect wrappers.  We can remove the entire `runIt` line, and this can probably become
```haskell
wigginsResult <- sendIO . runExecIO $ runWiggins binaryPaths wigginsOpts
```

.github/workflows/build.yml
```diff
@@ -76,33 +76,24 @@ jobs:
...
           ${{ runner.os }}-${{ matrix.ghc }}-
           ${{ runner.os }}-
+  
+    - name: Update vendored binaries
+      if: ${{ !contains(matrix.os, 'windows') }}
+      run: |
+        mkdir vendor
+        ./vendor_download.sh
+      env:
+        GITHUB_TOKEN: ${{ secrets.BASIS_ACCESS_TOKEN }}
```
scruffystuffs, line 86: REALLY happy to see a simplified build process. 👍

src/App/Fossa/VPS/Scan/Core.hs
```diff
@@ -2,33 +2,24 @@
...
-  , overrideScanFilters
-  , storeUpdatedScanFilters
+  , getSherlockInfo
   , Locator(..)
   , RevisionLocator(..)
   , SherlockInfo(..)
   )
 where
 
 -- I REALLY don't like mixing these modules together.
```
scruffystuffs, line 14: This comment is no longer relevant, which is just great.  Please take great pleasure in removing it.

src/App/Fossa/VPS/Scan/RunWiggins.hs
```diff
@@ -0,0 +1,86 @@
...
+      ++ optMaybeText "-link" projectLink
+      ++ optMaybeText "-policy" projectPolicy
+      ++ optMaybeText "-project-url" projectUrl
+      ++ optMaybeText "-team" projectTeam
+      ++ optMaybeText "-title" projectTitle
+      ++ optBool "-license-only" scanLicenseOnly
+      ++ optBool "-skip-ipr-scan" scanSkipIpr
+      ++ optBool "-debug" (logSeverity == SevDebug)
+      ++ optFilterExpressions fileFilters
+      ++ ["."]
```
scruffystuffs, line 55: [nit]: This seems a little simpler as a list literal.

```suggestion
  [ "analyze",
    "-endpoint", renderUri apiOptsUri,
    "-fossa-api-key", unApiKey apiOptsApiKey,
    optMaybeText "-name" overrideName,
    optMaybeText "-revision" overrideRevision,
    optMaybeText "-jira-project-key" projectJiraKey,
    optMaybeText "-link" projectLink,
    optMaybeText "-policy" projectPolicy,
    optMaybeText "-project-url" projectUrl,
    optMaybeText "-team" projectTeam,
    optMaybeText "-title" projectTitle,
    optBool "-license-only" scanLicenseOnly,
    optBool "-skip-ipr-scan" scanSkipIpr,
    optBool "-debug" (logSeverity == SevDebug),
    optFilterExpressions fileFilters,
    "."
  ]
```

src/App/Fossa/VPS/Scan/Core.hs
```diff
@@ -66,67 +51,13 @@ createRevisionLocator projectName organizationId revision = do
...
-completeProjectEndpoint baseurl = baseurl /: "api" /: "vendored-package-scan" /: "ci" /: "complete"
-
--- /api/vendored-package-scan/project-scan-filters/:locator
-projectScanFiltersEndpoint :: Url 'Https -> Locator -> Url 'Https
-projectScanFiltersEndpoint baseurl locator = baseurl /: "api" /: "vendored-package-scan" /: "project-scan-filters" /: unLocator locator
-
 {-
   FIXME: Every function below this line is using a data structure designed for the CLI.
   This tightly couples us to our CLI API, and is very tedious to change with the merge of `vpscli` and `fossa` exe's.
 -}
```
scruffystuffs, line 57: Since we removed the functions that this comment was referring to (except one), we can remove the comment as well.

scruffystuffs (changes requested):
Structurally looks solid to me.  The important changes are the logging/effect bits, and removing now-useless comments.

src/App/Fossa/VPS/Scan/RunWiggins.hs
kitified: This doesn't seem to work given that we are evaluating type `[Text]` from these `opt*` expressions. Am I missing something?

src/App/Fossa/VPS/Scan.hs
kitified: This doesn't seem to result in a `(Right, Left)` tuple, which I think we need for rendering the failure right?

src/App/Fossa/VPS/Scan/RunWiggins.hs
scruffystuffs: Ah, that's right.  I think there's a way to make this more readable, but it might take some thinking and I'd rather not block on it, it was a nit anyway.

src/App/Fossa/VPS/Scan.hs
scruffystuffs: Any errors in the execution should be "thrown" when we run the effect at the `runExecIO` call.  That will raise any errors into the `Diagnostics` effect we're currently working in (this function has the constraint `Has Diagnostics sig m`), and therefore has to be handled by the caller.  Since `runWiggins` actually returns `m ()`, this line could probably just be `runExecIO $ runWiggins binaryPaths wigginsOpts`, without `sendIO` or `result <-`.  Actually throwing the error should be lower in the call chain, probably in `execWiggins`, but reporting it to console is done in `scanMain` in this file.

kitified: Ok, that makes sense. Made this change and the one in your other outstanding comment and both error and non-error cases work great! Thanks!

src/App/Fossa/VPS/Scan.hs
kitified: Made this change and the one in your other outstanding comment and both error and non-error cases work great! Thanks!

scruffystuffs (approved):
I'm cool with this, but I'd personally prefer if @cnr reviewed as well before merging.

cnr (approved):
This looks great

Please make sure to resolve the endpoint-trailing-slash issue in wiggins; it feels like something that will bite us eventually if left unresolved

spatten (approved):
This looks good to me as well. So much deleted code!

kitified:
> This looks great
> 
> Please make sure to resolve the endpoint-trailing-slash issue in wiggins; it feels like something that will bite us eventually if left unresolved

This is actually resolved at this point in wiggins, thanks for the reminder- removed the workaround from spectrometer.

kitified:
Final wiggins changes are landed, merging!

PR #166, opened 2020-12-02
Friendly errors if analysis is empty

scruffystuffs:
Please review the error messages to help make them more clear.

If analysis returns no results (i.e. discovery found nothing):
```
$ fossa analyze --debug --endpoint http://localhost:9578
No projects were discovered
```

If analysis had some results, but the results were filtered out by the directory name:
```
$ fossa analyze --debug --endpoint http://localhost:9578
Analyzing setuptools project at /home/wes/dvlp/fake/docs/

Filtered out all 1 projects due to directory name
Excluded by directory name:  /home/wes/dvlp/fake/docs/
 ```

*Note that filtered project list is only available under debug logs.*

### DIFF
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -29,6 +29,7 @@ import Data.ByteString (ByteString)
 import Data.Flag (Flag, fromFlag)
 import Data.Foldable (traverse_)
 import Data.List (isInfixOf, stripPrefix)
+import qualified Data.List.NonEmpty as NE
 import Data.Maybe (fromMaybe)
 import Data.Set (Set)
 import Data.Text (Text)
@@ -74,6 +75,7 @@ import qualified Strategy.Rebar3 as Rebar3
 import qualified Strategy.RPM as RPM
 import qualified Strategy.Scala as Scala
 import qualified Strategy.Yarn as Yarn
+import System.Exit (exitFailure)
 import Text.URI (URI)
 import qualified Text.URI as URI
 import Types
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/App/Fossa/Analyze.hs [hunk 1]
```diff
@@ -180,33 +181,38 @@ analyze basedir destination override unpackArchives filters = do
   case destination of
     OutputStdout -> logStdout $ pretty (decodeUtf8 (Aeson.encode (buildResult projectResults)))
     UploadScan apiOpts metadata -> do
-      revision <- mergeOverride override <$> inferProject (unBaseDir basedir)
+      let maybeResults = NE.nonEmpty projectResults
+      case maybeResults of
+        Nothing -> logError "No analysis to upload"
```
zlav, line 186: I think it would be helpful if we added some more information here like "the fossa cli did not detect any dependencies in this directory", this message right now is a little unclear about whats happening and I don't think all uses will know what "No analysis" means.

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -147,8 +148,12 @@ uploadAnalysis rootDir apiOpts ProjectRevision{..} metadata projects = fossaReq
           <> mkMetadataOpts metadata projectName
           -- Don't include branch if it doesn't exist, core may not handle empty string properly.
           <> maybe mempty ("branch" =:) projectBranch
-  resp <- req POST (uploadUrl baseUrl) (ReqBodyJson sourceUnits) jsonResponse (baseOpts <> opts)
-  pure (responseBody resp)
+  -- We do more filtering above and therefore must recheck for empty upload body
+  if null filteredProjects
+    then fatalText "No projects to upload (analyzed projects were filtered out by directory name)"
```
zlav, line 153: Can we output the list of filtered projects here? This way users will know if they should instruct the cli to not filter them.

zlav (changes requested):
Left a comment about improving the output message to users, all in all this looks great though!

src/App/Fossa/FossaAPIV1.hs
scruffystuffs: They can't instruct the CLI not to filter them.  This filtering is based on the projects being in a `docs/` directory, for example.  This is completely unrelated to build-target filtering.  We don't currently support checking those folders from outside, and this wouldn't be the PR to address that.

I can log the relative paths to debug, and change this message to point to debug logs.

src/App/Fossa/Analyze.hs
scruffystuffs: I agree with the point, but it's not "dependencies", it's really analyses.  In this state we couldn't find anything that we could even **attempt** to analyze.  If we analyzed, but found 0 dependencies, this message isn't shown.

zlav (approved):
Looks great! Thanks for addressing the concerns

PR #169, opened 2020-12-11
Update with new Wiggins release

kitified:
Updates `fossa` with a newly released version of the Wiggins CLI.
Comment added so that this Spectrometer release doesn't have the same git hash as previous.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+# v2.4.5
+
+- Update Wiggins CLI plugin to version `2020-12-11-5d581ea`
+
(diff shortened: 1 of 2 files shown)


scruffystuffs (changes requested):
Can you update the changelog instead of this comment?  That should still cause a rebuild, since we don't cache build artifacts in CI.  Also, We should have a changelog entry noting that we've updated wiggins, so it should work out fine anyway.

zlav (approved):
Thanks for adding the changelog!

PR #170, opened 2020-12-16
Fix `fossa test`/`fossa vps test` output for push-only API keys

cnr:
When a push-only API key is used, the `issues` field on the response is always empty. Instead, we need to check the `count` field.


zlav (approved):
Looks good! Great log message as well.

PR #171, opened 2020-12-16
Gomod: ignore local modules in a multi-module project

cnr:
The `replace` directive in gomodules projects is problematic: in addition to its typically-bad use, it's also used to emulate multi-project structures if the replace target is a _directory_, rather than the usual package/version combo:

```
replace foo => bar v2 // typical, bad
replace foo => ../foo // multi-project!
```

the Right Way™️ to fix this is to do what we've done with maven projects previously, which is to construct multi-project "closures" and extract dependency graphs from those. I've done this for gomodules projects on another branch: https://github.com/fossas/spectrometer/tree/go-multi-module-projects

This is a simpler approach. It handles the vast majority of usecases seen in the wild. We improve our fallback static-analysis strategy by ignoring local packages.

The best-analysis case (via running commands with the `go` cli) was already covered in a previous PR.

### DIFF
--- a/test/Go/testdata/go.mod.edgecases
+++ b/test/Go/testdata/go.mod.edgecases
@@ -20,6 +20,8 @@ replace repo/B => alias/repo/B v0.1.0
 replace (
 	repo/C => alias/repo/C v0.0.0-20180207000608-000000000003
 	repo/E => alias/repo/E v0.0.0-20170808103936-000000000005+incompatible
+    foo v0 => ../foo
+    bar => /foo/bar/baz
     repo/F_underscore v1.0.0 => repo/F_underscore v2.0.0
 )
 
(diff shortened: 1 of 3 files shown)


test/Go/testdata/go.mod.edgecases
```diff
@@ -20,6 +20,8 @@ replace repo/B => alias/repo/B v0.1.0
 replace (
 	repo/C => alias/repo/C v0.0.0-20180207000608-000000000003
 	repo/E => alias/repo/E v0.0.0-20170808103936-000000000005+incompatible
+    foo v0 => ../foo
+    bar => /foo/bar/baz
```
scruffystuffs, line 24: [nit] Are the differences in leading whitespace significant?  If not, can we normalize them?

scruffystuffs (approved):
👍

test/Go/testdata/go.mod.edgecases
cnr: it was completely unintentional, but they're accidentally significant insofar as proving our parser is whitespace-insensitive -- so I guess I'll keep them

PR #172, opened 2020-12-16
Never prune dependencies for graphs that don't have nodes marked as direct

cnr:
At some point, we became much more strict about pruning "unreachable" dependencies in our graphs. This unfortunately breaks the analyzers that don't necessarily mark dependencies as direct -- because often times, e.g., for `yarn.lock`, we don't know which dependencies are direct.

Fixes https://github.com/fossas/issues/issues/335


zlav (approved):
Looks good to me! Glad this got caught.

PR #173, opened 2020-12-17
Container scan (Analyze/Test w/ upload)

scruffystuffs:
Fixes fossas/issues#312
Fixes fossas/issues#313
Fixes fossas/issues#314
Fixes fossas/issues#315

See fossas/issues#312 for more info.

Note that this is unlikely to merge until testing is complete, but is ready for review.


src/App/Fossa/Container.hs
```diff
@@ -0,0 +1,198 @@
...
+extractRevision :: OverrideProject -> ContainerScan -> ProjectRevision
+extractRevision OverrideProject {..} ContainerScan {..} = ProjectRevision name revision Nothing
+  where
+    name = fromMaybe imageTag overrideName
+    revision = fromMaybe imageDigest overrideRevision
+
+
+toContainerScan :: Has Diagnostics sig m => SyftResponse -> m ContainerScan
+toContainerScan SyftResponse {..} = do
+  let convertArtifact (ResponseArtifact a b c d e f) = ContainerArtifact a b c d e f
```
cnr, line 169: [no action] I hate it but I like that we're keeping the internal type separate from the transport type

src/App/Fossa/Container/Test.hs
```diff
@@ -0,0 +1,80 @@
...
+  logInfo ("Using project revision: `" <> pretty (projectRevision revision) <> "`")
+
+  logSticky "[ Waiting for build completion ]"
+  waitForBuild apiOpts revision
+
+  logSticky "[ Waiting for issue scan completion ]"
+  issues <- waitForIssues apiOpts revision
+  logSticky ""
+
+  if null $ issuesIssues issues
```
cnr, line 74: Make sure to update this logic to reflect the new way we check for issues in `fossa test` or `fossa vps test`

src/App/Fossa/EmbeddedBinary.hs
```diff
@@ -0,0 +1,125 @@
...
+import Path
+import Path.IO
+import Prelude hiding (writeFile)
+
+-- data BinaryPaths = BinaryPaths
+-- { binaryPathContainer :: Path Abs Dir
+-- , wigginsBinaryPath :: Path Abs File
+-- }
+
+data PackagedBinary
```
cnr, line 32: This is a clever way of structuring this, and is probably good. I'm going to suggest another similar way to structure this, but don't feel obligated to change anything.

Rather than using `PackagedBinary` as a key for other functions like `extractedPath`, we can directly include that information in a record.

I would use this structure for PackagedBinary:
```hs
data PackagedBinary
  { packagedBinaryName :: Text
  , packagedBinaryBS :: ByteString
  }
```

and then we could replace, e.g.,
```hs
embeddedBinaryWiggins :: ByteString
```

with 
```hs
embeddedBinaryWiggins :: PackagedBinary
```

---

I understand this loses us the clever `allBins` implementation, and compile-time warnings if we forget to include something in there, so I'm kind of split.

src/App/Fossa/EmbeddedBinary.hs
```diff
@@ -0,0 +1,125 @@
...
+import Control.Effect.Exception (bracket)
+import Control.Effect.Lift (Has, Lift, sendIO)
+import Control.Monad.IO.Class
+import Data.ByteString (ByteString, writeFile)
+import Data.FileEmbed.Extra
+import Path
+import Path.IO
+import Prelude hiding (writeFile)
+
+-- data BinaryPaths = BinaryPaths
```
cnr, line 27: [nit] kill this comment block

src/App/Fossa/Main.hs
```diff
@@ -100,6 +105,29 @@ appMain = do
           unless vpsReportJsonOutput $ die "report command currently only supports JSON output.  Please try `fossa report --json REPORT_NAME`"
           baseDir <- validateDir vpsReportBaseDir
           VPSReport.reportMain baseDir apiOpts logSeverity vpsReportTimeout vpsReportType override
+    --
+    ContainerCommand ContainerOptions {..} -> do
+      dieOnWindows "container scanning"
```
cnr, line 110: 😢

src/App/Fossa/Main.hs
```diff
@@ -100,6 +105,29 @@ appMain = do
...
+            else do
+              apikey <- requireKey maybeApiKey
+              let apiOpts = ApiOpts optBaseUrl apikey
+              ContainerAnalyze.analyzeMain (UploadScan apiOpts containerMetadata) logSeverity override containerAnalyzeImage
+        ContainerTest ContainerTestOptions {..} -> do
+          apikey <- requireKey maybeApiKey
+          let apiOpts = ApiOpts optBaseUrl apikey
+          ContainerTest.testMain apiOpts logSeverity containerTestTimeout containerTestOutputType override containerTestImage
+    --
+    DumpBinsCommand dir -> do
```
cnr, line 124: This is a great idea to add

cnr (approved):
This looks great to me. A couple of minor comments, but nothing major

src/App/Fossa/EmbeddedBinary.hs
scruffystuffs: Hmm, I'm not sure I understand the advantage of doing it the way you proposed.  Can you explain a little more in detail **why** we might want to do it that way?

src/App/Fossa/Main.hs
scruffystuffs: I think windows support is later.

src/App/Fossa/Container.hs
scruffystuffs: I hated this too, and so I made it more explicit.

src/App/Fossa/Container.hs
```diff
@@ -0,0 +1,208 @@
...
+      }
+
+instance FromJSON ResponseArtifact where
+  parseJSON = withObject "ResponseArtifact" $ \obj ->
+    ResponseArtifact <$> obj .: "name"
+      <*> obj .: "version"
+      <*> obj .: "type"
+      <*> obj .: "purl"
+      <*> obj .: "metadataType"
+      -- We delete "files" as early as possible, which reduces
```
zlav, line 76: Could you elaborate on what a file is? I get why we need a comment here but I don't really understand what it is saying.

src/App/Fossa/Container/Test.hs
```diff
@@ -0,0 +1,84 @@
...
+  waitForBuild apiOpts revision
+
+  logSticky "[ Waiting for issue scan completion ]"
+  issues <- waitForIssues apiOpts revision
+  logSticky ""
+
+  case issuesCount issues of
+    0 -> logInfo "Test passed! 0 issues found"
+    n -> do
+      logError $ "Test failed. Number of issues found: " <> pretty n
```
zlav, line 77: I know you were mirroring what we do in `fossa test` right now, but I feel like this should only print if the user has a full API token. Right now I think the user will see the following if they have a push only token.
```
Test failed. Number of issues found: Check webapp for more details, or use a full-access API key (currently using a push-only API key)
```
We should probably split this up.

scruffystuffs: Why should this be different from `fossa test`?  Also, why should we hide the number of issues when the count is always sent back?

src/App/Fossa/Container.hs
scruffystuffs: There's some output from the `syft` binary that is under the nested json key `artifacts[n].metadata.files`.

```json5
{
  "artifacts": [
    {
      "metadata": {
        "files": [
          // enormous array of info we don't need
        ]
      }
    }
  ]
}
```

Since this value is unnecessary AND huge, we take steps to prevent loading it into memory as much as possible.  Since haskell lazily evaluates data by default, and the `Data.Map` and `Data.Map.Lazy` internal map structure are strict in keys, but lazy in values, we can potentially remove the files key from this map without ever trying to hold the entire array in memory at once.

The comment is saying "we retain the json-like map structure, but we want to ignore the 'files' field to save a ton of memory."

PR #178, opened 2021-01-20
Enable container scanning

scruffystuffs:
Fixes during E2E testing:

* Send `managedBuild` query param.
* Send the full payload over network.
* Use the common build link logic


microsoftly (dismissed):
can this also include the `image` key in the payload ?

```typescript
type ContainerScanUploadBody = {	
  image: {	 
    os: string; // distro -> name	  
    osRelease: string; // distro -> version	  
    artifacts: Array<SyftApkArtifact | SyftDpkgArtifact | SyftRpmdbArtifact>;	
  };	
}
```

cnr (approved):

PR #179, opened 2021-01-21
v1 Compatibility mode

zlav:
This PR introduces a `fossa compatibility` command which runs `fossa analyze` with the legacy fossa cli binary. The main use case of this feature is to allow users to access fossa cli v1's archive upload feature.

### DIFF
--- a/src/App/Fossa/Compatibility.hs
+++ b/src/App/Fossa/Compatibility.hs
@@ -3,7 +3,7 @@
 
 
 module App.Fossa.Compatibility
-  ( main,
+  ( compatibilityMain,
     argumentParser,
     Argument,
   )
@@ -25,10 +25,10 @@ type Argument = Text
 argumentParser :: Parser Argument
 argumentParser = pack <$> argument str (metavar "IMAGE" <> help "The image to scan")
 
-main ::
+compatibilityMain ::
   [Argument] ->
   IO ()
-main args = withLogger SevInfo . runExecIO . withCLIv1Binary $ \v1Bin -> do
+compatibilityMain args = withLogger SevInfo . runExecIO . withCLIv1Binary $ \v1Bin -> do
   logSticky "[ Waiting for fossa command completion ]"
   cmd <- exec [reldir|.|] $ v1Command v1Bin $ args
   logSticky ""
--- a/src/App/Fossa/Compatibility.hs
+++ b/src/App/Fossa/Compatibility.hs
@@ -1,0 +1,51 @@
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+
+module App.Fossa.Compatibility
+  ( main,
+    argumentParser,
+    Argument,
+  )
+where
+
+import App.Fossa.EmbeddedBinary (BinaryPaths, toExecutablePath, withCLIv1Binary)
+import Control.Effect.Lift (sendIO)
+import Data.Text (Text, pack)
+import Effect.Exec (CmdFailure(cmdFailureStdout), AllowErr (Never), Command (..), exec, runExecIO, cmdFailureStderr)
+import Path
+import qualified Data.ByteString.Lazy.Char8 as BL
+import Options.Applicative (Parser, argument, help, metavar, str)
+import System.Exit (exitFailure, exitSuccess)
+import Data.Text.Lazy.Encoding
+import Effect.Logger (Pretty(pretty), logInfo, logSticky, Severity(SevInfo), withLogger)
+
+type Argument = Text
+
+argumentParser :: Parser Argument
+argumentParser = pack <$> argument str (metavar "IMAGE" <> help "The image to scan")
+
+main ::
+  [Argument] ->
+  IO ()
+main args = withLogger SevInfo . runExecIO . withCLIv1Binary $ \v1Bin -> do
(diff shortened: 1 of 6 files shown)


src/App/Fossa/Compatibility.hs
```diff
@@ -0,0 +1,51 @@
...
+
+type Argument = Text
+
+argumentParser :: Parser Argument
+argumentParser = pack <$> argument str (metavar "IMAGE" <> help "The image to scan")
+
+main ::
+  [Argument] ->
+  IO ()
+main args = withLogger SevInfo . runExecIO . withCLIv1Binary $ \v1Bin -> do
```
scruffystuffs, line 31: Let's use a less ambiguous name, maybe `compatibilityMain`.  If we refactor the CLI side without qualified imports, this will be very confusing.

src/App/Fossa/Main.hs
```diff
@@ -216,7 +221,13 @@ hiddenCommands =
           ( info
               (ContainerCommand <$> containerOpts)
               (progDesc "Run in Container Scan mode")
-    )
+          )
+        <> command
+          "compatibility"
+          ( info
+              (CompatibilityCommand <$> compatibilityOpts)
+              (progDesc "Run fossa cli v1. Supply arguments as \"fossa compatibility -- analyze --project test\"")
```
scruffystuffs, line 229: Are we allowing people to run any v1 command here?  Is that why we're making this change?

vendor_download.sh
```diff
@@ -91,6 +91,34 @@ done
 rm $SYFT_RELEASE_JSON
 echo "Forked Syft download successful"
 
+echo "Downloading cliv1 binary"
```
scruffystuffs, line 94: Let's add an empty `echo` here for clearer logging separation.

src/App/Fossa/Main.hs
zlav: Correct, the main goal is to let them run archive scanning, but by allowing them to run anything we prevent ourselves from getting hung up in the future.

src/App/Fossa/Compatibility.hs
zlav: Good call, I was trying to not be redundant calling`Compatibility.compatibilityMain` but I overlooked being explicit with the import.

src/App/Fossa/Main.hs
scruffystuffs: I don't know if I agree with this.  I think I'd rather only allow certain options, or people will just continue using v2 as v1 in other scenarios.  I think this makes it much harder to deprecate v1 eventually.

scruffystuffs: Maybe it's reasonable to force the existence of `analyze` and whichever option triggers archive scanning.

zlav: I think forcing `fossa analyze` is reasonable. I really don't have a strong feeling about this, but I would prefer to be more restrictive to start, so I'll enforce this.

src/App/Fossa/Compatibility.hs
```diff
@@ -0,0 +1,51 @@
...
+  IO ()
+compatibilityMain args = withLogger SevInfo . runExecIO . withCLIv1Binary $ \v1Bin -> do
+  logSticky "[ Waiting for fossa analyze completion ]"
+  cmd <- exec [reldir|.|] $ v1Command v1Bin $ args
+  logSticky ""
+
+  case cmd of
+    Left err -> do
+      _ <- logInfo $ pretty $ decodeUtf8 $ cmdFailureStderr err
+      _ <- logInfo $ pretty $ decodeUtf8 $ cmdFailureStdout err
```
scruffystuffs, line 39: [nit]: We can simplify this a little (this makes it clear that we are applying the same pipeline to both, rather than the pipeline being the same coincidentally)
```suggestion
      for_ [cmdFailureStderr, cmdFailureStdout] $
        \accessor -> logInfo . pretty . decodeUtf8 $ accessor err
```
Or 

```suggestion
      traverse_ (\accessor -> logInfo . pretty . decodeUtf8 $ accessor err)  [cmdFailureStderr, cmdFailureStdout]
```

src/App/Fossa/Compatibility.hs
```diff
@@ -0,0 +1,51 @@
...
+      sendIO exitFailure
+    Right out -> do 
+      _ <- sendIO $ BL.putStr out
+      sendIO exitSuccess
```
scruffystuffs, line 43: [nit]:  Multiple `sendIO`s looks weird to me.  This is just preference, so up to you if you like it better either way.
```suggestion
    Right out -> sendIO (BL.putStr out >> exitSuccess)
```

src/App/Fossa/Compatibility.hs
```diff
@@ -0,0 +1,51 @@
...
+
+v1Command :: BinaryPaths -> [Text] -> Command
+v1Command bin args =
+  Command
+    { cmdName = pack . toFilePath $ toExecutablePath bin,
```
scruffystuffs, line 48: [no action]: `pack . toFilePath $ toExecutablePath` should probably be a utility of the `EmbeddedBinary` module.  I'll PR that later if I remember it.

scruffystuffs (approved):
Nits only at this point, nothing I would object to.  Nice work!

PR #181, opened 2021-01-26
Build urls with default branch

scruffystuffs:
Acceptance Criteria:

- In project with detached HEAD, we should build a url using the `master` branch.  This behavior mimics core's "default to `master`" behavior.


zlav (approved):
Looks great! Tested it and it works!

PR #182, opened 2021-01-26
Fix revision read

scruffystuffs:
We now properly read a cached `.fossa.revision` file.

There may still be a bug in `inferDefault` to not override that file, but that requires a few days to refactor and fix.  This should at least fix the issue where `fossa test` would NEVER work when using a non-VCS and non-CLI-option revision string.


zlav (approved):
Works as expected, looks good to me!

scruffystuffs:
fixes fossas/issues#304

PR #186, opened 2021-02-01
Create SAML link when available

scruffystuffs:
Significant points which should be validated:

- Link building always succeeds, even on IO failure.
- Diag helpers are good abstractions and not overkill or incorrectly formed.
- Analyze function is semantically unchanged, with exception of recovering from failed contrib upload and new url builder.

### DIFF
--- a/src/App/Fossa/API/BuildLink.hs
+++ b/src/App/Fossa/API/BuildLink.hs
@@ -19,10 +19,6 @@ import Network.HTTP.Types (urlEncode)
 import Srclib.Types (Locator (..))
 import qualified Text.URI as URI
 
--- If SAML is not used, neither is orgID (link is created from the locator)
-defaultOrg :: Organization
-defaultOrg = Organization 0 False
-
 fossaProjectUrlPath :: Locator -> ProjectRevision -> Text
 fossaProjectUrlPath Locator {..} ProjectRevision {..} = "/projects/" <> encodedProject <> buildSelector
   where
@@ -38,12 +34,10 @@ getFossaBuildUrl :: (Has Diagnostics sig m, Has (Lift IO) sig m) => ProjectRevis
 getFossaBuildUrl revision apiopts locator = do
   maybeOrg <- recover $ getOrganization apiopts
 
-  let org = fromMaybe defaultOrg maybeOrg
-      baseURI = apiOptsUri apiopts
-      relUriPath =
-        if orgUsesSAML org
-          then samlUrlPath org locator revision
-          else fossaProjectUrlPath locator revision
+  let baseURI = apiOptsUri apiopts
+      relUriPath = case maybeOrg of
+        Just org | orgUsesSAML org -> samlUrlPath org locator revision
+        _ -> fossaProjectUrlPath locator revision
   pure (URI.render baseURI <> relUriPath)
 
 samlUrlPath :: Organization -> Locator -> ProjectRevision -> Text
--- a/src/App/Fossa/API/BuildLink.hs
+++ b/src/App/Fossa/API/BuildLink.hs
@@ -1,0 +1,56 @@
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RecordWildCards #-}
+
+module App.Fossa.API.BuildLink
+  ( getFossaBuildUrl,
+    samlUrlPath,
+  )
+where
+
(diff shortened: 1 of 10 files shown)


scruffystuffs:
Fixes fossas/issues#347

src/App/Fossa/API/BuildLink.hs
```diff
@@ -0,0 +1,56 @@
...
+    encodedBranch = urlEncode' $ fromMaybe "master" projectBranch
+    buildSelector = "/refs/branch/" <> encodedBranch <> "/" <> encodedRevision
+
+getFossaBuildUrl :: (Has Diagnostics sig m, Has (Lift IO) sig m) => ProjectRevision -> ApiOpts -> Locator -> m Text
+getFossaBuildUrl revision apiopts locator = do
+  maybeOrg <- recover $ getOrganization apiopts
+
+  let org = fromMaybe defaultOrg maybeOrg
+      baseURI = apiOptsUri apiopts
+      relUriPath =
```
cnr, line 43: I'd like to eliminate `defaultOrg` here -- maybe we can use a guard. something like:

```hs
relUriPath =
  case maybeOrg of
    Just org | orgUsesSaml org -> samlUrlPath org locator revision
    _ -> fossaProjectUrlPath locator revision

File: src/App/Fossa/Analyze.hs [hunk 2]
```diff
@@ -199,25 +202,19 @@ analyze basedir destination override unpackArchives filters = do
...
-              , ""
-              , "============================================================"
-              ]
-            traverse_ (\err -> logError $ "FOSSA error: " <> viaShow err) (uploadError resp)
-
-            contribResult <- Diag.runDiagnostics $ runExecIO $ tryUploadContributors (unBaseDir basedir) apiOpts (uploadLocator resp)
-            case contribResult of
-              Left failure -> logDebug (Diag.renderFailureBundle failure)
-              Right _ -> pure ()
+        uploadResult <- uploadAnalysis apiOpts revision metadata someProjects
```
cnr, line 205: [no action] This looks so much better 🙌

src/Control/Carrier/Diagnostics.hs
```diff
@@ -75,6 +82,29 @@ renderFailureBundle FailureBundle {..} =
...
+    Left failure -> logErrorBundle failure >> pure Nothing
+    Right success -> Just <$> logResultWarnings success
+
+-- | Run a void Diagnostic effect into a logger, using the default error/warning renderers.
+-- | Exits with non-zero if the result is a failure.
+-- | Useful for setting up diagnostics from CLI entry points.
+logDiagnostic_ :: (Has (Lift IO) sig m, Has Logger sig m) => DiagnosticsC m () -> m ()
+logDiagnostic_ diag = logDiagnostic diag >>= maybe (sendIO exitFailure) pure
+  
+withDiagnosticLogger_ :: Has (Lift IO) sig m => Severity -> DiagnosticsC (LoggerC m) () -> m ()
```
cnr, line 105: I think this one might be overkill

src/Control/Carrier/Diagnostics.hs
```diff
@@ -75,6 +82,29 @@ renderFailureBundle FailureBundle {..} =
...
+logDiagnostic diag = do
+  result <- runDiagnostics diag
+  case result of
+    Left failure -> logErrorBundle failure >> pure Nothing
+    Right success -> Just <$> logResultWarnings success
+
+-- | Run a void Diagnostic effect into a logger, using the default error/warning renderers.
+-- | Exits with non-zero if the result is a failure.
+-- | Useful for setting up diagnostics from CLI entry points.
+logDiagnostic_ :: (Has (Lift IO) sig m, Has Logger sig m) => DiagnosticsC m () -> m ()
```
cnr, line 102: I like this. I wonder if there's a name we can use to better signal "this will immediately terminate the program when it encounters a failure"

cnr (approved):
Looks good, and the diagnostics helpers are a huge improvement

src/Control/Carrier/Diagnostics.hs
scruffystuffs: I had thought about `logDiagnostic!` as an alternative

scruffystuffs: For now, `Diag.logWithExit_` will do.

PR #187, opened 2021-02-03
Organization: make usesSAML field optional when deserializing

cnr:


zlav:
If we can add tests for this API deserialization that would be really helpful in preventing errors like this in the future.

cnr:
I'm not sure how to best handle cases like this -- because this deserialization will always work on up-to-date fossa instances. The issue we're seeing is from customers with out-of-date onprem instances

scruffystuffs:
I thought I added this into the original PR.  I think I forgot to re-add after fixing some stuff up.

scruffystuffs (approved):
Looks Good

PR #188, opened 2021-02-03
Split inferProject into inferProjectFromVCS + inferProjectDefault + saveRevision; add inferProjectCached

cnr:
This fixes a bug where a missing `.fossa.revision` file would prevent a successful analysis, and correctly overlays "override" information from the CLI args

Fixes fossas/issues#353

### DIFF
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -14,7 +14,7 @@ module App.Fossa.Analyze
 import App.Fossa.Analyze.GraphMangler (graphingToGraph)
 import App.Fossa.Analyze.Project (ProjectResult(..), mkResult)
 import App.Fossa.FossaAPIV1 (UploadResponse (..), uploadAnalysis, uploadContributors)
-import App.Fossa.ProjectInference (inferProject, mergeOverride)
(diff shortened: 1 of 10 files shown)


src/App/Fossa/Analyze.hs
```diff
@@ -194,7 +195,8 @@ analyze basedir destination override unpackArchives filters = do
     FoundSome someProjects -> case destination of
       OutputStdout -> logStdout . pretty . decodeUtf8 . Aeson.encode . buildResult $ NE.toList someProjects
       UploadScan apiOpts metadata -> do
-        revision <- mergeOverride override <$> inferProject (unBaseDir basedir)
+        revision <- mergeOverride override <$> (inferProjectFromVCS basedir <||> inferProjectDefault basedir)
+        saveRevision revision
```
cnr, line 199: The trick is that we use `(inferProjectFromVCS basedir <||> inferProjectDefault basedir)` followed by a `saveRevision` in places where we do analysis, and `(inferProjectFromVCS basedir <||> inferProjectCached basedir <||> inferProjectDefault basedir)` elsewhere

scruffystuffs (approved):
Great change!  Removing part of the mess that was here and fixing a critical issue.

PR #193, opened 2021-02-21
Create and use URI helpers

scruffystuffs:
_Note: Please leave review comments for missing docs and code comments, I don't think we should leave this as code-only._

In short, the most painful points of working with `modern-uri` are the lack of setters and the use of `RText`.  By creating our own simple structs, and setters which use those new structs, we can solve both problems.  We can also compose multiple setter actions with `>>=` (can't use `fmap` because of `MonadThrow` constraint on the `RText` smart constructors).

Additionally, because our new setters take `[a]` as input (where `a` is one of our simpler types), we get `Semigroup/Monoid` instances for free, and well as literal building syntax.

### DIFF
--- a/test/App/Fossa/API/BuildLinkSpec.hs
+++ b/test/App/Fossa/API/BuildLinkSpec.hs
@@ -2,18 +2,18 @@
 
 module App.Fossa.API.BuildLinkSpec (spec) where
 
-import Test.Hspec
 import App.Fossa.API.BuildLink
-import Srclib.Types (Locator(Locator))
-import App.Fossa.FossaAPIV1 (Organization(Organization))
-import App.Types (ProjectRevision(ProjectRevision))
-import Data.Text (Text)
-import Fossa.API.Types
-import Text.URI.QQ
-import Effect.Logger (IgnoreLoggerC, ignoreLogger)
+import App.Fossa.FossaAPIV1 (Organization (Organization))
+import App.Types (ProjectRevision (ProjectRevision))
 import Control.Carrier.Diagnostics (DiagnosticsC, logDiagnostic)
-import Data.Functor.Identity (Identity(runIdentity))
 import Control.Monad (join)
+import Data.Functor.Identity (Identity (runIdentity))
+import Data.Text (Text)
+import Effect.Logger (IgnoreLoggerC, ignoreLogger)
+import Fossa.API.Types
+import Srclib.Types (Locator (Locator))
+import Test.Hspec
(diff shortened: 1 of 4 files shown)


test/App/Fossa/API/BuildLinkSpec.hs
```diff
@@ -2,27 +2,34 @@
...
 
 fullSamlURL :: Text
-fullSamlURL = "https://app.fossa.com/account/saml/33?next=%2Fprojects%2Fa%252Bb%2Frefs%2Fbranch%2Fmaster%2Fc"
+fullSamlURL = "https://app.fossa.com/account/saml/33?next=/projects/a%2bb/refs/branch/master/c"
 
 simpleStandardURL :: Text
-simpleStandardURL = "https://app.fossa.com/projects/haskell%2B89%2Fspectrometer/refs/branch/master/revision123"
+simpleStandardURL = "https://app.fossa.com/projects/haskell+89%2fspectrometer/refs/branch/master/revision123"
+
+stripDiag :: DiagnosticsC (IgnoreLoggerC Maybe) a -> Maybe a
```
cnr, line 31: I think we can build a better test helper that'd `expectationFailure` when diagnostics fail, which would give us better test error messages. similar to [how hspec-megaparsec works](https://hackage.haskell.org/package/hspec-megaparsec-2.2.0/docs/Test-Hspec-Megaparsec.html#v:parseSatisfies)

we could probably widely use it in other tests too. I think we do a lot of tedious diagnostics pattern matching elsewhere in our tests

cnr (approved):
This is great. We should also figure out what a simple `optics`-based solution would look like here

test/App/Fossa/API/BuildLinkSpec.hs
scruffystuffs: What would the type of that helper be?

cnr: something like
```hs
idk :: DiagnosticsC IO a -> (a -> Expectation) -> Expectation

-- used like
test = do
  [...]
  idk (getBuildURLWithOrg org revision apiOpts locator) $ \url -> url `shouldBe` fullSamlURL
```

or maybe
```hs
idk :: Either FailureBundle (ResultBundle a) -> (a -> Expectation) -> Expectation

-- used like
test = do
  [...]
  let res = run . runDiagnostics $ getBuildURLWithOrg org revision apiOpts locator
  res `idk` \url -> url `shouldBe` fullSamlURL

PR #194, opened 2021-02-26
Support for release-group flag

zlav:
This pr adds support for the `release-group` and `release-group-version` flags to support adding projects to release groups from the CLI.

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -15,7 +15,7 @@
 4. [Common FOSSA Project Flags](#common-fossa-project-flags)
 5. [Frequently-Asked Questions](#frequently-asked-questions)
     - [`fossa analyze`: Why wasn't my project found?](#fossa-analyze-why-wasnt-my-project-found)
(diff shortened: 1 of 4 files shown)


scruffystuffs:
Likely needs to be rebased on current master

scruffystuffs:
Also, we need to make sure that the vps route in core supports these as well.  Container scanning goes through the same logic as the normal analyze function, so we will probably get that for free.

Should we block merging this until core is updated?

zlav:
@scruffystuffs I'm fine blocking the merge until core is updated, but I'm not familiar with the VPS codebase. Are you sure that VPS supports release groups?

scruffystuffs:
Release groups is likely only relevant at the project level.  So my guess is that they do, since there's no reason VPS should break this logic.  The relevant core code looks like it's in [this route.](https://github.com/fossas/FOSSA/blob/2cba9a96485fd9f01d4b81b684f97e5ad14a2f69/routes/vendoredPackageScan/index.ts#L179)

PR #195, opened 2021-02-26
Enable vps analyze on windows

kitified:
Enable running `fossa vps analyze` on Windows platforms so long as `--skip-ipr-scan` is used.
Still prevent the other VPS subcommands from running on Windows since they were originally marked as such.

**Still pending testing.** I need to finish making the VPS plugin work in Windows before I can test this, but wanted to go ahead and put it in for review for any initial comments or guidance on approach.


cnr (approved):
This approach is good.

In addition, with `fossa vps analyze` available, we should probably also remove the `dieOnWindows` from `VPSTestCommand` and `VPSReportCommand`

PR #196, opened 2021-03-01
Bundle wiggins

kitified:
Bundle specific Wiggins releases instead of downloading them from Github.

This allows us to easily pin a specific version of Wiggins when doing Spectrometer releases, and additionally adds support for bundling the Windows build.


cnr:
This is going to pretty drastically increase the size of the git repo, especially as additional releases of wiggins get pinned. Is there a reason we can't continue to download the binaries in CI?

I see in one of the earlier commits that `vendor_download.cmd` is an option you explored, but we should be able to use `vendor_download.sh` on windows too -- the CI is running under mingw bash

src/App/Fossa/EmbeddedBinary.hs
```diff
@@ -109,7 +110,7 @@ writeExecutable path content = do
 extractedPath :: PackagedBinary -> Path Rel File
 extractedPath bin = case bin of
   Syft -> $(mkRelFile "syft")
-  Wiggins -> $(mkRelFile "wiggins")
+  Wiggins -> $(mkRelFile $ if SysInfo.os == "mingw32" then "wiggins.exe" else "wiggins")
```
cnr, line 113: I don't think this should be necessary; is there an issue you ran into without the `.exe` extension?

kitified: I didn't, this may have been premature.

cnr: Okay cool. Hoping we can avoid it; trying to keep OS-specific code out of things as much as possible 🙂

scruffystuffs: @cnr what makes this unnecessary?

cnr: I'd be shocked if `process` requires a `.exe` suffix, especially when we use an absolute path to the executable in the `exec` invocation

kitified:
@cnr Updated to use the script. Will ensure I test before merging, but if you could look over this I'd appreciate- trying to get this out today for Commvault if you agree with the changes.

cnr (approved):
lgtm whenever it's tested

kitified:
Yay! Works on both Windows...
Screen Shot 2021-03-01 at 5 58 59 PM

And macOS.
Screen Shot 2021-03-01 at 5 58 43 PM

Thanks @cnr!

PR #201, opened 2021-03-02
Skip target folders in maven; Skip subdirectories of failed scala analysis directories

cnr:
Two issues:

1. After running sbt analysis, `sbt` generates pom.xml files that are picked up by the maven analyzer on the next cli run;
2. Scala projects, when they fail, pretty strongly indicate that analysis in subdirectories is going to fail too. `sbt` is an expensive tool to run, so in failing multi-module projects, this can drastically increase the scan time

### DIFF
--- a/src/Strategy/Maven/Pom/Closure.hs
+++ b/src/Strategy/Maven/Pom/Closure.hs
@@ -37,7 +37,7 @@ findPomFiles dir = execState @[Path Abs File] [] $
     let poms = filter (\file -> "pom.xml" `isSuffixOf` fileName file || ".pom" `isSuffixOf` fileName file) files
     traverse_ (modify . (:)) poms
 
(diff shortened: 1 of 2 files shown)


scruffystuffs (approved):
Nice, I was wondering how we would handle this.  Glad to see it's this clean and easy.

liftM:
Can we get a test for this? Just in general, I worry that if we don't add automated integration tests for these small changes that we'll forget why we made them and regress accidentally.

Also, does SBT always `makePom` into `target` and does Maven never have real POMs in `target`?

cnr:
`target` is the standard "build products" directory for the jvm ecosystem, so we should've been skipping that directory already.

I suspect we were erroneously picking up other duplicate projects with "build product" poms that are generated and used to, e.g., upload to a registry

PR #202, opened 2021-03-04
Update subcommand for AOSP Notice Files

kitified:
* Updates the AOSP Notice Files subcommand to provide the latest options to the VPS plugin.
* Updates the VPS plugin.
* Removes support for the `ninja-graph` subcommand. Right now we just `die` if the command is invoked, but I plan to remove the actual code for this as well in a separate PR.


kitified:
@cnr, @scruffystuffs I tested this on our AOSP instance and it worked great! It's ready for review when you are. Thanks!

cnr (approved):
lgtm. Love the detail of adding a newtype for the input files

PR #206, opened 2021-03-10
Fix scala infinite-looping analysis

cnr:
### Background

Fixes https://github.com/fossas/issues/issues/421

Our scala project analysis executes `sbt makePom` to generate pom files, then uses our existing pom machinery to perform dependency analysis.

`sbt makePom` has a drawback: it doesn't generate the proper `<parent>` references to allow us to bundle poms into multi-module projects.

While we were able to infer what those relationships should have been, `sbt makePom` has another issue: the pom files it generates can contain dependency cycles when the poms are combined into multi-module projects.

---

### Changes

This PR introduces a couple of changes:

1. We no longer try to infer multi-project structure and missing parent relationships. The primary user-facing consequence of this is that some of the "deeper" dependencies in scala subprojects appear closer to the root of the dependency graph than they would have prior to this chang. The discovered dependencies and relationships between them are otherwise unchanged.
2. We prefer static analysis of generated `pom.xml`s, rather than the previous default of using our maven-plugin-based dynamic analysis. The plugin has diminishing returns when run on projects generated by `sbt makePom`, and the runtime is substantially longer

---

### Testing

These changes were integration tested against the scala/sbt projects in `spectrometer-tests`


scruffystuffs (approved):
Looks good to me.  If integration tests are passing, I think we're in good shape.

I also really like that we're preferring static analysis in one more case.  Especially on a maven-based buildtool.

PR #207, opened 2021-03-11
fix(composer): Don't crash when packages in lockfiles are missing source field

liftM:
## Overview

Fixes fossas/issues#417.

`composer.lock` packages only need to have _either_ `source` or `dist`, not both (note that metapackages may have neither!). See documentation [here](https://getcomposer.org/doc/05-repositories.md#package-2). But we never use that field anyway, so I just deleted the parsing.

## Testing plan

Run `fossa analyze` on a test project constructed from the `composer.lock` in fossas/issues#417. See that the analysis does not crash, and uploads correct dependencies.


cnr (approved):
Hopefully one day we'll use the `source` field.. for now lgtm

PR #208, opened 2021-03-11
fix(ci): Fix `cabal install` and CI breakages

liftM:
## Overview

This PR fixes our CI tests.

The dependency scan was failing because our build is broken with GHC 9, which is what the `ubuntu-latest` container was running. Obviously, this prevented a scan from working because we don't support static analysis with Haskell.

Other changes:
- Fully specify `extra-source-files`, to partially fix `cabal install`. Originally, the glob wasn't working: `cabal install` would report that they were invalid. This doesn't really make a difference, since `cabal build` ignores this field, and `cabal install` is still broken because we're not copying `.git` (see comment below).
- Ran `cabal-fmt` over `spectrometer.cabal`.
- Ran `ormolu` over some files I touched.
- Fully specified imports in files I touched.

## Testing plan

1. Check that the CI now passes.
2. Compile with `cabal build` and check that `cabal run fossa -- --version` still produces expected output.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,8 +65,9 @@ jobs:
 
     - name: Debugging information
       run: |
-        ghc --version
-        cabal --version
+        ghc --version || echo "no ghc"
+        cabal --version || echo "no cabal"
+        ghcup --version || echo "no ghcup"
 
     - uses: actions/cache@v2
       name: Cache cabal store
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,6 +63,11 @@ jobs:
       with:
         ghc-version: ${{ matrix.ghc }}
 
+    - name: Debugging information
+      run: |
+        ghc --version
+        cabal --version
+
     - uses: actions/cache@v2
       name: Cache cabal store
       with:
@@ -72,7 +77,7 @@ jobs:
           ${{ runner.os }}-${{ matrix.ghc }}-cabal-cache-
           ${{ runner.os }}-${{ matrix.ghc }}-
           ${{ runner.os }}-
-  
+
     - name: Update vendored binaries
       run: |
         mkdir vendor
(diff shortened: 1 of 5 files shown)


liftM:
Actually, hold on - let me see if I can fix the `cabal install` problem too.

liftM:
Okay, I tried very hard to fix `cabal install`, and came to the conclusion that it wasn't worth it. For posterity, here's what's going on:

When you run `cabal install`, Cabal will copy source files into a directory in `/tmp` to run a build. "Source files" means Haskell files specified `.cabal` file, plus anything in `extra-source-files`. Notice that `.git` is not in `extra-source-files`.

In `App.Version` and `App.Version.TH`, we attempt to shell out to `git` in order to get the current revision. This fails because there is no local `.git` folder in the build folder (the one in `/tmp`), so `git` exits 1.

The right solution to this is to add the needed files (in `.git`) to `extra-source-files`, so they get copied to the build folder. Unfortunately, this is _really annoying_ due to design decisions by Cabal:

- You can't specify a folder name in `extra-source-files`.
- You can't glob arbitrary files without file extensions (e.g. `.git/**/*`).
- You have to basically list every single file in `.git` individually. Also, you have to do this every build because the files in `.git` will change (and, importantly, tags are represented as new individual files!).
- `cabal-fmt expand` and `cabal-fmt glob-files` don't work on `extra-source-files`.

So, I give up. To get `cabal install` to work, we clearly need to programmatically generate our `.cabal` file at build-time, or alternatively patch Cabal upstream to be less annoying. Both are way more effort than it's worth, since nobody should be using `cabal install` anyway. (Maybe this is a use case for `Setup.hs`? Ugh.)

cnr:
> `cabal-fmt glob-files` doesn't work on `extra-source-files`
that's.. a frustrating limitation

cnr (approved):

PR #210, opened 2021-03-15
Filtered container scan

scruffystuffs:
Fixes fossas/issues#453

## Description
When scanning container images, `syft` finds many packages in a large variety of formats.  For the purposes of container scanning, at least in it's current mode, we only care about "system-level" packages (don't read into the name too much, it's not official at all).  Specifically, we care about debian, redhat, and alpine packages, which are given the type identifiers `deb`, `rpm`, and `apk`.  The CLI should not report these "artifacts" (syft terminology) as part of the build.

## This PR
This PR performs three tasks:
* (For debugging) Add a hidden command `fossa container parse-file FILE` which checks that a given syft output file will parse correctly, and print the parsed and transformed result.  
* (For debugging) Add a hidden command `fossa container dump-scan IMAGE` which will print the raw output from the syft scan to stdout (or the file provided by `-o/--output-file FILE`).
* Filter the artifacts at parse-time, only accepting those whose types are one of the accepted artifact types.

### DIFF
--- a/src/App/Fossa/Container.hs
+++ b/src/App/Fossa/Container.hs
@@ -17,13 +17,15 @@ module App.Fossa.Container
     runSyft,
     toContainerScan,
     extractRevision,
+    parseSyftOutput,
+    parseSyftOutputMain,
   )
 where
 
 import App.Fossa.EmbeddedBinary (BinaryPaths, toExecutablePath, withSyftBinary)
 import App.Types (ProjectRevision (..), OverrideProject (..))
-import Control.Effect.Diagnostics hiding (fromMaybe)
-import Control.Effect.Lift (Lift)
+import Control.Carrier.Diagnostics hiding (fromMaybe)
+import Control.Effect.Lift (Lift, sendIO)
 import Control.Monad.IO.Class
 import Data.Aeson
 import qualified Data.Map.Lazy as LMap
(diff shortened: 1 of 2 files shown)


src/App/Fossa/Container.hs
```diff
@@ -206,3 +220,41 @@ syftCommand bin (ImageText image) =
...
+  path <- resolveFile curdir $ pack filepath
+  logDebug "Reading JSON file"
+  rawvalue <- readContentsJson @Value path
+  logDebug "Parsing JSON contents"
+  response <- fromEitherShow $ parseEither (parseJSON @SyftResponse) rawvalue
+  logDebug "Converting to FOSSA payload"
+  payload <- toContainerScan response
+  logInfo "Payload is valid!"
+
+  sendIO . BL.putStr $ encode payload
```
cnr, line 240: It doesn't matter much here, but we do have `logStdout` for this

cnr (approved):

src/App/Fossa/Container.hs
scruffystuffs: I got stuck trying to do this and forgot to fix it.  Lazy `Bytestring` doesn't have a `Pretty` instance, and I couldn't find `Data.Text.Lazy.Encoding (decodeUtf8)` until just now.

Good callout, btw.  We should actually be much stricter about this because the write ordering is non-deterministic when using the logger mixed with raw IO.  Because the logging effects are sequenced, using the logger is deterministic.

scruffystuffs: Given that these are printed to different handles, it doesn't make a difference programmatically, but in a shell with no redirection, there's no guarantee they won't clobber each other.

PR #211, opened 2021-03-17
Scala: use sbt.ci flag instead of closing stdin

cnr:
In a previous PR, we decided to close stdin because some tools, like
sbt, start a repl and hang forever when a command fails. Closing stdin
prevents the tools from doing so.

Unfortunately this has a downside: other buildtools, like gradle, will
sometimes hang forever when stdin is closed.

This commit instead uses the `sbt.ci` flag, which disables its repl by
default, and restores the previous behavior of "inheriting" stdin from
the parent process

---

*Testing*

These changes were tested against:
- a local sbt project with a failing build, which would previously cause analysis to hang forever under the default "inherit stdin" behavior. I'm going to open a companion PR to spectrometer-tests that includes the failing project and a test for this behavior;
- the spectrometer-tests projects

### DIFF
--- a/src/Effect/Exec.hs
+++ b/src/Effect/Exec.hs
@@ -189,7 +189,7 @@ instance Has (Lift IO) sig m => Algebra (Exec :+: sig) (ExecIOC m) where
           ioExceptionToCmdFailure :: IOException -> CmdFailure
           ioExceptionToCmdFailure = mkFailure (ExitFailure 1) "" . fromString . show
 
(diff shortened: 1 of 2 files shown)


scruffystuffs (changes requested):
Code is fine, but we probably need to rebase on master to fix the build issue in the dependency-scan job.

scruffystuffs (approved):
I can't find any documentation for the --batch flag, can you leave it in a comment before merging?  Otherwise LGTM

liftM:
Resolves an issue for https://github.com/fossas/issues/issues/423.

PR #212, opened 2021-03-19
Add (very) basic replay logging documentation

cnr:
In addition to adding very basic documentation for replay logging, this PR improves the ergonomics of using replay logging by adding new (hidden) `--record` and `--replay <file>` flags to `fossa analyze`.

This also conceptually separates `--debug` mode from `--record` mode, whereas the former used to imply the latter.

### DIFF
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -7,14 +7,17 @@ module App.Fossa.Analyze
   , ScanDestination(..)
   , UnpackArchives(..)
   , discoverFuncs
+  , RecordMode(..)
   ) where
 
 import App.Fossa.API.BuildLink (getFossaBuildUrl)
 import App.Fossa.Analyze.GraphMangler (graphingToGraph)
 import App.Fossa.Analyze.Project (ProjectResult (..), mkResult)
+import App.Fossa.Analyze.Record (AnalyzeEffects (..), AnalyzeJournal (..), loadReplayLog, saveReplayLog)
 import App.Fossa.FossaAPIV1 (UploadResponse (..), uploadAnalysis, uploadContributors)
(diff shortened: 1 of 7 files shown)


src/App/Fossa/Analyze/Record.hs
```diff
@@ -0,0 +1,81 @@
...
+
+instance ToJSON AnalyzeEffects where
+  toJSON AnalyzeEffects {..} =
+    object
+      [ "ReadFS" .= effectsReadFS,
+        "Exec" .= effectsExec
+      ]
+
+saveReplayLog :: Journal -> Journal -> FilePath -> IO ()
+saveReplayLog readFSJournal execJournal path = do
```
scruffystuffs, line 61: [question]: Is it worth making these more type-safe so we don't mix them up?  Either newtypes or something that looks like `Journal eff` where `eff` is some first-order effect?

src/App/Fossa/Analyze.hs
```diff
@@ -86,20 +91,38 @@ data ScanDestination
...
+analyzeMain workdir recordMode logSeverity destination project unpackArchives filters =
   withLogger logSeverity
     . Diag.logWithExit_
     . runReadFSIO
     . runExecIO
-    $ if not debugMode
-      then analyze basedir destination project unpackArchives filters
-      else do
+    $ case recordMode of
+      RecordModeNone -> do
```
scruffystuffs, line 107: Really glad to see this not tied to debug mode.  I was going to file this as a separate issue until I saw it here.

scruffystuffs (approved):
Looks good.  One question about type-safety, but not necessary for this PR.

src/App/Fossa/Analyze/Record.hs
cnr: Oh yeah, good idea. a phantom type on `Journal` would work great here

PR #217, opened 2021-03-26
AOSP notice files: accept paths as arguments

kitified:
Updates `fossa vps aosp-notice-file` to require a `--ninja-files` argument, which is a comma separated list of Ninja file paths to use for reconstructing the build graph.

This replaces the existing functionality of "identify and use _all_ ninja files in the scan directory", which was not always valid.

### DIFF
--- a/src/App/Fossa/Main.hs
+++ b/src/App/Fossa/Main.hs
@@ -20,7 +20,7 @@ import qualified App.Fossa.VPS.Report as VPSReport
 import App.Fossa.VPS.Scan (LicenseOnlyScan (..), SkipIPRScan (..), scanMain)
 import App.Fossa.VPS.AOSPNotice (aospNoticeMain)
 import qualified App.Fossa.VPS.Test as VPSTest
-import App.Fossa.VPS.Types (FilterExpressions (..), NinjaScanID (..),NinjaFilePaths (NinjaFilePaths))
+import App.Fossa.VPS.Types (FilterExpressions (..), NinjaScanID (..), NinjaFilePaths (..))
 import App.OptionExtensions
 import App.Types
 import App.Util (validateDir, validateFile)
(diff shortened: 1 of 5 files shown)


src/App/Fossa/VPS/Scan/RunWiggins.hs
```diff
@@ -41,19 +42,19 @@ generateWigginsScanOpts :: Path Abs Dir -> Severity -> ProjectRevision -> ScanTy
...
+generateSpectrometerAOSPNoticeArgs :: Severity -> ApiOpts -> ProjectRevision -> NinjaScanID -> NinjaFilePaths -> [Text]
 generateSpectrometerAOSPNoticeArgs logSeverity ApiOpts{..} ProjectRevision{..} ninjaScanId ninjaInputFiles =
   ["aosp-notice-files"]
       ++ optBool "-debug" (logSeverity == SevDebug)
       ++ ["-endpoint", render apiOptsUri, "-fossa-api-key", unApiKey apiOptsApiKey]
       ++ ["-scan-id", unNinjaScanID ninjaScanId]
       ++ ["-name", projectName]
       ++ ["."]
-      ++ unNinjaInputFiles ninjaInputFiles
+      ++ (T.pack . show <$> unNinjaFilePaths ninjaInputFiles)
```
scruffystuffs, line 57: The function we normally use is `toFilePath`.  While the `path` docs do guarantee that `show` is basically the same, it's a little unintuitive, since `show` often returns a representation of the value rather than a rendered value.  I had to look at the docs to confirm it was the same, since I doubted it at first.
```suggestion
      ++ (T.pack . toFilePath <$> unNinjaFilePaths ninjaInputFiles)
```

src/App/Fossa/Main.hs
```diff
@@ -149,6 +151,10 @@ appMain = do
 dieOnWindows :: String -> IO ()
 dieOnWindows op = when (SysInfo.os == windowsOsName) $ die $ "Operation is not supported on Windows: " <> op
 
+
+parseCommaSeparatedFileArg :: Text -> IO [Path Abs File]
+parseCommaSeparatedFileArg arg = sequence $ validateFile . T.unpack <$> T.splitOn arg ","
```
scruffystuffs, line 156: [nit]: The mix of operators is confusing (to me at least).  I'm sure it's doing what it looks like, from right to left, but it's a little hard when `$`, `.`, and `<$>` are all mixed in a one-liner.  I'd recommend parens for clarity:
```suggestion
parseCommaSeparatedFileArg arg = sequence (validateFile . T.unpack <$> T.splitOn arg ",")
```

scruffystuffs (approved):
Two small comments, but overall this looks good.

PR #221, opened 2021-04-07
Glide.lock: Parse versions as Text

cnr:
We were previously parsing them as integers, which passed the tests from cliv1.

Updated the test `glide.lock` to add an alphanumeric hexadecimal version string for one of the dependencies

### DIFF
--- a/src/Strategy/Go/GlideLock.hs
+++ b/src/Strategy/Go/GlideLock.hs
@@ -11,6 +11,7 @@ module Strategy.Go.GlideLock
   )
   where
 
+import Control.Applicative ((<|>))
 import Control.Effect.Diagnostics
 import Data.Aeson
 import qualified Data.Map.Strict as M
@@ -32,7 +33,7 @@ buildGraph lockfile = Graphing.fromList (map toDependency direct)
   toDependency GlideDep{..} =
     Dependency { dependencyType = GoType
                , dependencyName = depName
-               , dependencyVersion = Just (CEq $ T.pack (show depVersion))
+               , dependencyVersion = Just (CEq depVersion)
                , dependencyLocations = []
                , dependencyEnvironments = []
                , dependencyTags = M.empty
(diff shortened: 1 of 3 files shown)


src/Strategy/Go/GlideLock.hs
```diff
@@ -59,5 +60,9 @@ instance FromJSON GlideLockfile where
 instance FromJSON GlideDep where
   parseJSON = withObject "GlideDep" $ \obj ->
     GlideDep <$> obj .:  "name"
-               <*> obj .: "version"
+               -- version field can be text or an int (hexadecimal hash)
+               <*> (obj .: "version" <|> (intToText <$> obj .: "version"))
                <*> obj .:? "repo"
+
+intToText :: Int -> Text
+intToText = T.pack . show
```
scruffystuffs, line 68: `Data.Text.Extra` has `showT :: Show a => a -> Text`

src/Strategy/Go/GlideLock.hs
```diff
@@ -59,5 +60,9 @@ instance FromJSON GlideLockfile where
 instance FromJSON GlideDep where
   parseJSON = withObject "GlideDep" $ \obj ->
     GlideDep <$> obj .:  "name"
-               <*> obj .: "version"
+               -- version field can be text or an int (hexadecimal hash)
+               <*> (obj .: "version" <|> (intToText <$> obj .: "version"))
```
scruffystuffs, line 64: Why are we not just parsing `Text`?  If we're validating input, we can just validate that text is digits or a hex hash, right?

cnr: This is to handle the overwhelmingly-rare edge case where the hexadecimal number is a valid base-10 int (e.g., `1234` from the test file). In that case, it becomes a yaml `int`, so trying to parse it as `Text` would fail

scruffystuffs: Another damn reason to hate yaml

cnr: We're just not  p r a g m a t i c  enough. golang's yaml parser will happily coerce numbers to strings! (everything is a string if you squint hard enough)

scruffystuffs (approved):
Looks good.

PR #222, opened 2021-04-08
chore: Normalize syntax, normalize extensions, and lock dependency install plan

liftM:
Changes:

- Lock dependency resolution to be reproducible by specifying the specific Hackage index state.
- Update our default language extensions to align with the Haskell 2021 proposal (see https://github.com/ghc-proposals/ghc-proposals/blob/ghc2021/proposals/0000-ghc2021.rst).
- Delete unused language pragmas, as detected by `hlint`.
- Update the version of GHC to 8.10.4 (see https://github.com/fossas/haskell-static-alpine).
- Remove redundant pragmas (specified in `default-extensions`).
- Remove leading and trailing module whitespace.
- Remove trailing line whitespace.


liftM:
Ugh, hold on, it looks like GHC 8.8 is breaking the Linux build.

liftM:
Okay, I fixed the build issue for Linux by updating our build container image. I think the problem was an outdated version of `cabal` that used a different syntax for the `index-state` field of `cabal-project`.

I've also removed now-redundant pragmas. Here's the script I used:

```js
const fs = require('fs');
const fsw = require('@nodelib/fs.walk');

const extensions = [
  'BangPatterns',
  'BinaryLiterals',
  'ConstrainedClassMethods',
  'ConstraintKinds',
  'DeriveDataTypeable',
  'DeriveFoldable',
  'DeriveFunctor',
  'DeriveGeneric',
  'DeriveLift',
  'DeriveTraversable',
  'EmptyCase',
  'EmptyDataDeriving',
  'ExistentialQuantification',
  'ExplicitForAll',
  'FlexibleContexts',
  'FlexibleInstances',
  'GADTSyntax',
  'GeneralisedNewtypeDeriving',
  'HexFloatLiterals',
  'ImportQualifiedPost',
  'InstanceSigs',
  'KindSignatures',
  'MultiParamTypeClasses',
  'NamedFieldPuns',
  'NamedWildCards',
  'NumericUnderscores',
  'OverloadedStrings',
  'PolyKinds',
  'PostfixOperators',
  'RankNTypes',
  'ScopedTypeVariables',
  'StandaloneDeriving',
  'StandaloneKindSignatures',
  'StrictData',
  'TupleSections',
  'TypeApplications',
  'TypeOperators',
  'TypeSynonymInstances',
];

const pragmas = extensions.map(extension => `{-# LANGUAGE ${extension} #-}`);

const files = fsw.walkSync('.')
  .map(entry => entry.path)
  .filter(path => path.endsWith('.hs'))
  .filter(path => !path.startsWith('./dist-newstyle'));

files.forEach(file => {
  const contents = fs.readFileSync(file);
  const lines = contents.toString().split('\n');
  const newContents = lines
    .map(line => line.trimEnd())
    .filter(line => pragmas.indexOf(line) == -1)
    .join('\n')
    .trimStart()
    .trimEnd() + '\n';
  fs.writeFileSync(file, newContents);
});
```

There were also some pragmas that `hlint` detected as unused - I removed those by hand.

scruffystuffs (approved):
Non-blocking question: how do we set, update, and sync the index-state when it needs to be changed?

Proposal: update the index state after every release.  This prevents us from stagnating our dependencies.  We could have a bot make a PR as a hook after the release is created, and attempt to auto-merge when the release is published.

liftM:
> update the index state after every release

This is a really good idea - I'm happy to put something together for this. It might just be a checklist at first.

PR #226, opened 2021-04-14
Conda support

anuccio1:
This PR adds support for [conda](https://docs.conda.io/en/latest/). We will support it with 2 strategies:
- `conda list`: this will output the dependencies to stdout
- `environment.yml`: if `conda list` fails, then we fall back to parsing a yml file


src/Strategy/Conda.hs
```diff
@@ -0,0 +1,57 @@
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Strategy.Conda (
```
scruffystuffs, line 4: Unused extensions
```suggestion
module Strategy.Conda (
```

src/Strategy/Conda.hs
```diff
@@ -0,0 +1,57 @@
...
+  DiscoveredProject
+    { projectType = "conda",
+      projectBuildTargets = mempty,
+      projectDependencyGraph = const $ getDeps project, -- what does const do here?
+      projectPath = condaDir project,
+      projectLicenses = pure []
+    }
+
+getDeps :: (Has Exec sig m, Has ReadFS sig m, Has Diagnostics sig m) => CondaProject -> m (Graphing Dependency)
+getDeps project = analyzeCondaList project <||> analyzeEnvironmentYml project
```
scruffystuffs, line 51: We should use `combineSuccessful` here, so that we try both concurrently (or at least concurrent-able) and merge them if they are both successful.
```suggestion
getDeps project = combineSuccessful "failed to analyze conda project" [analyzeCondaList project,  analyzeEnvironmentYml project]
```

src/Strategy/Conda.hs
```diff
@@ -0,0 +1,57 @@
...
+              }
+      pure([project], WalkSkipAll) -- Once we find an environment.yml file, skip the rest of the walk
+
+data CondaProject =
+    CondaProject
+      { condaDir :: Path Abs Dir,
+        condaEnvironmentYml :: Path Abs File
+      } deriving (Eq, Ord, Show)
+
+mkProject :: (Has ReadFS sig n, Has Exec sig n, Has Diagnostics sig n) => CondaProject -> DiscoveredProject n
```
scruffystuffs, line 40: I know it's silly, but this is weird to me that these are `n` instead of `m`.  `m` actually stands for `Monad`, to signal that we are specifically using monads to make everything line up together.  Although the `m` representing "monad" is entirely coincidence, it is intentional coincidence.
```suggestion
mkProject :: (Has ReadFS sig m, Has Exec sig m, Has Diagnostics sig m) => CondaProject -> DiscoveredProject m
```

src/Strategy/Conda.hs
```diff
@@ -0,0 +1,57 @@
...
+mkProject project =
+  DiscoveredProject
+    { projectType = "conda",
+      projectBuildTargets = mempty,
+      projectDependencyGraph = const $ getDeps project, -- what does const do here?
```
scruffystuffs, line 45: TLDR: `const` here represents that we don't care about or cannot filter on subprojects.

### Long version
For tools like `gradle`, subprojects can be hugely important, and scanning them all at once is simply not a one-size-fits-all solution.  We have to use `BuildTarget`s to signal that we only care about certain subprojects, and if a strategy can honor that, it will do that by using a function other than `const`.  If we cared about subprojects we could define `getSpecificDeps :: FooProject -> Set BuildTarget -> Graphing Dependency`, and then use `getSpecificDeps project` instead of `const $ getDeps project`

`const :: a -> b -> a`, when applied with only one argument, creates a function that ignores it's argument and returns the first applied argument.

```haskell
const :: a -> b -> a
const a _ = a

const 3 4 = 3
const 4 3 = 4
const "hello" (Just "darkness") = "hello"

**Review by scruffystuffs — COMMENTED** [review]
Left a lot of comments, mostly small stuff like simplifications.

The two critical changes are `combineSuccessful` and not using `head`.

I also have a sneaking suspicion that you have not set up the linter and formatter.  We use [hlint](https://github.com/ndmitchell/hlint/releases/latest) for linting and [ormolu](https://github.com/tweag/ormolu/releases/latest) for formatting.  haskell-language-server integrates with both if they're on the PATH, and configured correctly (vscode is the reference for configuration, I think).

File: src/Strategy/Conda.hs (hunk 4 above)
Comment (anuccio1) [reply src/Strategy/Conda.hs:45]: that makes sense, good to know

File: src/Strategy/Conda.hs (hunk 2 above)
Comment (anuccio1) [reply src/Strategy/Conda.hs:51]: Wouldn't that combine the results from both?

```
-- | Run a list of actions, combining the successful ones. If all actions fail, 'fatalText' is invoked with the provided @Text@ message.
```

Comment (scruffystuffs) [reply src/Strategy/Conda.hs:51]: Is there a reason we shouldn't combine them?

Comment (scruffystuffs) [reply src/Strategy/Conda.hs:48]: That doesn't really explain why we wouldn't combine them.  Any dependency found by both will be merge as part of a dedupe operation, so combining should be safe. The more expensive one will always be tried first, so there's no real performance concern.

Unless there's reason to expect that running both will produce incorrect results when combined, we should run both.

Comment (anuccio1) [reply src/Strategy/Conda.hs:48]: Example:

There might be a dep with a version spec in an `environment.yml` file: i.e. `conda+foo$`1.2.*`, and perhaps the same dep resolved to a known version in the users virtual environment: i.e. `conda+foo$1.2.4` (we get that form `conda list`. If we combined the results then we would include both of those deps in the result, which is not correct behavior. We want to take `conda list` results if there, and if we can't get them for whatever reason, fallback and take the results from parsing the `environment.yml` file

File: src/Strategy/Conda/CondaList.hs [hunk 5]
```diff
@@ -0,0 +1,62 @@
...
+      {
+        dependencyType = CondaType,
+        dependencyName = listName,
+        dependencyVersion = CEq <$> listVersion,
+        dependencyLocations = [],
+        dependencyEnvironments = [],
+        dependencyTags = M.empty
+      }
+
+analyze' ::
```
scruffystuffs, line 42: This should just be `analyze`.  We only ever use `'` to denote multiple versions of the same function under different constraints or semantics.  Since there's no `analyze`, there should be no `analyze'`.

Btw, we may have this in other strategies, please point them out if you see them, those are (in my opinion) incorrect, and leftover from a refactor.

src/Strategy/Conda/EnvironmentYml.hs
```diff
@@ -0,0 +1,94 @@
...
+       {
+        dependencyType = CondaType,
+        dependencyName = depName,
+        dependencyVersion = CEq <$> depVersion, -- todo - properly handle version constraints
+        dependencyLocations = [],
+        dependencyEnvironments = [],
+        dependencyTags = M.empty
+      }
+
+analyze' ::
```
scruffystuffs, line 40: same comment as `Strategy.Conda.CondaList.analyze'`

src/Strategy/Conda/EnvironmentYml.hs
```diff
@@ -0,0 +1,94 @@
...
+  where
+    depSplit = T.split (== '=') rcd
+
+    name = fromMaybe rcd (depSplit !? 0)
+    version = depSplit !? 1 -- TODO: this may contain constraints that we need to parse
+    build = depSplit !? 2
+    fullVersion = getFullVersion version build
+
+    -- if we have a version AND a build, we combine them to form a full version
+    -- i.e. Just "1.2.3" Just "build" => Just "1.2.3=build"
```
scruffystuffs, line 70: [nit]: More haddock styling.  We don't do much of this, but I think that we should where it's easy.
[reference for doc examples](https://haskell-haddock.readthedocs.io/en/latest/markup.html#examples)
```suggestion
    >>> getFullVersion (Just "1.2.3") (Just "build")
    Just "1.2.3=build"
```

test/Conda/EnvironmentYmlSpec.hs
```diff
@@ -0,0 +1,84 @@
+module Conda.EnvironmentYmlSpec
+  ( spec
+  )
+where
+
+import Prelude
```
scruffystuffs, line 6: The prelude should be imported automatically.  Can you try removing this?

src/Data/List/Extra.hs
```diff
@@ -0,0 +1,9 @@
+module Data.List.Extra ( (!?) ) where
+
+-- infix operator to access an array (similar to !!)
```
scruffystuffs, line 3: [nit]: Haddock styling
```suggestion
-- | infix operator to access an array (similar to 'Data.List.!!')
```

src/Srclib/Converter.hs
```diff
@@ -108,3 +108,5 @@ depTypeToFetcher = \case
   URLType -> "url"
   ComposerType -> "comp"
   HackageType -> "hackage"
+  CondaType -> "conda"
+
```
scruffystuffs, line 112: [nit]: Looks like trailing whitespace with no newline.

src/Strategy/Conda.hs
```diff
@@ -0,0 +1,58 @@
+module Strategy.Conda (
```
scruffystuffs, line 1: Please run ormolu on this file.

src/Data/List/Extra.hs
```diff
@@ -0,0 +1,9 @@
+module Data.List.Extra ( (!?) ) where
```
scruffystuffs, line 1: Please run ormolu on this file.

test/Conda/EnvironmentYmlSpec.hs
```diff
@@ -0,0 +1,84 @@
+module Conda.EnvironmentYmlSpec
```
scruffystuffs, line 1: Please run ormolu on this file.

src/Strategy/Conda/EnvironmentYml.hs
```diff
@@ -0,0 +1,94 @@
...
+ ( Has ReadFS sig m
+ , Has Diagnostics sig m
+ )
+  => Path Abs File -> m (Graphing Dependency)
+analyze' envYml = buildGraph <$> readContentsYaml @EnvironmentYmlFile envYml
+
+
+-- an example Text: "biopython=1.78=py38haf1e3a3_0"
+-- '=' is a delimeter for <name>=<version>=<build>
+-- where <version> and <build> are optional
```
scruffystuffs, line 50: [nit]: We use haddock-style comments where applicable.  [Here is the haddock documentation](https://haskell-haddock.readthedocs.io/en/latest/markup.html)

scruffystuffs (approved):
Overall, looks good to me.

I left a bunch of doc comments as nits, but the formatting and naming of `analyze` functions should probably be taken care of before merging.

src/Strategy/Conda.hs
anuccio1: Is there a more automated way of doing this? I feel like this could be a pre-commit hook or something

scruffystuffs: If you want to set up your own pre-commit hook, you can.  See [the ormolu usage docs](https://github.com/tweag/ormolu#usage) for invocation help.

However, it's worth noting that I do not think we should **enforce** git hooks in any repository.  Anything that needs to be checked should be done in CI.  We can do that using `ormolu --mode check FILES`.

scruffystuffs: There is also an ormolu github action [here](https://github.com/marketplace/actions/ormolu-action)

PR #228, opened 2021-04-15
Change container scanning layer scope

Serubin:
Change syft to capture [all layers scope rather than a squashed scoped](https://github.com/anchore/syft#getting-started) to improve layer location results.

Also de-dups `locations`.  
Previously, we would see duplicate location entries, with the same layerId due to the `path` parameter we filter out.

Out put gets reduced as follows:
**From**
```
      "locations": [
        {
          "path": "/var/lib/dpkg/status",
          "layerID": "sha256:e2c6ff46235709f5178ab7c1939f4fba7237ffde84e13b1582fa5f0837c1d978"
        },
        {
          "path": "/var/lib/dpkg/info/zlib1g:amd64.md5sums",
          "layerID": "sha256:e2c6ff46235709f5178ab7c1939f4fba7237ffde84e13b1582fa5f0837c1d978"
        },
        {
          "path": "/usr/share/doc/zlib1g/copyright",
          "layerID": "sha256:e2c6ff46235709f5178ab7c1939f4fba7237ffde84e13b1582fa5f0837c1d978"
        }
      ],
```

**To**
```
        "locations": [
          {
            "layerId": "sha256:e2c6ff46235709f5178ab7c1939f4fba7237ffde84e13b1582fa5f0837c1d978"
          }
        ],
```

### DIFF
--- a/src/App/Fossa/Container.hs
+++ b/src/App/Fossa/Container.hs
@@ -260,7 +260,7 @@ syftCommand :: BinaryPaths -> ImageText -> Command
 syftCommand bin (ImageText image) =
   Command
     { cmdName = pack . toFilePath $ toExecutablePath bin,
-      cmdArgs = ["-o", "json", image],
+      cmdArgs = ["--scope", "all-layers", "-o", "json", image],
       cmdAllowErr = Never
     }
 
--- a/src/App/Fossa/Container.hs
+++ b/src/App/Fossa/Container.hs
@@ -257,6 +257,8 @@ runSyft ::
 runSyft image = runExecIO . withSyftBinary $ \syftBin -> do
   execJson @SyftResponse [reldir|.|] $ syftCommand syftBin image
 
+-- Scope to all layers, to prevent odd "squashing" that syft does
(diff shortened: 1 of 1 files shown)


src/App/Fossa/Container.hs
```diff
@@ -260,7 +261,7 @@ syftCommand :: BinaryPaths -> ImageText -> Command
 syftCommand bin (ImageText image) =
   Command
     { cmdName = pack . toFilePath $ toExecutablePath bin,
-      cmdArgs = ["-o", "json", image],
+      cmdArgs = ["--scope", "all-layers", "-o", "json", image],
```
scruffystuffs, line 264: Can we post a short comment on these options, and also maybe link to syft docs explaining more if needed?

scruffystuffs (approved):
One bit about comments, but this is fine.

src/App/Fossa/Container.hs
Serubin: Add comments - the docs are not particularly useful so I left them out. if the docs update, I'll add them.

PR #234, opened 2021-04-27
Golang: map package imports to modules

cnr:
`go list -json all` emits golang dependencies the _package_ level; e.g.,
`github.com/example/foo/some/package`. The fossa backend handles
package-level imports poorly, especially when there are many of them
from the same go module. See the ticket for details.

As a workaround, we map package imports to their modules, and use the
modules in the final dependency graph.

Fixes https://github.com/fossas/team-analysis/issues/514

### DIFF
--- a/test/Go/TransitiveSpec.hs
+++ b/test/Go/TransitiveSpec.hs
@@ -1,0 +1,160 @@
+module Go.TransitiveSpec (spec) where
+
+import Control.Algebra (run)
+import DepTypes
+import GraphUtil
+import Strategy.Go.Transitive
+import Strategy.Go.Types (graphingGolang)
+import Test.Hspec
+
+spec :: Spec
+spec = spec_packageToModule
+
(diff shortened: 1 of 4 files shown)


test/Go/TransitiveSpec.hs
```diff
@@ -0,0 +1,160 @@
...
+spec :: Spec
+spec = spec_packageToModule
+
+-- HACK(fossas/team-analysis#514) see Strategy.Go.Transitive for more details
+--
+-- Ensure that packages are correctly mapped to their modules, and that packages
+-- aren't dropped from the dependency graph
+spec_packageToModule :: Spec
+spec_packageToModule =
+  describe "normalizeImportsToModules" $ do
```
liftM, line 19: Can you add a separate test in here that checks that not-module packages are correctly handled? I know this is already covered by `should map package imports to their modules`, but I prefer explicitly encoding this intent.

liftM (approved):
LGTM. Feel free to merge.

PR #235, opened 2021-05-05
Requirements.txt: fix hang when parsing unsupported fields

cnr:
-------

# Overview

When scanning a project, a customer had an issue with the CLI hanging during python analysis: [ZD2450](https://fossa.zendesk.com/agent/tickets/2450)

This fixes the hang by ignoring lines in requirements.txt that we're unable to parse

## Acceptance criteria

- Lines with unexpected content in a requirements.txt file, e.g. a section marker like `[foo]`, should not cause the analysis to hang

## Testing plan

- Added a `[foo]` field to the file we use to test our requirements.txt parser
- Verified the fix by running analysis against requirements.txt files supplied by the customer files supplied by the customer. Zip file available on the ticket.

## Risks

None.

## References

Fixes https://github.com/fossas/team-analysis/issues/521

### DIFF
--- a/src/Strategy/Python/ReqTxt.hs
+++ b/src/Strategy/Python/ReqTxt.hs
@@ -34,22 +34,18 @@ reqParser = [] <$ char '-' <* ignored -- pip options
      <|> [] <$ char '/' <* ignored -- absolute path
      <|> [] <$ oneOfS ["http:", "https:", "git+", "hg+", "svn+", "bzr+"] <* ignored -- URLs
      <|> [] <$ comment
-     <|> (pure <$> requirementParser <* ignored)
-     <|> pure [] -- empty line
+     <|> (pure <$> try requirementParser <* ignored)
+     <|> [] <$ ignored -- something else we're not able to parse
 
   where
   isEndLine :: Char -> Bool
   isEndLine '\n' = True
   isEndLine '\r' = True
-  isEndLine '\\' = True
   isEndLine _    = False
 
   -- ignore content until the end of the line
   ignored :: Parser ()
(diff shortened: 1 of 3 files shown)


Changelog.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
```
zlav, line 1: Can you add this to the PR template?

Right now it says " If this PR marked a release, I updated CHANGELOG.md.", but I think updating the changelog should be mandatory every time. So maybe "I updated CHANGELOG.md." would work.

src/Strategy/Python/ReqTxt.hs
```diff
@@ -34,22 +34,18 @@ reqParser = [] <$ char '-' <* ignored -- pip options
...
   isEndLine '\n' = True
   isEndLine '\r' = True
-  isEndLine '\\' = True
   isEndLine _    = False
 
   -- ignore content until the end of the line
   ignored :: Parser ()
-  ignored = () <$ takeWhileP (Just "ignored") (not . isEndLine) <* (try newLine <|> (() <$ takeWhileP (Just "end of line") isEndLine))
-
-  newLine :: Parser ()
```
zlav, line 51: I'm curious if removing this newLine parsing will ever come back to bite us, although I don't see any difference between the way I implemented it and the way you are handling it. If we ever want to capture the data in the newlines, such as `hash` like in the test file, this may come in handy, but until then I don't care.

zlav (approved):
Looks good! I'm surprised the newLine parser that I wrote is causing the parser to hang. If you know why or have any more information you should comment that so we don't make this mistake again.

Changelog.md
zlav: Updated this in my new PR so ignore this comment.

PR #237, opened 2021-05-13
[DO NOT REVIEW] Format everything

scruffystuffs:
_Note: I do not intend to merge this, if we want to do this we can resubmit it later, this will conflict with current work.
You do not need to review this in depth, and doing so is a waste of time.  This is a POC pull request and will not be merged._

Here's what it looks like if we were to reformat the entire codebase using `ormolu`.

A few issues came up when running this naively:
- We use the `TypeApplications` extension everywhere, so many files failed initially.
- We have exactly 1 instance of postpositive qualified import, in `Strategy.Go.Transitive`.
  - postpositive import looks like `import Foo qualified as F` as opposed to the default (prepositive) of `import qualified Foo as F`
  - When we enable this extension to allow ormolu to scan this file, it picks up every prepositive import and converts it to postpositive.
 - `ormolu` DOES NOT pick up extensions from cabal, and may never do so.  We might need to script this in haskell using `cabal-the-library`.
 - `ormolu` DOES NOT currently attempt to recursively scan for files (in any mode), so invoking requires some file-finding scripting.
 - There was one instance of a re-ordered import causing a redundant import warning.

The final invocation which worked:
``` shell
ormolu \
  --check-idempotence \
  --mode inplace \
  --ghc-opt -XTypeApplications \
  --ghc-opt -XImportQualifiedPost \
  $(fd --type f '(src|test|app)/.*\.hs')
```

`fd` is [fd-find](https://github.com/sharkdp/fd), but any find tool can be used.  I was able to accomplish the same file list with shell globbing.

We should make a decision about post-positive vs. pre-positive qualified imports, and then apply that decision by either:
- Fixing the single postpositive import and rerunning with the default import style
- Not fixing it and converting every other import to postpositive (ormolu does it automatically)

### DIFF
--- a/app/fossa/Main.hs
+++ b/app/fossa/Main.hs
@@ -1,8 +1,7 @@
 module Main (main) where
 
-import Prelude
-
 import App.Fossa.Main (appMain)
+import Prelude
 
 main :: IO ()
 main = appMain
--- a/app/pathfinder/Main.hs
+++ b/app/pathfinder/Main.hs
@@ -1,8 +1,7 @@
 module Main (main) where
 
-import Prelude
-
 import App.Pathfinder.Main (appMain)
+import Prelude
 
 main :: IO ()
 main = appMain
--- a/src/App/Fossa/API/BuildLink.hs
+++ b/src/App/Fossa/API/BuildLink.hs
@@ -1,5 +1,5 @@
-{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE RecordWildCards #-}
 
 module App.Fossa.API.BuildLink
   ( getBuildURLWithOrg,
@@ -17,7 +17,7 @@ import Data.Text (Text)
 import Data.Text.Extra (showT)
 import Fossa.API.Types (ApiOpts (..))
 import Srclib.Types (Locator (..))
-import qualified Text.URI as URI
+import Text.URI qualified as URI
 import Text.URI.Builder
 import Text.URI.QQ (uri)
 
--- a/src/App/Fossa/API/BuildWait.hs
+++ b/src/App/Fossa/API/BuildWait.hs
@@ -6,13 +6,13 @@ module App.Fossa.API.BuildWait
   )
 where
 
-import qualified App.Fossa.FossaAPIV1 as Fossa
-import qualified App.Fossa.VPS.Scan.Core as VPSCore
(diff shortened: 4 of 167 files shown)


liftM:
Nice spike on this. If we were to hypothetically merge, two more things to consider:

1. We should pin the version of `ormolu` somehow, especially the version used in CI (so people aren't accidentally using different formatter versions and getting different results).
2. We should configure this to run in CI, and block commits.

scruffystuffs:
I don't know if we can realistically pin ormolu locally.  I think we can suggest a pin easily, but enforcing it seems pretty difficult.

I think we should avoid pinning the CI version, so that we always get the latest stable version.  There's also an ormolu github action that probably doesn't pin its version.  I think that we should have an error message on failure that reports the version of ormolu used, and instructs the author to check the version they're using locally.

A github bot that formats the entire codebase (in the current branch) on command, and commits the result to the branch, is probably the ideal solution here.  It's even easy to undo, just `git reset --hard` to the commit before the format, and the bot doesn't care. We could just leave a comment that says `@botname ormolu`

cnr:
I've been using [fourmolu](https://hackage.haskell.org/package/fourmolu) lately, which is a fork of ormolu with some customization options.

Ormolu was by far the best option when we started, because it doesn't choke on GHC extensions like most formatters do -- and it idempotently formats all haskell code, so running it multiple times on the same code results in the same formatted output

Fourmolu is similar:
- it's built on top of ormolu
- it has the same idempotence

..but it gives us much more reasonable formatting output. Ormolu's output at times is just terrible, and it way too aggressively kills whitespace

This is my `fourmolu.yaml` (you can just drop this in the spectrometer folder):
```yaml
indentation: 2
comma-style: leading # for lists, tuples etc. - can also be 'trailing'
record-brace-space: false # rec {x = 1} vs. rec{x = 1}
indent-wheres: true # 'false' means save space by only half-indenting the 'where' keyword
diff-friendly-import-export: true # 'false' uses Ormolu-style lists
respectful: true # don't be too opinionated about newlines etc.
haddock-style: single-line # '--' vs. '{-'
newlines-between-decls: 1 # number of newlines between top-level declarations
```

and haskell-language-server can be configured to use fourmolu instead of ormolu. In VSCode, this should be in the configuration section for the `haskell` extension

unfortunately that adds a manual step for setting up a new dev environment, because we can't use a project-level configuration file for hls

PR #238, opened 2021-05-17
Fix container scanning project flag.

zlav:
# Overview

We are currently not handling the `--project` flag or configuration file project field for container scanning. This PR addresses this concern and makes sure we will be uploading a project with the correct `project` name.

## Acceptance criteria
`fossa container analyze --project` properly sets the project field on upload

## Testing plan

I tested this by making my changes, building a binary, and seeing that it correctly uploaded a project with the correct project name. I used the binary that was created by github actions.

## Checklist

- [X] I linted (`hlint`) and formatted (`ormolu`) any files I touched in this PR.
- [X] I updated `CHANGELOG.md`. If the PR did not mark a release, update the `#Unreleased section at the top`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,5 @@
 # Unreleased
-- Fix `fossa container analyze --project` flag ([#238](https://github.com/fossas/spectrometer/pull/238)))
+- Fix `fossa container analyze` `--project` and `--revision` flags ([#238](https://github.com/fossas/spectrometer/pull/238)))
 # v2.5.16
 
 - Support for manually specified dependencies through `fossa-deps.yaml` ([#236](https://github.com/fossas/spectrometer/pull/236))
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,7 @@
 # Unreleased
+- Fix `fossa container analyze --project` flag ([#238](https://github.com/fossas/spectrometer/pull/238)))
+# v2.5.16
+
 - Support for manually specified dependencies through `fossa-deps.yaml` ([#236](https://github.com/fossas/spectrometer/pull/236))
 
 # v2.5.15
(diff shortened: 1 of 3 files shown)


Changelog.md
```diff
@@ -1,4 +1,7 @@
 # Unreleased
+- Fix `fossa container analyze --project` flag ([#238](https://github.com/fossas/spectrometer/pull/238)))
```
scruffystuffs, line 2: [nit]: This should probably also mention the revision flag, since that was fixed as well.

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -323,8 +329,9 @@ uploadContributors :: (Has (Lift IO) sig m, Has Diagnostics sig m) => ApiOpts ->
 uploadContributors apiOpts locator contributors = fossaReq $ do
   (baseUrl, baseOpts) <- useApiOpts apiOpts
 
-  let opts = baseOpts
-        <> "locator" =: locator
+  let opts =
+        baseOpts
+          <> "locator" =: locator
```
scruffystuffs, line 334: [nit]: I think the formatter should have one-lined this.  Let's trick it into doing that:
```suggestion
  let opts = baseOpts <> ("locator" =: locator)
```

scruffystuffs (approved):
LGTM, one formatting nit and one changelog nit.

src/App/Fossa/FossaAPIV1.hs
zlav: I agree, but ormolu did this everywhere it could. If there is a way we can force this that would be much better.

scruffystuffs: Ormolu tries to respect whitespace and newlining, so if we don't use newlines, it won't force us unless it thinks it has to.  The original was two lines, so ormolu went full-throttle on newlines.  But if we shorten it to one line, and then format again, I think it will let it go through.

PR #245, opened 2021-06-01
User defined deps

scruffystuffs:
# Overview

This PR adds the functionality for a user to specify their own dependencies, even if we cannot find or scan them.  This is referred to as _user-defined dependencies_, and works with the pre-existing _manually-specified dependencies_ (aka `fossa-deps.yml`, or just `fossa-deps` for this PR write-up).

To accomplish this, a few changes were necessary.

- Anything found in `fossa-deps` should no longer be part of our normal strategy mechanism, but should instead be a separate mechanism alongside standard analysis.
- Discovery may return zero projects and we can still upload results (provided there are results from `fossa-deps`).
- The `fossa-deps` parser now returns more specific error messages for some obvious footguns, so users can take clearer action in fixing the file.
- If `fossa-deps` exists, but fails to parse, we abort before discovery/analysis.  This is for user benefit: a user that has an incorrect `fossa-deps` would probably like to know about it before running the full scan, which can be very expensive.
- Analysis upload function just takes a `NonEmpty SourceUnit`.  We can't support anything else at the moment, so there's no need to pass the full project result structures.

Additionally, a few other changes came about:

- `fossa-deps.yml` can also be `fossa-deps.yaml`, and if both are present we will throw an error, since there is no way to know what the user wants.
- Post-analysis upload is refactored into its own function.
- `Path b t` gains instances for `ToString`, `ToText`, `ToLText`.
- Some markdown files were adjusted to match linting rules.

## Acceptance criteria

- Users can run user-defined dependencies through the new user-defined pipeline in core (fossas/team-analysis#446, fossas/team-analysis#447, fossas/team-analysis#448, fossas/team-analysis#449)
- Error messages and documentation are clear to someone who has not seen user-defined deps before.

## Testing plan

TESTING STILL IN-PROGRESS

- [x] Run malformed `fossa-deps` files (mostly unit-testable, migrate to unit tests before merge)
  - [x] File exists, but has no deps (can be due to incorrect names of sections)
  - [x] Top-level arrays have bad tables (partially unit tested)
- [x] Run with both `yml` and `yaml` extensions (should fail)
- [x] Run successful analyses through `--output`:
  - [x] With no `fossa-deps` file
  - [x] With manual/no user-def
  - [x] With user-def only
  - [x] With manual + user-def
- [ ] Run successful analysis through local core, check body
- [ ] Run successful analysis through staging/prod, build should complete

## Risks

I don't believe this change will cause any issue via additional sourceunits being added to the task queue.

## References

Closes fossas/team-analysis#445

## Notes

An important implementation detail is that we can merge all deps from the `fossa-deps` file into a single source unit, and upload that along with any source units we get from standard discovery/analysis.  This means that `referenced-dependencies` (standard fetchers) and `custom-dependencies` (user fetcher) are merged into a single source unit.

## Checklist

- [ ] I linted (`hlint`) and formatted (`ormolu`) any files I touched in this PR.
  - We are in-flux on competing formatters so I did not consistently format.
- [x] I updated `CHANGELOG.md`. If the PR did not mark a release, update the `#Unreleased section at the top`.


scruffystuffs:
Unit tests aren't working yet, I had to make a lot of breaking changes, so there's a decent amount of work left to fix what I broke.

cnr:
I have some general concerns about the way `fossa-deps.yml` is structured, especially in combination with the upcoming PR to support archive uploads

The `dependencies` field will soon be overloaded with three concepts:
1. dependencies that can be fetched from a registry (the current meaning of the `dependencies` field)
2. dependencies that users are defining that don't exist in a registry -- this PR
3. directories the user would like to scan for licenses -- upcoming PR

From a user perspective, this feels very, very confusing. The error messages definitely help, but it's still not clear to me why these concepts wouldn't be split up into three separate fields

scruffystuffs:
I don't know that the new license scan part should go in the `dependencies` field.  I think that that deserves its own section in the file.  As far as this PR, I'm happy to take suggestions, but I think the UX is pretty strong this way (it's the first thing I settled on before writing any code).  I thought it would be easier to add things in together than separate into a new section.

zlav:
Connor I originally agreed with you here entirely, but after floating the two files below (without user deps added) to Solomon and Deepak I've been persuaded to think overloading the `dependencies` field is more intuitive for users. The main feedback was that if the sections are split, a user may expect them to appear different in the Fossa UI, which is counterintuitive. The user will see each entry displayed as a dependency alongside each other in the UI. Other feedback I got was that the combination between selecting a `type` of dependency and having accurate error messages would be sufficient at preventing users from misusing types and expected fields. 

That said, I don't think the examples I have below are particularly great and I do understand the concerns. I think the field names I chose here are a big issue so if either of you have suggestions I'd be happy to change them and get more feedback from other Fossa members.

Overloading the existing file.
```yaml
dependencies:
- package: rubyXL
  type: gem
  version: 3.4.16
- package: Django
  type: license-scan
  path: vendor/Django-3.4.16.tar
- package: my-new-dependency
  type: user
  version: 1
```

Separate sections
```yml
dependencies:
- package: rubyXL
  type: gem
  version: 3.4.16

license-scanned-dependencies:
- package: Django
  path: vendor/Django-3.4.16.tar

user-dependencies:
- package: my-new-dependency
  type: user
  version: 1
  license: MIT
```

liftM:
> The `dependencies` field will soon be overloaded with three concepts:
> 
> 1. dependencies that can be fetched from a registry (the current meaning of the `dependencies` field)
> 2. dependencies that users are defining that don't exist in a registry -- this PR
> 3. directories the user would like to scan for licenses -- upcoming PR
>
> From a user perspective, this feels very, very confusing. The error messages definitely help, but it's still not clear to me why these concepts wouldn't be split up into three separate fields

I think having these as one field is fine. From a user's perspective, do users think of these three kinds of dependencies as different things? I would argue they don't. I would argue that users think of _all_ dependencies as `(name, version, license, etc.)` tuples (i.e. as (2)), and see (1) and (3) as _conveniences_ for constructing (2).

scruffystuffs:
Context: we had a meeting and mostly agreed on the split format instead of the overloaded format.  This comment explains the biggest reason for that decision (in my opinion)

I think the biggest issue is that if a user skims the docs, the overloaded format provides a really easy footgun (though a mild one).  With the split format, there's no footgun (except maybe concerning naming, which is a much smaller footgun, and can be shrunk even further).  In split format, it's overly clear that the sections don't behave the same way, and that the way to use them is different.

While it's important to have clear error messages, it's also important to try to limit the cases where a user encounters an error at all.  I believe the split format's clearer divide prevents silly mistakes much easier.

zlav:
@liftM I understand exactly what you are saying and this was my initial hesitation to breaking up the file into separate blocks. However, I do think that a user's mental model between "normal" deps, license scanning, and user-defined deps is different. 1 of them the user inputs a name and we magically find the thing, another the user must enter a path, and for the third, the user must add a license themselves. I think enforcing this mental model to the user is the correct thing to do as well if it doesn't already exist. I personally believe this is an issue with how modules are listed in CLI v1 and how "raw" types are allowed to be alongside "normal" modules. Separating the types may add a little more work, but only for the users who want additional functionality.

liftM:
I seriously disagree with this decision. We should use the overloaded format.

> I think the biggest issue is that if a user skims the docs, the overloaded format provides a really easy footgun (though a mild one).

What footgun?

> If the user's mental model doesn't fit with that, then their mental model is wrong, and this is the way to correct it.

No. The problem is that their mental model doesn't fit the model of _FOSSA's implementation of dependencies_, but it _does_ fit the model of _how people think about dependencies in general_. The fact that our current implementation is locator-centric isn't important, and forcing people to understand that model _isn't helpful_. The right product decision here is to bring our implementation closer to how people think about the problem, not forcing users to deeply understand our implementation.

---

Examining the split section proposal, quoted from above:

```yml
dependencies:
- package: rubyXL
  type: gem
  version: 3.4.16

license-scanned-dependencies:
- package: Django
  path: vendor/Django-3.4.16.tar

user-dependencies:
- package: my-new-dependency
  type: user
  version: 1
  license: MIT
```

First, some nits:

1. [nit] Why is there a `type: user` in user-dependencies? This is already implied by the section.
2. [nit] We should rename `package` to `name`, since e.g. the license scanned folders aren't always _packages_ per se.

But more substantively, I think this split and naming is more confusing than it is helpful.

1. Users _already know_ that different `type`s of dependencies do different things. It's _obvious_ that `type: gem` is not `type: node`, and also obvious that `type: license-scan` or `type: user` will be different too. Since the different implementation is already obvious, splitting out sections by types _doesn't provide extra value_.
    - In fact, I wouldn't be surprised if different `type`s of Locator dependencies wound up having different fields anyway! For example, we might augment `type: maven` with e.g. `groupID` or `artifactID` without adding those to the other `type`s.
2. Splitting out sections creates confusing implications for users that aren't actually important. For example, one of these sections is named `license-scanned-dependencies`. Does that mean other dependencies are _not_ license scanned? Of course not - but this is the immediate implication.

---

I could be convinced to allow split sections only if we name and implement the sections so they are aligned to the semantics of _the domain of dependencies_, not the semantics of _FOSSA's implementation_. For example, an acceptable UI to me would be:

```yml
referenced-dependencies: # Perhaps "managed-dependencies", but I prefer "referenced".
- type: gem
  name: rubyXL
  version: 3.4.16

vendored-dependencies: # Perhaps "local-dependencies", but I prefer "vendored".
- name: Django
  path: vendor/Django-3.4.16.tar

custom-dependencies:
- name: my-new-dependency
  version: 1
  license: MIT
```

scruffystuffs:
@liftM Other than nits and naming (which we were still bikeshedding, thanks for the input there), your split format is **exactly** what we proposed.  So, do you really disagree here?

I'm going to move forward with this as we proposed, but with your preferred names (because you just proposed the exact same thing with better names), since I'm 100% blocked if I don't move in *some* direction.

liftM:
@scruffystuffs I'm okay with move forward with my proposed alternative names (referenced, vendored, custom).

I think the overloaded format is _clearer_ because it maps closer to how users think about dependencies. But I'm okay to proceed with the split format with alternative names because I think those sections and names align well with the actual domain of how we think about dependencies.

My main objection was that I don't want the sections to be split on the basis of how FOSSA _implements_ dependencies, but I'm okay with a split along actual domain semantic differences. (In this case, fortunately, it seems the two happen to line up perfectly, but I think the rename is still important).

zlav:
@liftM I don't think we made it clear anywhere that the naming is not settled yet, but we should have. We all agreed that `license-scanned-dependencies` is a terrible name, and your point about `type: user` is accurate (I think that was sloppy copy and paste). The example was meant as an example format but not with decisions made about the actual keys. We are still discussing what we like and I think your example is pretty solid.

I do think your point about `type: mvn` is a good one but I don't believe that the format you suggested prevents that in the future. I think there is a fundamental difference between telling FOSSA to find a dependency in a registry and finding a dependency locally.

src/App/Fossa/Analyze.hs
```diff
@@ -257,26 +277,43 @@ analyze (BaseDir basedir) destination override unpackArchives filters = do
         -- Warn on contributor errors, never fail
         void . Diag.recover . runExecIO $ tryUploadContributors basedir apiOpts (uploadLocator uploadResult)
 
+
 data CountedResult
   = NoneDiscovered
   | FilteredAll Int
-  | FoundSome (NE.NonEmpty ProjectResult)
+  | ManualOnly SourceUnit
+  | FoundSome (NE.NonEmpty SourceUnit)
```
zlav, line 285: I'm not sure I understand why we have two different types here. I think separating `FoundSome` and `ManualOnly` adds a bit of confusion as well as the necessity to handle both cases near line 237. In my opinion this would look cleaner if we only used `FoundSome`.

scruffystuffs: I think you're right, this was necessary for a different attempt at merging the file with normal analysis.  I can revisit to check.

src/App/Fossa/YamlDeps.hs
```diff
@@ -0,0 +1,196 @@
...
+instance FromJSON ReferencedDependency where
+  parseJSON = withObject "ReferencedDependency" $ \obj ->
+    ReferencedDependency <$> obj .: "package"
+      <*> (obj .: "type" >>= depTypeParser)
+      <*> obj .:? "version"
+      <* forbidMembers "referenced dependencies" ["license", "description", "url"] obj
+
+instance FromJSON CustomDependency where
+  parseJSON = withObject "CustomDependency" $ \obj ->
+    CustomDependency <$> obj .: "package"
```
zlav, line 156: I think we agreed upon changing `package` to `name` for all use dependency sections.

src/App/Fossa/YamlDeps.hs
```diff
@@ -0,0 +1,196 @@
...
+      <*> obj .:? "url"
+      <* forbidMembers "custom dependencies" ["type"] obj
+
+-- | Parser insert to prevent specific fields from being used in parsers
+-- Primarily useful for rejecting aeson fields which should be reported with custom error messages
+--
+-- >  parseJSON = withObject "MyDataType" $ \obj ->
+-- >   MyDataType <$> obj .: "my-data-field"
+-- >     <* forbidMembers "Custom error message" ["badfield1", "badfield2"] obj
+forbidMembers :: Text -> [Text] -> Object -> Parser ()
```
zlav, line 169: This is awesome, eases most of the concerns we had about users entering incorrect data and expecting it to behave differently.

zlav (approved):
Looks great to me once the nits about `package` are addressed.

src/App/Fossa/YamlDeps.hs
```diff
@@ -0,0 +1,196 @@
...
+    length customDependencies
+  ]
+
+data YamlDependencies = YamlDependencies
+  { referencedDependencies :: [ReferencedDependency],
+    customDependencies :: [CustomDependency]
+  } deriving (Eq, Ord, Show)
+
+data ReferencedDependency = ReferencedDependency
+  { locDepPackage :: Text,
```
zlav, line 118: [nit] change the variable names from `package` to `name` as well.

src/App/Fossa/YamlDeps.hs
cnr: +1 -- this is great

src/DepTypes.hs
```diff
@@ -63,6 +63,7 @@ data DepType =
   | CargoType -- ^ Rust Cargo Dependency
   | RPMType -- ^ RPM dependency
   | URLType -- ^ URL dependency
+  | UserType -- ^ No registry or fetch work needed
```
cnr, line 66: Let's kill this constructor -- I don't think it's used anywhere

src/App/Fossa/YamlDeps.hs
```diff
@@ -0,0 +1,196 @@
...
+    toImport :: ReferencedDependency -> Locator
+    toImport ReferencedDependency {..} =
+      Locator
+        { locatorFetcher = depTypeToFetcher locDepType,
+          locatorProject = locDepName,
+          locatorRevision = locDepVersion
+        }
+
+    addEmptyDep :: Locator -> (Locator, SourceUnitDependency)
+    addEmptyDep loc = (loc, SourceUnitDependency loc [])
```
cnr, line 92: I think this might be clearer without the `unzip`

```suggestion
      buildImports = imports,
      buildDependencies = map addEmptyDep imports
    }
  where
    imports = map toImport $ NE.toList deps
    
    addEmptyDep :: Locator -> SourceUnitDependency
    addEmptyDep loc = SourceUnitDependency loc []
```

test/App/Fossa/YamlDepsSpec.hs
```diff
@@ -0,0 +1,64 @@
...
+
+getTestDataFile :: String -> SpecM a BS.ByteString
+getTestDataFile name = runIO . BS.readFile $ "test/App/Fossa/testdata/" <> name
+
+theWorks :: YamlDependencies
+theWorks = YamlDependencies references customs
+  where
+    references =
+      map
+        (uncurry3 ReferencedDependency)
```
cnr, line 32: [nit] I think using `uncurry3`/`uncurry5` here doesn't help very much; might be better to inline the constructor in the list

Changelog.md
```diff
@@ -1,136 +1,148 @@
-# 2.7.1
+# Spectrometer Changelog
+
+## Unreleased
+
+- Add support and documentation for user-defined dependencies ([#245](https://github.com/fossas/spectrometer/pull/245))
+- Allow using `.yml` or `.yaml` extensions for `fossa-deps` file, but not both ([#245](https://github.com/fossas/spectrometer/pull/245))
+- `fossa-deps` file is checked before running discovery/analysis, and is no longer run in parallel with other analysis functions ([#245](https://github.com/fossas/spectrometer/pull/245))
```
cnr, line 7: [nit] for changelog entries, unlike commit messages, use e.g. `Adds support ...` and `Allows ...`

docs/userguide.md
```diff
@@ -171,26 +172,31 @@ We support the following archive formats:
...
-  package: iron
+  name: iron
 - type: pip
-  package: Django
+  name: Django
   version: 2.1.7
 ```
-The `package` and `type` fields are required and specify the name of the dependency and where to find it. The `version` field is optional and specifies the preferred version of dependency.
+
+The `package` and `type` fields are required and specify the name of the dependency and where to find it. The `version` field is optional and specifies the preferred version of dependency.  Note that many version numbers can look like actual numbers in YAML, but they should always be strings!  You can quote them to make sure YAML treats them as strings, otherwise you may run into an error.
```
cnr, line 195: > You can quote them to make sure YAML treats them as strings, otherwise you may run into an error

I ran into this issue in the `yarn.lock` PR, which resulted creating the the `TextLike` type in `Data.Aeson.Extra`. It allows us to parse accidental ints/doubles as `Text`, which would prevent us from pushing this obligation to the user

src/App/Fossa/YamlDeps.hs
```diff
@@ -0,0 +1,196 @@
...
+          srcUserDepLicense = customLicense,
+          srcUserDepDescription = customDescription,
+          srcUserDepUrl = customUrl
+        }
+
+hasNoDeps :: YamlDependencies -> Bool
+hasNoDeps YamlDependencies{..} = all (0 ==)
+  [ length referencedDependencies,
+    length customDependencies
+  ]
```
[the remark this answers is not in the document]
scruffystuffs: First one doesn't work because the two lists are heterogeneous.  Once I realized that I should have just done the second.

src/App/Fossa/YamlDeps.hs
```diff
@@ -0,0 +1,196 @@
...
+findFossaDepsFile root = do
+  let ymlFile = root </> $(mkRelFile "fossa-deps.yml")
+      yamlFile = root </> $(mkRelFile "fossa-deps.yaml")
+  ymlExists <- doesFileExist ymlFile
+  yamlExists <- doesFileExist yamlFile
+  case (ymlExists, yamlExists) of
+    (True, True) -> fatalText "Found '.yml' and '.yaml' files when searching for fossa-deps, only one is permitted if present."
+    (True, False) -> pure $ Just ymlFile
+    (False, True) -> pure $ Just yamlFile
+    (False, False) -> pure Nothing
```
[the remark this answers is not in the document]
zlav: I'm not entirely sure I agree. I could see this saving us a few debugging calls in the future. I'm fine with us making sure we only use `.yml` for all documentation and debugging materials, but I think this is beneficial.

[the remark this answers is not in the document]
scruffystuffs: IDK about this.  Perhaps a compromise:

```haskell
  case (ymlExists, yamlExists) of
    (True, True) -> warn that we found both, but use the correct one
    (True, False) -> pure $ Just ymlFile
    (False, True) -> error, we found .yaml but only accept .yml
    (False, False) -> pure Nothing

File: src/App/Fossa/YamlDeps.hs (hunk 3 above)
Comment (scruffystuffs) [reply src/App/Fossa/YamlDeps.hs:169]: Should this be promoted to `Data.Aeson.Extra`?  Do we ever expect to use it outside of `fossa-deps`?  We can always move it later.

Also worth noting that generic aeson parsing does this automatically (if given a certain option), and we may need to try using that along with `-XDuplicateRecordFields` and `-Werror -Wambiguous-fields`.  Since we haven't taken direct input from files yet, we haven't experimented with generic JSON implementations.

**Review by liftM — APPROVED** [review]
Reviewed UI and docs, LGTM. Also I tried out some `fossa-deps.yml` uploads and they seemed to work end-to-end.

File: docs/userguide.md [hunk 12]
```diff
@@ -201,9 +207,63 @@ Supported dependency types:
 - `maven` - Maven dependencies that can be found at many different sources. Specified as `package: javax.xml.bind:jaxb-api` where the convention is `groupId:artifactId`.
 - `npm` - Javascript dependencies found at [npmjs.com](https://www.npmjs.com/).
 - `nuget` - .NET dependencies found at [NuGet.org](https://www.nuget.org/).
-- `python` - Python dependencies found at [Pypi.org](https://pypi.org/).
+- `python` - Python dependencies that are typically found at [Pypi.org](https://pypi.org/).
```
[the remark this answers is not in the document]
scruffystuffs: This was not changed from Zach's original PR.  I prefer calling it `python`, since we fetch it from `pypi` (via `pip`, `pipenv`, or eventually `poetry`, which does not use pip at all), not from `pip` specifically.  The use of `pip` is an implementation detail, and one that most people probably don't care about.  Python dependencies are like nodejs dependencies, they are unaware of package manager at runtime, they just get picked up because they're in the right place.

If we change it, it should be `pypi` instead of python.

docs/userguide.md
```diff
@@ -171,26 +172,31 @@ We support the following archive formats:
...
 
 FOSSA offers a way to manually upload dependencies provided we support the dependency type. Manually specifying dependencies is very helpful in the event your package manager is unsupported or you are using a custom and nonstandard dependency management solution.
 
-The FOSSA CLI will automatically read `fossa-deps.yml` in the root directory when `fossa analyze` is run and parse dependencies from it.
+The FOSSA CLI will automatically read `fossa-deps.yml` in the root directory (usually the current working directory) when `fossa analyze` is run and parse dependencies from it.
+The CLI also accepts `fossa-deps.yaml`, but will abort if it finds both.
 
 > Tip: Use a script to generate this file before running `fossa analyze` to keep your results updated.
 
+To manually specify a dependency, you must provide the package type, package name, and optionally a package version, under the `referenced-dependencies` array, as shown here:
```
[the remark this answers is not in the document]
scruffystuffs: That doc will need some bike-shedding from the team, and I don't think it should block this.  I'll ticket it.

docs/userguide.md
```diff
@@ -171,26 +172,31 @@ We support the following archive formats:
...
+referenced-dependencies:
 - type: gem
   package: iron
 - type: pip
   package: Django
   version: 2.1.7
 ```
-The `package` and `type` fields are required and specify the name of the dependency and where to find it. The `version` field is optional and specifies the preferred version of dependency.
+
+The `package` and `type` fields are required and specify the name of the dependency and where to find it. The `version` field is optional and specifies the preferred version of dependency.  Note that many version numbers can look like actual numbers in YAML, but they should always be strings!  You can quote them to make sure YAML treats them as strings, otherwise you may run into an error.
```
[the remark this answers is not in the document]
scruffystuffs: using `TextLike`, we now avoid this entirely

src/App/Fossa/YamlDeps.hs
cnr: I think it's useful enough to move to `Data.Aeson.Extra`

re: generic aeson: I'd like to avoid using it if possible. I like the explicit nature of `FromJSON` instances -- and it's usually minimal overhead to write those instances. It doesn't help that `DuplicateRecordFields` behaves really poorly in most places

It might be worthwhile to add the opposite of `forbidMembers` -- `expectOnlyMembers` or whatever -- that would fail if there were any other fields in the object

scruffystuffs: Though it would be easy to implement.

docs/userguide.md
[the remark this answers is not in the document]
zlav: I think `pypi` is actually more accurate. I'm fine if you change it to that Wes. I really don't like `pip` because that really doesn't clarify anything, it only describes what we do in the backend with the fetcher.

src/App/Fossa/YamlDeps.hs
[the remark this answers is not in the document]
cnr: I'm fine with supporting both, with the existing logic for only accepting one of them

We should also update our config file parser to accept `.fossa.yaml`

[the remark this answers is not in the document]
scruffystuffs: I'm not sure anymore.  I think supporting one is simpler to reason about, but we should still report cleanly.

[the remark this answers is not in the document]
zlav: Yea I agree with this, we can add support for both configuration files in another PR.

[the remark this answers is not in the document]
scruffystuffs: I'll ticket the config file

PR #247, opened 2021-06-04
PackageReference: parse version from xml attr

cnr:
# Overview

In fossa-cli and here, we're parsing versions incorrectly in `.csproj`/`.vbproj`/etc files

We were looking for `Version` as a child element, when it's actually an attribute:
- https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#packagereference
- https://cloud.google.com/functions/docs/writing/specifying-dependencies-dotnet

## Acceptance criteria

Versions are parsed correctly for packages in `PackageReference` elements

## Testing plan

The relevant test file was updated

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,11 @@
+# Unreleased
+
+- Fixes the dependency version parser for `.csproj`, `.vbproj`, and similar .NET files ([#247](https://github.com/fossas/spectrometer/pull/247))
+
 # v2.7.0
 
 - Adds support for the Conda package manager ([#226](https://github.com/fossas/spectrometer/pull/226))
+
 # v2.6.0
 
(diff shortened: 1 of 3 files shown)


zlav:
Great find although I think this is only correct for newer versions of `dotnet` but I don't believe it has always been true. This is the repository that I tested against and a [link to the csproj file](https://github.com/UiPath/orchestrator-powershell/blob/2baa031ec66337a812ff3f7d7a7976a5ff4087c8/UiPath.PowerShell/UiPath.PowerShell.csproj). The most recent master commit on the repository does follow the format you are fixing for, but I believe this format was allowed for quite some time. 

How difficult would it be to parse both version fields as optional and then choose one? I think this would be the correct thing to do.

cnr:
Interesting -- I'll have it parse both

zlav (approved):
Good catch!

PR #248, opened 2021-06-04
Display `fossa test`/`fossa report`/etc status updates by default in ansi-incompatible terminals

cnr:
## Overview

As of https://github.com/fossas/spectrometer/pull/239, `--debug` mode is required to see status updates in `fossa test` and similar commands when running in an ansi-incompatible environment, like CI. This means that, e.g., for `fossa test`, users would see no output until both project- and issue-scans were complete

## Testing plan

- Run `fossa test >/dev/null` -- status messages are still produced while we wait for project and issue scan completion


zlav (approved):
This is really great. Looks good to me!

PR #249, opened 2021-06-04
Use fourmolu as our preferred formatter

cnr:
We've long used ormolu as our preferred formatter, but I've often avoided using it because it often produces really poor output. Between aggressively removing whitespace, weird choices in terms of leading commas, and aggressive layout changes, leaving the source files unformatted was often the better option.

[fourmolu](https://hackage.haskell.org/package/fourmolu) is the answer. It's ormolu, but with some configuration options to make the output nicer. I've been using it on every added/modified file in my recent PRs, and I'm really satisfied with the result

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -27,7 +27,7 @@ _Example:_ Closes https://github.com/example/repo/issues/123.
 ## Checklist
 
 - [] I added tests for this PR's change.
-- [] I linted (`hlint`) and formatted (`ormolu`) any files I touched in this PR.
+- [] I linted and formatted (via `haskell-language-server`) any files I touched in this PR.
 - [] If this PR introduced a user-visible change, I added documentation into `docs/`.
 - [] I updated `Changelog.md`. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.
 - [] I linked this PR to any referenced GitHub issues.
--- a/HACKING.md
+++ b/HACKING.md
@@ -73,9 +73,11 @@ You do not need to enforce that `hlint` passes to submit a PR, but it does help 
 
 ## Formatting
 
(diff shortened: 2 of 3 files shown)


scruffystuffs (approved):
Yep, this is good.

I will note that I prefer trailing commas instead of leading (but not nearly enough to hold the line):

- Easier to type (personal opinion, but seems to hold when adjusting existing comma styles)
- More familiar to new haskellers
- Indentation shape makes it more where the block starts and ends (see below for example)

```haskell
-- leading
getDeps ::
  ( Has (Lift IO) sig m
  , Has Diagnostics sig m
  , Has ReadFS sig m
  , Has Exec sig m
  ) =>
  PomClosure.MavenProjectClosure ->
  m (Graphing Dependency)

-- leading with shape only
XXXXXXX XX
  X XXX XXXXXXXXX XXX X
  X XXX XXXXXXXXXXX XXX X
  X XXX XXXXXX XXX X
  X XXX XXXX XX X
  X XX
  XXXXXXXXXXXXXXXXXXXXXXXXXXXXXX XX
  X XXXXXXXXX XXXXXXXXXXX

-- trailing
getDeps ::
  ( Has (Lift IO) sig m,
    Has Diagnostics sig m,
    Has ReadFS sig m,
    Has Exec sig m,
  ) =>
  PomClosure.MavenProjectClosure ->
  m (Graphing Dependency)

-- trailing with shape only
XXXXXXX XX
  X XXX XXXXXXXXX XXX XX
    XXX XXXXXXXXXXX XXX XX
    XXX XXXXXX XXX XX
    XXX XXXX XX XX
  X XX
  XXXXXXXXXXXXXXXXXXXXXXXXXXXXXX XX
  X XXXXXXXXX XXXXXXXXXXX
```

PR #252, opened 2021-06-09
Format all files with fourmolu

cnr:
As a followup to making fourmolu a requirement in https://github.com/fossas/spectrometer/pull/249, this PR runs the formatter over all source files in the project.

The command I ran to do this is:
```
find src test -type f -name '*.hs' -exec fourmolu -m inplace --ghc-opt -XImportQualifiedPost --ghc-opt -XTypeApplications {} \;
```

Importantly, running that command a second time results in no modified files (so running fourmolu on our codebase is guaranteed to be an idempotent operation)

### DIFF
--- a/src/App/Fossa/Container.hs
+++ b/src/App/Fossa/Container.hs
@@ -1,48 +1,47 @@
 {-# LANGUAGE QuasiQuotes #-}
 {-# LANGUAGE RecordWildCards #-}
 
-module App.Fossa.Container
-  ( ImageText (..),
-    imageTextArg,
-    Locator (..),
-    SyftResponse (..),
-    ResponseArtifact (..),
-    ResponseSource (..),
-    ResponseDistro (..),
-    SourceTarget (..),
-    ContainerScan (..),
-    ContainerImage (..),
-    ContainerArtifact (..),
-    runSyft,
-    toContainerScan,
-    extractRevision,
-    parseSyftOutput,
-    parseSyftOutputMain,
-    dumpSyftScanMain,
-  )
-where
+module App.Fossa.Container (
+  ImageText (..),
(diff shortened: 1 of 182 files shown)


src/App/Fossa/Container.hs
```diff
@@ -132,114 +126,110 @@ instance FromJSON SourceTarget where
...
 -- Define Layer/Location type to capture layers in which a dep is found
 -- omitting "path" from the object to reduce noise
-newtype ContainerLocation
-  = ContainerLocation {
-    conLayerId :: Text
-  } deriving (Eq, Show, Ord)
+newtype ContainerLocation = ContainerLocation
+  { conLayerId :: Text
+  }
+  deriving (Eq, Show, Ord)
```
scruffystuffs, line 173: Should we one-line this?

src/Control/Carrier/TaskPool.hs
```diff
@@ -87,39 +92,42 @@ stopWhenDone State{..} act = do
...
+                writeTVar stQueued xs
+                addRunning
+                pure $
+                  mask $ \restore -> do
+                    res <- try $ restore x
+                    case res of
+                      Left err -> do
+                        _ <- sendIO $ atomically $ tryPutTMVar stError err
+                        pure ()
+                      Right () -> complete *> loop
```
scruffystuffs, line 112: Damn, so much for better treatment of whitespace.  But this is an edge case, after all.

scruffystuffs (approved):
I ran through a lot of files to see if there's any issue, but I don't think there's significant, and I don't think I need to look at everything for a format-only PR (plus a compile-error fix)

Let's merge this ASAP and require it going forward for all changes.

PR #255, opened 2021-06-14
`fossa analyze`: Output project information to stdout

cnr:
# Overview

`fossa analyze` (in non-`--output` mode) with the `--json` flag outputs project information (revision, project name, fossa url, etc) in a user-friendly textual format.

To make this information programmatically accessible, these changes output the same information to stdout in JSON format:
```json
{"project":{"name":"tmp28","branch":null,"revision":"1623707515","url":"https://app.fossa.com/projects/custom+23784%2ftmp28/refs/branch/master/1623707515","locator":"custom+23784/tmp28$1623707515"}}
```

## Acceptance criteria

Fossa project information is programmatically accessible from the output of `fossa analyze --json`

## Testing plan

Manual testing: to ensure the output is programmatically accessible, pipe the output of `fossa analyze --json` to `jq`. This should succeed without error

## Risks

None -- our other logging takes place on stderr (through our `Logger` effect), and this project information is the only thing being output to stdout

## References

Closes https://github.com/fossas/team-analysis/issues/441

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,26 +1,31 @@
 # Spectrometer Changelog
 
-## 2.9.0
-
+## v2.9.2
 - Adds JSON-formatted project information to the output of `fossa analyze` with `--json` ([#255](https://github.com/fossas/spectrometer/pull/255))
+
+## v2.9.1
+- Bump wiggins - Updated vps aosp-notice-file subcommand to upload ninja files & trigger async task. ([#272](https://github.com/fossas/spectrometer/pull/272))
+
+## v2.9.0
+
 - Fix an issue where stdout doesn't always flush to the console ([#265](https://github.com/fossas/spectrometer/pull/265))
 - Fix an issue when referenced-dependencies are not being uploaded ([#262](https://github.com/fossas/spectrometer/pull/262))
 - Adds support for `fossa-deps.json` ([#261](https://github.com/fossas/spectrometer/pull/261))
 - Adds support for `vendored-dependencies` to be licensed scanned ([#257](https://github.com/fossas/spectrometer/pull/257))
 
-## 2.8.0
+## v2.8.0
 
 - Adds support for `--branch` flag on `fossa container analyze` command ([#253](https://github.com/fossas/spectrometer/pull/253))
 - Adds support and documentation for user-defined dependencies ([#245](https://github.com/fossas/spectrometer/pull/245))
 - Allows using `.yml` or `.yaml` extensions for `fossa-deps` file, but not both ([#245](https://github.com/fossas/spectrometer/pull/245))
 - `fossa-deps` file is checked before running discovery/analysis, and is no longer run in parallel with other analysis functions ([#245](https://github.com/fossas/spectrometer/pull/245))
 
-## 2.7.2
+## v2.7.2
 
 - Updates the VSI Plugin.
 - Adds support for VSI powered dependency discovery as a strategy.
 
-## 2.7.1
+## v2.7.1
 
 - Adds support for Yarn v2 lockfiles ([#244](https://github.com/fossas/spectrometer/pull/244))
 - Fixes the dependency version parser for `.csproj`, `.vbproj`, and similar .NET files ([#247](https://github.com/fossas/spectrometer/pull/247))
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,7 @@
 - Adds support and documentation for user-defined dependencies ([#245](https://github.com/fossas/spectrometer/pull/245))
 - Allows using `.yml` or `.yaml` extensions for `fossa-deps` file, but not both ([#245](https://github.com/fossas/spectrometer/pull/245))
 - `fossa-deps` file is checked before running discovery/analysis, and is no longer run in parallel with other analysis functions ([#245](https://github.com/fossas/spectrometer/pull/245))
+- Adds JSON-formatted project information to the output of `fossa analyze` ([#255](https://github.com/fossas/spectrometer/pull/255))
 
 ## 2.7.2
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 ## 2.9.0
 
+- Adds JSON-formatted project information to the output of `fossa analyze` with `--json` ([#255](https://github.com/fossas/spectrometer/pull/255))
 - Fix an issue where stdout doesn't always flush to the console ([#265](https://github.com/fossas/spectrometer/pull/265))
 - Fix an issue when referenced-dependencies are not being uploaded ([#262](https://github.com/fossas/spectrometer/pull/262))
 - Adds support for `fossa-deps.json` ([#261](https://github.com/fossas/spectrometer/pull/261))
@@ -13,7 +14,6 @@
 - Adds support and documentation for user-defined dependencies ([#245](https://github.com/fossas/spectrometer/pull/245))
 - Allows using `.yml` or `.yaml` extensions for `fossa-deps` file, but not both ([#245](https://github.com/fossas/spectrometer/pull/245))
 - `fossa-deps` file is checked before running discovery/analysis, and is no longer run in parallel with other analysis functions ([#245](https://github.com/fossas/spectrometer/pull/245))
-- Adds JSON-formatted project information to the output of `fossa analyze` ([#255](https://github.com/fossas/spectrometer/pull/255))
 
 ## 2.7.2
 
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -134,12 +134,12 @@ analyzeMain workdir recordMode logSeverity destination project unpackArchives js
     $ case recordMode of
       RecordModeNone -> do
         basedir <- sendIO $ validateDir workdir
-        analyze basedir destination project unpackArchives jsonOutput enableVSI filters
+        doAnalyze basedir
       RecordModeRecord -> do
         basedir <- sendIO $ validateDir workdir
         (execLogs, (readFSLogs, ())) <-
-          runRecord @Exec . runRecord @ReadFS $
-            analyze basedir destination project unpackArchives jsonOutput enableVSI filters
+          runRecord @Exec . runRecord @ReadFS 
+          $ doAnalyze basedir
         sendIO $ saveReplayLog readFSLogs execLogs "fossa.debug.json"
       RecordModeReplay file -> do
         basedir <- BaseDir <$> P.resolveDir' workdir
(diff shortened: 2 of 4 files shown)


scruffystuffs:
I'm not sure that this is the right approach, from a UX perspective.  Yes, it definitely works for those who need that info programmatically, but it's not great for users who are just reading it in a log.  Also, at least one major customer is using the previous UX and parsing the output from text, so this will break them.

I think that this belongs behind a global flag:  `--info-format < plaintext | json | others-if-we-implement-them-later >`

I also think we should default to plaintext, since that's our current behavior, and most customers will probably prefer that.

cnr:
It's mostly unobtrusive in the visual log output -- and pretty clearly delineated from the visual information output:
```
[ INFO] Analyzing yarn project at /Users/connor/Desktop/tmp28/
[ INFO]
[ INFO] Using project name: `tmp28`
[ INFO] Using revision: `1623707515`
[ INFO] Using branch: `No branch (detached HEAD)`
[ INFO] ============================================================

      View FOSSA Report:
      https://app.fossa.com/projects/custom+23784%2ftmp28/refs/branch/master/1623707515

  ============================================================
{"project":{"name":"tmp28","branch":null,"revision":"1623707515","url":"https://app.fossa.com/projects/custom+23784%2ftmp28/refs/branch/master/1623707515","locator":"custom+23784/tmp28$1623707515"}}
``` 

(except with line-wrapping, of course)

This might grow some more if/when we add more info output -- though we've been displaying the same amount of information since the beginning of the project

---

In general I'd like to avoid adding yet another flag to the `analyze` command, and it feels like outputting project information in a programmatically-consumable way (in addition to the user-friendly output) is a really intuitive default

Curious to hear how others feel about this @zlav @liftM

scruffystuffs:
Oh nvm, I was commenting under the idea that this would replace some of the plaintext output.  But that's really obviously not true.

scruffystuffs:
Currently, if there's any error, then we don't report any of this to the user.  For now, I think we can live with this problem, but I'd like to find a way to provide things like inference/override results even if analysis fails, or without running analysis at all.

Changelog.md
```diff
@@ -6,6 +6,7 @@
 - Adds support and documentation for user-defined dependencies ([#245](https://github.com/fossas/spectrometer/pull/245))
 - Allows using `.yml` or `.yaml` extensions for `fossa-deps` file, but not both ([#245](https://github.com/fossas/spectrometer/pull/245))
 - `fossa-deps` file is checked before running discovery/analysis, and is no longer run in parallel with other analysis functions ([#245](https://github.com/fossas/spectrometer/pull/245))
+- Adds JSON-formatted project information to the output of `fossa analyze` ([#255](https://github.com/fossas/spectrometer/pull/255))
```
scruffystuffs, line 9: Let's make sure we move this to the correct section before merging.

src/App/Fossa/Analyze.hs
```diff
@@ -290,6 +290,8 @@ uploadSuccessfulAnalysis (BaseDir basedir) apiOpts metadata override units = do
   -- Warn on contributor errors, never fail
   void . Diag.recover . runExecIO $ tryUploadContributors basedir apiOpts (uploadLocator uploadResult)
 
+  logStdout . decodeUtf8 . Aeson.encode $ buildProjectSummary revision (uploadLocator uploadResult) buildUrl
```
scruffystuffs, line 293: Should we do this before trying contributors?  I don't know if it makes a difference.

scruffystuffs (approved):
LGTM, left a reminder about the changelog section.  We need to wait for @liftM in any case, but I'm good with this.

zlav:
The output format looks great, although I'm not sure I agree with outputting project information after every `fossa analyze` run. One big reason is that if we were to output the locator information from a flag we can always go back to outputting on every run, but if we do it this way we are committing to this for good.

My biggest reservation is that I think output by default is a little noisy and won't be used by the majority of people who use the CLI which may provoke more questions than save them. I would personally prefer something like `fossa analyze --json` that would behave similarly to how `fossa test` and `fossa report` behave, where the lack of `--json` results in only human-readable output. What do you guys think?

I'm not set in stone here, but I'm not yet convinced we should be adding more to the default output that may confuse users.

scruffystuffs:
I think that printing it everytime is the best move.  Like @cnr mentioned, it's a very intuitive default.  I don't think any users who are already successfully using the CLI will even notice unless they need it, and if they do ask questions, those questions are very easy to answer.

liftM:
I would really prefer if this were behind a `--json` flag. This reasoning is going to sound dumb, but IMO it looks sloppy during demos and first runs to emit unexplained blobs of output.

zlav:
I also agree with that @liftM here. I think it makes the output for the majority of users who don't care to use this feature look worse. If this was a widely requested feature I think I'd agree with outputting it every time, but since this was the first request I'm not convinced.

scruffystuffs:
> Unexplained blobs of output

We can add a log message that says "Emitting project info as JSON".  I don't think that "unexplained" is an unaddressable concern.

I completely disagree that this is sloppy.  I think that most people who don't care about it won't even notice that it's there.  And as @cnr mentioned, we previously weren't outputting anything on stdout.  

I think the real winning argument here is that outputting this is a really intuitive default behavior, and there's a lot of precedent for it in unix tools whose job is to output data to stdout, but have long-running processes to get there (`curl` is a top-tier example here, but there are many others).

liftM:
> I think that most people who don't care about it won't even notice that it's there.

I disagree with this. In my experience, people don't _ignore_ output that they don't understand. Usually the reaction is something like "wait, what is this? Am I doing something wrong? Am I supposed to do something with this output? Is there a bug? Is this broken?". Unexplained output causes _tons_ of confusion (and support tickets).

> I think the real winning argument here is that outputting this is a really intuitive default behavior

Intuitive to who? The average user on their first run experience isn't expecting to script around this. They're just going to be confused by this blob.

> We can add a log message that says "Emitting project info as JSON".

This still feels like a subpar experience to me. We shouldn't provide output just because we can - we should provide output when it's asked for and when we expect the user to use that output. Otherwise, users feel the same confusion - "what should I do with this?".

> there's a lot of precedent for it in unix tools whose job is to output data to stdout

A lot of our users aren't familiar with Unix tools. Many of them work on Windows for their day jobs, or they're not experienced sysadmins. And in either case, I don't think there's any precedence for this behavior among user-facing vendor tools. Does `snyk` do this? What about `docker`? What about `kubectl`? What about `npm`?

People _expect_ Unix tools to be unfriendly, because they carry a ton of baggage. They don't expect that out of new tools.

liftM (dismissed):
Blocking until UX discussion is resolved.

scruffystuffs:
@liftM @zlav and I discussed this offline.  We are all in favor of putting this behind a flag, but we also agree that `--json` is confusing given the existence of `--output`, so we'd like to come up with a new name for the flag that this hides behind.

src/App/Fossa/Main.hs
```diff
@@ -269,6 +269,7 @@ analyzeOpts =
   AnalyzeOptions
     <$> switch (long "output" <> short 'o' <> help "Output results to stdout instead of uploading to fossa")
     <*> flagOpt UnpackArchives (long "unpack-archives" <> help "Recursively unpack and analyze discovered archives")
+    <*> flagOpt PrintMetadata (long "print-metadata" <> help "Output project metadata. Useful for communicating with the FOSSA API")
```
scruffystuffs, line 272: [nit]:
```suggestion
    <*> flagOpt PrintMetadata (long "print-metadata" <> help "Output project metadata to the console. Useful for communicating with the FOSSA API")
```

src/App/Fossa/Analyze.hs
```diff
@@ -290,7 +295,7 @@ uploadSuccessfulAnalysis (BaseDir basedir) apiOpts metadata override units = do
   -- Warn on contributor errors, never fail
   void . Diag.recover . runExecIO $ tryUploadContributors basedir apiOpts (uploadLocator uploadResult)
 
-  logStdout . decodeUtf8 . Aeson.encode $ buildProjectSummary revision (uploadLocator uploadResult) buildUrl
+  if fromFlag PrintMetadata printMetadata then logStdout . decodeUtf8 . Aeson.encode $ buildProjectSummary revision (uploadLocator uploadResult) buildUrl else pure ()
```
scruffystuffs, line 298: [formatting]: This should not be a one-liner

src/App/Fossa/Analyze.hs
```diff
@@ -147,7 +150,7 @@ analyzeMain workdir recordMode logSeverity destination project unpackArchives en
             let effects = analyzeEffects journal
             runReplay @ReadFS (effectsReadFS effects)
               . runReplay @Exec (effectsExec effects)
-              $ analyze basedir destination project unpackArchives enableVSI filters
+              $ analyze basedir destination project unpackArchives printMetadata enableVSI filters
```
scruffystuffs, line 153: [nit]: This analyze call should be bound in a `where` clause so we don't have to repeat the changes three times.
```suggestion
              $ doAnalyze
  where
    doAnalyze = analyze basedir destination project unpackArchives printMetadata enableVSI filters
```

src/App/Fossa/Main.hs
```diff
@@ -90,12 +90,12 @@ appMain = do
...
       let analyzeOverride = override{overrideBranch = analyzeBranch <|> ((fileConfig >>= configRevision) >>= configBranch)}
       if analyzeOutput
-        then analyzeMain analyzeBaseDir analyzeRecordMode logSeverity OutputStdout analyzeOverride analyzeUnpackArchives analyzeVSIMode analyzeBuildTargetFilters
+        then analyzeMain analyzeBaseDir analyzeRecordMode logSeverity OutputStdout analyzeOverride analyzeUnpackArchives analyzePrintMetadata analyzeVSIMode analyzeBuildTargetFilters
         else do
           key <- requireKey maybeApiKey
           let apiOpts = ApiOpts optBaseUrl key
           let metadata = maybe analyzeMetadata (mergeFileCmdMetadata analyzeMetadata) fileConfig
-          analyzeMain analyzeBaseDir analyzeRecordMode logSeverity (UploadScan apiOpts metadata) analyzeOverride analyzeUnpackArchives analyzeVSIMode analyzeBuildTargetFilters
+          analyzeMain analyzeBaseDir analyzeRecordMode logSeverity (UploadScan apiOpts metadata) analyzeOverride analyzeUnpackArchives analyzePrintMetadata analyzeVSIMode analyzeBuildTargetFilters
```
scruffystuffs, line 98: [nit]: This `analyzeMain` should be in a let binding to deduplicate the code.
```haskell
let doAnalyzeMain destination = analyzeBaseDir analyzeRecordMode logSeverity destination analyzeOverride analyzeUnpackArchives analyzePrintMetadata analyzeVSIMode analyzeBuildTargetFilters
```

scruffystuffs (approved):
Tentative approval once we agree on the final name

liftM (approved):
LGTM, with final nit: can you rename `project.locator` to `project.id`, and `project.name` to `project.title`?

scruffystuffs:
I'm not sure `title` is correct.  I believe that name is the `package` field of a locator.

zlav:
1. I think `id` is absolutely correct.
2. I'm not sold on title, since this differs from the `--title` flag that we offer in the CLI. If a user runs `fossa analyze --title tmp --json`, the `title` in the output is going to differ. More confusingly the flag that does align is `--project`. Personally I think name is the most correct.

liftM:
Wait, what? Why would someone need the parsed locator? Why are we providing that field?

zlav:
We settled on the following after a discussion over slack about fields a user would be interested in.
```
{
  "project": package portion of revision locator
  "revision": revision portion of revision locator
  "branch": branch of revision
  "url": web app URL
  "id": full locator of revision
}
```

The main focus is on
- things you'd use as CLI flags
- link to the web app
- an ID to use with the API

PR #256, opened 2021-06-14
Add simpler effect interpreter machinery

cnr:
Effects are intimidating to implement and maintain; this makes it easier. Introduces `SimpleC`, `interpret`, `SimpleStateC`, `interpretState`.

See the haddocks in `Control.Carrier.Simple` for details.

This is the effect implemented in the `Control.Carrier.Simple` haddocks:
```hs
----- Effect

data Filesystem m a where
  ReadFile :: FilePath -> Filesystem m String
  WriteFile :: FilePath -> String -> Filesystem m ()

readFile :: Has Filesystem sig m => FilePath -> m String
readFile file = send (ReadFile file)

writeFile :: Has Filesystem sig m => FilePath -> String -> m ()
writeFile file content = send (WriteFile file content)

----- Interpreters

type FilesystemC = SimpleC Filesystem

runFilesystemIO :: Has (Lift IO) sig m => FilesystemC m a -> m a
runFilesystemIO = interpret $ \case
  ReadFile file -> sendIO (Prelude.readFile file)
  WriteFile file content -> sendIO (Prelude.writeFile file content)

runFilesystemDumb :: Algebra sig m => FilesystemC m a -> m a
runFilesystemDumb = interpret $ \case
  ReadFile _ -> pure "content!"
  WriteFile _ _ -> pure ()

----- Examples

addNewlineToEndOfFile :: Has Filesystem sig m => FilePath -> m ()
addNewlineToEndOfFile file = do
  content <- readFile file
  writeFile file (content ++ "\n")

-- does the thing!
example1 :: IO ()
example1 = runM $ runFilesystemIO (addNewlineToEndOfFile "/path/to/file")

-- does nothing
example2 :: ()
example2 = run $ runFilesystemDumb (addNewlineToEndOfFile "/path/to/file")

-- "content!"
example3 :: String
example3 = run $ runFilesystemDumb (readFile "/path/to/file")
```

### DIFF
--- a/spectrometer.cabal
+++ b/spectrometer.cabal
@@ -36,6 +36,7 @@ common lang
     ImportQualifiedPost
     InstanceSigs
     KindSignatures
+    LambdaCase
     MultiParamTypeClasses
     NamedFieldPuns
     NamedWildCards
@@ -171,6 +172,7 @@ library
     Control.Carrier.Diagnostics.StickyContext
     Control.Carrier.Finally
     Control.Carrier.Output.IO
+    Control.Carrier.Simple
     Control.Carrier.StickyLogger
     Control.Carrier.TaskPool
     Control.Carrier.Threaded
--- a/src/App/Fossa/Analyze/GraphBuilder.hs
+++ b/src/App/Fossa/Analyze/GraphBuilder.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE GADTs #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE UndecidableInstances #-}
 
 -- | Legacy/deprecated GraphBuilder interface. Kept for compatibility with old code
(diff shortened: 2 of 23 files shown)


scruffystuffs (approved):
LGTM, awesome helper.  Why doesn't fused-effects do this normally?

cnr:
This approach incurs a slight performance hit.

In some cases, we were able to use monad transformers in the shape, e.g., `newtype ReadFSIO = ReadFSIO { runReadFSIO :: m a }`, where no additional structure was being added to `m a` -- which is great! because `>>=` for `ReadFSIO` is identical to `>>=` for the underlying `m a`, so the wrapper would get optimized away.

But now with these transformers, everything has a `ReaderC` wrapper, which is harder to optimize away. And ghc is bad at inlining across module boundaries.

It turns out that the performance hit doesn't matter much for our usecase though, because we're _overwhelmingly_ IO-bound (and especially command-execution-bound), so making effects more approachable/maintainable feels like the right tradeoff to me.

PR #257, opened 2021-06-15
License scan dependencies

zlav:
# Overview

This PR adds support for license scanning vendored dependencies in a user's project. This initial implementation relies on the archive uploader. 

These dependencies can be specified in the `fossa-deps.yml` file as `vendored-dependencies`
```yml
vendored-dependencies:
- name: Django
  path: vendor/Django-3.4.16.tar
```

## Acceptance criteria

Adding a dependency to the `fossa-deps.yml` file causes it to be archive uploaded, license scanned on the backend, and displayed in the projects list of dependencies.

## Testing plan

Validate that the archive upload works end to end
- Create a `fossa-deps.yml` file with a dependency in the vendored section
- Run `fossa analyze`
- Inspect the project in the FOSSA UI and validate that the vendored dependency is present.=

> Need to make sure files that don't exist are handled correctly.

## Risks

I am most concerned about decisions that won't be reversible in the future. The biggest risk is unintuitive behavior to the user, so if anyone sees a spot a warning or an error should exist let me know.

## References

https://github.com/fossas/spectrometer/pull/245

Closes https://github.com/fossas/team-analysis/issues/544

## Checklist

- [X] I added tests for this PR's change.


src/App/Fossa/Analyze.hs
```diff
@@ -232,7 +234,7 @@ analyze (BaseDir basedir) destination override unpackArchives enableVSI filters
   -- This is done because the VSI discover function requires more information than other discover functions do, and only matters for analysis.
   let discoverFuncs' = discoverFuncs ++ [vsiDiscoverFunc enableVSI destination]
 
-  manualSrcUnit <- analyzeFossaDepsYaml basedir
+  (manualSrcUnit, vendoredDeps) <- analyzeFossaDepsYaml basedir
```
cnr, line 237: I think it would substantially simplify the logic here if we:
- move all of the archive upload logic to be called as part of `analyzeFossaDepsYaml` (possibly pass a `Maybe ApiOpts` in there?)
- ignore archives in `--output` mode.

I'd suggest that if we find vendored dependencies in `fossa-deps.yml` in `--output` mode, we'd just log a warning and ignore those dependencies for the scan

src/App/Fossa/ArchiveUploader.hs
```diff
@@ -0,0 +1,119 @@
...
+        , locatorRevision = Just $ archiveVersion arc
+        }
+
+    addDepList :: Locator -> SourceUnitDependency
+    addDepList loc = SourceUnitDependency loc []
+
+compressFile :: Path Abs Dir -> Path Abs Dir -> FilePath -> IO FilePath
+compressFile outputDir directory fileToTar = do
+  -- Without using `fromAbsDir` for each of these directories, the conversion
+  -- is incorrect. `show outputDir` gives an incorrect result even though it typechecks.
```
cnr, line 107: [no action] another option here is `toFilePath :: Path a b -> FilePath` -- though using the specific `fromAbsDir` / `fromAbsFile` / etc is definitely much more descriptive

the reason `show` doesn't work here is that `show` behaves similar to the way it would for `FilePath` (or `Text` or other string types):
```hs
>>> show "foo"
\"foo\"
```

the result gets quoted

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -236,6 +271,87 @@ getLatestBuild apiOpts ProjectRevision{..} = fossaReq $ do
...
+  m (Maybe C.ByteString)
+archiveBuildUpload apiOpts archiveProjects = runEmpty $
+  fossaReqAllow401 $ do
+    (baseUrl, baseOpts) <- useApiOpts apiOpts
+
+    let opts = "dependency" =: True <> "rawLicenseScan" =: True
+        archiveBuildContext = "Queuing a build for an archive project"
+    -- The response appears to either be "Created" for new builds, or an error message for existing builds.
+    -- Making the actual return value of "Created" essentially worthless.
+    resp <- context archiveBuildContext $ req POST (archiveBuildURL baseUrl) (ReqBodyJson archiveProjects) bsResponse (baseOpts <> opts)
```
cnr, line 292: [nit] I think it's easier to read `context` when it's inlined -- so I'd inline `archiveBuildContext` and put a newline after `$` for spacing

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -236,6 +271,87 @@ getLatestBuild apiOpts ProjectRevision{..} = fossaReq $ do
...
+  Text ->
+  Text ->
+  m SignedURL
+getSignedURL apiOpts revision packageName = fossaReq $ do
+  (baseUrl, baseOpts) <- useApiOpts apiOpts
+
+  let opts = "packageSpec" =: packageName <> "revision" =: revision
+      signedURLContext = "Retrieving a signed S3 URL"
+
+  response <- context signedURLContext $ req GET (signedURLEndpoint baseUrl) NoReqBody jsonResponse (baseOpts <> opts)
```
cnr, line 312: [nit] same here wrt `context`

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -236,6 +271,87 @@ getLatestBuild apiOpts ProjectRevision{..} = fossaReq $ do
...
+  m String
+archiveUpload signedArcURI arcFile = fossaReq $ do
+  let arcURL = URI.mkURI $ signedURL signedArcURI
+
+  case arcURL >>= useHttpsURI of
+    Nothing -> case arcURL of
+      Nothing -> fatalText ("Error attempting to archive upload file: " <> T.pack arcFile <> ". No signed URL supplied")
+      Just resultURL -> fatalText ("Invalid URL: " <> URI.render resultURL)
+    Just (url, options) -> do
+      let archiveContext = "Attempting to archive upload a project"
```
cnr, line 330: [nit]
```suggestion
      let archiveContext = "Uploading project archive"
```

(also inline it like the other [nit] comments)

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -236,6 +271,87 @@ getLatestBuild apiOpts ProjectRevision{..} = fossaReq $ do
...
+      res <- context archiveContext $ reqCb PUT url (ReqBodyFile arcFile) lbsResponse options (pure . requestEncoder)
+      pure $ show res
```
cnr, line 332: use `decodeUtf8` from `Data.String.Conversion` instead of `show` on bytestrings

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -236,6 +271,87 @@ getLatestBuild apiOpts ProjectRevision{..} = fossaReq $ do
...
+  SignedURL ->
+  FilePath ->
+  m String
+archiveUpload signedArcURI arcFile = fossaReq $ do
+  let arcURL = URI.mkURI $ signedURL signedArcURI
+
+  case arcURL >>= useHttpsURI of
+    Nothing -> case arcURL of
+      Nothing -> fatalText ("Error attempting to archive upload file: " <> T.pack arcFile <> ". No signed URL supplied")
+      Just resultURL -> fatalText ("Invalid URL: " <> URI.render resultURL)
```
cnr, line 328: I might be misunderstanding these branches, but I think the `Nothing` case here is actually the "invalid URL" branch, and the `Just` branch here is actually successful/fine -- so no error wrt the URL

I think we can use `fromMaybeText` from `Control.Error.Diagnostics` to help us here:
```hs
fromMaybeText :: Has Diagnostics sig m => Text -> Maybe a -> m a
```

like so:
```hs
let arcURL = URI.mkURI $ signedURL signedArcURI

uri <- fromMaybeText ("Invalid URL: " <> signedURL signedArcURI) arcURL
(url, options) <- fromMaybeText ("Invalid HTTPS URI: " <> T.pack (show uri)) (useHttpsURI uri)
```

cnr (commented):
Overall looks really good -- one major suggestion and some nits

src/App/Fossa/FossaAPIV1.hs
zlav: That's fair `archiveBuildContext` is only used in one place as well so its not really necessary to do this.

src/App/Fossa/FossaAPIV1.hs
zlav: The `Just` branch here is not successful, it signifies what happens if a user supplies a URL, but cannot be converted to a valid `HTTPS` URI by `useHttpsURI`. The Nothing branch is what happens if `URI.mkURI` fails. The reason I split these was to give better messages.

Regardless, your suggestion still works better here so I'm going to use it.

src/App/Fossa/Analyze.hs
zlav: I strongly agree with your idea to condense the upload logic into part of `analyzeFossaDepsYaml`, but I don't agree with omitting archive deps from the `--output` output. They're easy add and should make debugging projects better. I don't see the benefit of omitting them, other than simplifying the code. Instead of `apiOpts`, we could pass in the `ScanDestination`.

zlav: I addressed your first point and condensed the archive logic into the yamldeps file. This really cleaned up Analyze. It makes the upload portion of analyze slightly more hidden, but I think it looks cleaner.

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -236,6 +272,85 @@ getLatestBuild apiOpts ProjectRevision{..} = fossaReq $ do
...
+    req GET (signedURLEndpoint baseUrl) NoReqBody jsonResponse (baseOpts <> opts)
+  pure (responseBody response)
+
+---------- The archive upload function uploads the file it is given directly to the signed URL it is provided.
+
+archiveUpload ::
+  (Has (Lift IO) sig m, Has Diagnostics sig m) =>
+  SignedURL ->
+  FilePath ->
+  m String
```
scruffystuffs, line 323: [nit]: Should we just return `m ()`?  We aren't using this string at all, so why bother keeping it around?

scruffystuffs (changes requested):
Left a ton of nits, but found a major issue that might allow files with no dependencies.

This should also be formatted

src/App/Fossa/FossaAPIV1.hs
zlav: I think returning m () is valid. The only reason I could see keeping it around is if the API was modified to return actionable information. I had it this way so I could validate we got "Created", but I removed that change for the final PR.

src/App/Fossa/ArchiveUploader.hs
```diff
@@ -0,0 +1,140 @@
...
+  where
+    updateArcName :: Text -> Archive -> Archive
+    updateArcName updateText arc = arc{archiveName = updateText <> "/" <> archiveName arc}
+
+-- archiveNoUploadSourceUnit exists for when users run `fossa analyze -o` and do not upload their source units.
+archiveNoUploadSourceUnit :: [VendoredDependency] -> Maybe SourceUnit
+archiveNoUploadSourceUnit deps = Just $ archivesToSourceUnit (unsafeVendoredToArchive <$> deps)
+
+unsafeVendoredToArchive :: VendoredDependency -> Archive
+unsafeVendoredToArchive dep = Archive (vendoredName dep) (fromMaybe "" $ vendoredVersion dep)
```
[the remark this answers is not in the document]
zlav: I called it `unsafe` because VendoredDependency has an optional version field whereas Archive requires it. This function is used when `fossa analyze -o` is run to output the deps without having to hash the files to acquire versions.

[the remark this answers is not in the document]
zlav: I renamed it to `forceVendoredToArchive` but if you have a better name let me know.

src/App/Fossa/FossaAPIV1.hs
scruffystuffs: If we do decide to return something, we should make it a real datatype, rather than just a string.

zlav: Agreed, for now I'm leaving it as `m ()`

src/App/Fossa/ArchiveUploader.hs
```diff
@@ -0,0 +1,142 @@
...
+    addDepList loc = SourceUnitDependency loc []
+
+compressFile :: Path Abs Dir -> Path Abs Dir -> FilePath -> IO FilePath
+compressFile outputDir directory fileToTar = do
+  -- Without using `fromAbsDir` for each of these directories, the conversion
+  -- is incorrect. `show outputDir` gives an incorrect result even though it typechecks.
+  -- let finalFile = fromAbsDir outputDir </> fileToTar
+  let finalFile = toString outputDir </> fileToTar
+  -- entries <- Tar.pack (fromAbsDir directory) [fileToTar]
+  entries <- Tar.pack (toString directory) [fileToTar]
```
scruffystuffs, line 132: [nit]: comments are probably not needed here.
```suggestion
  let finalFile = toString outputDir </> fileToTar
  entries <- Tar.pack (toString directory) [fileToTar]
```

scruffystuffs (approved):
LGTM, a few leftover nits, but this is ready to go.

src/App/Fossa/ArchiveUploader.hs
```diff
@@ -0,0 +1,142 @@
...
+  let build = toBuildData arcs
+      srcUnit =
+        SourceUnit
+          { sourceUnitName = "archive deps"
+          , sourceUnitManifest = "archive deps"
+          , sourceUnitType = "archive-uploaded-dependencies"
+          , sourceUnitBuild = Just build
+          , additionalData = Nothing
+          }
+  srcUnit
```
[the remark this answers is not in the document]
zlav: Oh yea, this is leftover from when I initially implemented the function.

PR #258, opened 2021-06-15
Hlint everything

scruffystuffs:
# Overview

We should keep the code linted at all times.  This PR removes all outstanding lint errors, and updates the config file to ignore the invalid ones.

We are not checking for:
* `Redundant bracket` -> Extra parenthesis can make code easier to read, so we allow it.
  * A good example is when using named infix functions: ```hello darkness `my` old friend``` as opposed to ```(hello darkness) `my` (old friend)```.  With the unnecessary parens, it's a lot easier to immediately understand the groupings.
* `Use list comprehension` -> Some list comprehensions are harder to read than their `if a then b else c` counterparts.
  * Compare `\x -> if x > 0 then [x -2] else []` to hlint's suggestion `\x -> [x -2 | x > 0]`.  Nearly unreadable without an already-ingrained instinct for haskell's list comprehension syntax.

## Acceptance criteria

Code builds, no real semantic changes.

## Risks
If there are any risks, then this is a bad change

## References

Closes fossas/team-analysis#584

## Checklist

- [x] I updated `Changelog.md`. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.

### DIFF
--- a/src/Strategy/Maven/PluginStrategy.hs
+++ b/src/Strategy/Maven/PluginStrategy.hs
@@ -50,7 +50,7 @@ buildGraph PluginOutput{..} = run $
         , dependencyName = artifactGroupId <> ":" <> artifactArtifactId
         , dependencyVersion = Just (CEq artifactVersion)
         , dependencyLocations = []
-        , dependencyEnvironments = if "test" `elem` artifactScopes then [EnvTesting] else []
+        , dependencyEnvironments = [EnvTesting | "test" `elem` artifactScopes]
         , dependencyTags =
             M.fromList $
               ("scopes", artifactScopes) :
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,1 +1,4 @@
 - arguments: [-XTypeApplications, -XNumericUnderscores]
+# Often, code is more readable with extra parens, or less readable with list comprehensions
+- ignore: {name: "Redundant bracket"}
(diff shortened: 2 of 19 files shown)


src/Strategy/Maven/PluginStrategy.hs
```diff
@@ -50,7 +50,7 @@ buildGraph PluginOutput{..} = run $
         , dependencyName = artifactGroupId <> ":" <> artifactArtifactId
         , dependencyVersion = Just (CEq artifactVersion)
         , dependencyLocations = []
-        , dependencyEnvironments = if "test" `elem` artifactScopes then [EnvTesting] else []
+        , dependencyEnvironments = [EnvTesting | "test" `elem` artifactScopes]
```
zlav, line 53: I think this is a scenario where the previous `if else` block is more clear. Can we ignore warnings for this issue too?

scruffystuffs: This is already covered by `Use list comprehesion`.  We can just revert this one.

zlav: Ah, didn't realize this was covered by that ignore.

zlav (approved):
Love the addition of the ignores. Looks good

cnr (approved):
I'm shocked this diff is so small

PR #259, opened 2021-06-15
Makefile proposal

scruffystuffs:
# Overview

This is a 15-min attempt at producing a convenience makefile.  This is just to see how people feel about introducing something like this into the codebase.  If we all like it, we can merge it, but if not, we can throw it away.

This can greatly help us when introducing linting and formatting into CI, although it's not necessary.

## Acceptance criteria

* There's a makefile
* It works for macOS and linux

## Checklist

- [x] I updated `Changelog.md`. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.


.gitignore
```diff
@@ -8,5 +8,10 @@ dist-newstyle/
...
 *.hp
 
-# vendor directory
-vendor
\ No newline at end of file
+# Development files/directories
+vendor/
+sandbox/
+.vscode/
+# Executables
```
liftM, line 15: [nit] Consistent whitespace between sections.

```suggestion

# Executables
```

.gitignore
```diff
@@ -8,5 +8,10 @@ dist-newstyle/
 *.prof
 *.hp
 
-# vendor directory
-vendor
\ No newline at end of file
+# Development files/directories
+vendor/
+sandbox/
```
liftM, line 13: Why `sandbox`? Are multiple people using this convention? I'm personally already using `.scratch`, with my own folder structure for test projects.

Makefile
```diff
@@ -0,0 +1,42 @@
...
+
+# Copy the built binary into the local root
+fossa:
+	cp $(shell cabal list-bin fossa) ./fossa
+
+install-local: fossa
+
+# Run analysis on the sandbox directory
+sandbox: fossa
+	./fossa analyze --output --debug --record ./sandbox > fossa.json
```
liftM, line 22: Like above, I'd remove this unless other folks are already using `./sandbox` as a convention.

Makefile
```diff
@@ -0,0 +1,42 @@
...
+	cabal test
+
+# Dogfood the dev version
+analyze:
+	cabal run fossa -- analyze --output --debug --filter 'cabal@./'
+
+# Copy the built binary into the local root
+fossa:
+	cp $(shell cabal list-bin fossa) ./fossa
+
```
liftM, line 17: I'd just call this `make install`, not `make fossa` or `make install-local`.

Makefile
```diff
@@ -0,0 +1,42 @@
...
+install-local: fossa
+
+# Run analysis on the sandbox directory
+sandbox: fossa
+	./fossa analyze --output --debug --record ./sandbox > fossa.json
+
+# Run the sandbox with replay mode on
+replay: fossa fossa.debug.json
+	./fossa analyze --output --debug --replay fossa.debug.json ./sandbox > fossa.replay.json
```
liftM, line 26: This is interesting.

I'm not sure this is _useful_ in general. You can't replay a customer debug log unless you use the same folder name, and there's no point in replaying your own `./sandbox` run because you could just re-run `make build && make sandbox` since you'd have to have the project files locally to generate the record in the first place.

Unless I'm missing something, this doesn't seem useful. I'd remove it.

Makefile
```diff
@@ -0,0 +1,42 @@
...
+
+# Format everything (don't use this, use your IDE to format)
+# `@command` does not echo the command before running
+fmt:
+	@fourmolu --mode inplace ${FMT_OPTS} $(shell find ${FIND_OPTS})
+
+# Confirm everything is formatted without changing anything
+fmt-ck:
+	@fourmolu --mode check ${FMT_OPTS} $(shell find ${FIND_OPTS})
+	@echo "No formatting errors found"
```
liftM, line 36: I'd combine this into `make lint`, so there's a single command that you run to make sure everything is okay before a merge.

Makefile
```diff
@@ -0,0 +1,42 @@
...
+
+# Lint everything (This parses wrong sometimes, use your editor instead)
```
liftM, line 38: > This parses wrong sometimes

That's alarming. When does this happen?

liftM (changes requested):
I like it, but I would remove a couple of the less general targets.

Makefile
scruffystuffs: When extensions are not activated.  For now it seems to be OK, and it helps that we don't use default extensions in cabal unless they're no-brainers.  But it can happen, and HLS prevents it.

Makefile
scruffystuffs: I purposely left these separate so that CI can run the two processes independently, that way if both fail, you get all the info you need to fix it.

Makefile
liftM: We can configure hlint extensions via flags or `.hlint.yaml` ([doc](https://github.com/ndmitchell/hlint#language-extensions)). Let's do that instead of leaving this spooky comment in.

Makefile
scruffystuffs: We can combine them into a `check` target, but keeping them separate is intentional.

liftM: `make check` SGTM. In that case, I would still rename `fmt-ck` to `check-fmt` (or something less terse).

Makefile
scruffystuffs: It's intentionally given the same name as the file it produces.  `make` will not rebuild it if it already exists, which saves some time.

.gitignore
scruffystuffs: Just an idea.  I liked the idea of sandbox because of the pre-existing connotation of it being isolated.

.github/pull_request_template.md
```diff
@@ -22,12 +22,12 @@ _List any referenced GitHub issues. If PR references tickets in other systems (e
 
 _Make sure to use keywords to link this PR to GitHub issues ([GitHub Docs](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))._
 
-_Example:_ Closes https://github.com/example/repo/issues/123.
+_Example:_ Closes repo/issues#123.
```
liftM, line 25: Will this shortened form work for repos across the `fossas` org?

Makefile
scruffystuffs: Yeah, I forgot about the working directory part when writing this.

.gitignore
liftM: Given that we already have `make fossa`, I don't encoding the `sandbox` convention adds much value. I'd personally remove references to `sandbox`, but if you want to start this as a convention and drive any tooling around it, I'm okay with that too.

.github/pull_request_template.md
scruffystuffs: Yep.  Edit this comment and see how fossas/team-analysis#1 is rendered.  I've been doing it on PR's for a few months now.  It actually works across GH, not just within an org.

scruffystuffs: I think it should be `org/repo#number` in the example text, though

.gitignore
scruffystuffs: We need somewhere that git will ignore, it's really annoying to have it always tell me there are untracked files.  I'll leave it, but I'll also make it more modular so we can decide on some other name if we want.  And this only affects us, so I'm not worried about committing to anything.

scruffystuffs: If I develop other tooling it will be modular with that name as well.

liftM: > We need somewhere that git will ignore

I don't think picking an arbitrary folder name is the right solution for this. Developers should configure their personal preferences. You can set up [personal repository exclusion settings](https://stackoverflow.com/questions/25912441/can-i-add-a-gitignore-file-just-for-me-that-overrides-the-project-gitignore), although I use a [personal global gitignore](https://stackoverflow.com/questions/5724455/can-i-make-a-user-specific-gitignore-file).

scruffystuffs: I also avoid doing either of those exclusion methods, since it can be difficult to track down why something is excluded if it's not right there in the project.  Many people don't know about either of those features well enough to use them without googling it (though they are pretty simple once you look) and providing an obvious place to put test projects seems a lot simpler to me.  

Is picking an arbitrary name really so bad?  I don't see how this would have such a negative effect.  We could just remove it if no one else uses it.

liftM: > Is picking an arbitrary name really so bad? I don't see how this would have such a negative effect. We could just remove it if no one else uses it.

I just think that this behavior ("hey there's a folder named `sandbox` that for some reason is ignored") is surprising, and that it provides little value given that developers can really easily set up personal gitignores. The articles I linked were the top 2 results for "personal gitignore".

I'm not going to fight hard to block this, but I don't think it's the right tool for this problem.

liftM (approved):

Makefile
liftM: Before you merge - did you add these flags @scruffystuffs? I'm not seeing an `.hlint.yaml` either.

scruffystuffs: That's already in the repo from an earlier PR.  See #258

scruffystuffs:
@spatten @cnr I'm going to merge this, we can remove if it's not the right approach.

PR #260, opened 2021-06-16
Url dependencies refactor

anuccio1:
# Overview

As part of the addition of manually added dependencies in the UI, we are planning on allowing users to add dependencies from an `archive`, a `url`, a `locator`, or `user-defined`.

As part of the flow for a `url` dep, we are also allowing the user to specify a name, version, a description, and a homepage for the dependency:

Screen Shot 2021-06-15 at 8 11 35 PM
view of this change. Describe the intent of this change, and how it implements that intent._

Because a user can specify any combination of the metadata they please, the plan is to add a new fetcher so that we can do the following:

- scope to an organization like we do for archives (i.e. prepend `<org_id>/` in the package
- stop [downloading the data from the url](https://github.com/fossas/FOSSA/blob/1afc001e80d4c3c1bddf220b4e040c453c8afdcf/modules/fetchers/url.ts#L54) inside the `resolve` step.

To accommodate these changes, the following PR makes these changes to how the CLI treats `url` locators:

- Creates a new `remote-dependencies` section in the `fossa-deps` file
- move `url` locators into the `AdditionalDependencyData` section of the upload payload to allow FOSSA backend to handle prepending the organization data, as well as populating the metadata.
- modify the `remote-dependencies` and `custom-dependencies` to include optional fields in the `metadata` section

## Acceptance criteria

- url dependencies can now be properly specified in this new `remote-dependencies` section of the `fossa-deps` file

## Testing plan

- automated tests
- built and tested on my own local `fossa-deps` file

## Risks
n/a

## Checklist

- [] I updated `Changelog.md`. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -210,7 +210,7 @@ Supported dependency types:
 - `cocoapods` - Swift and Objective-C dependencies found at [Cocoapods.org](https://cocoapods.org/).
 - `url` - The URL type allows you to specify only the download location of an archive (e.g.: `.zip`, .`tar.gz`, etc.) in the `name` field and the FOSSA backend will attempt to download and scan it. Example for a github source dependency `https://github.com/fossas/spectrometer/archive/refs/tags/v2.7.2.tar.gz`. The `version` field will be silently ignored for `url` type dependencies.
 
-### User-defined dependencies
+### Custom dependencies
 
 FOSSA supports users that have dependencies that can't be automatically discovered or identified, by offering the ability to define new dependencies.
 
@@ -289,6 +289,66 @@ referenced-dependencies:
 The `fossa-deps` scanner also requires at least one valid dependency if the file exists.  This prevents the file from being created with the wrong array names and us silently ignoring them.
 
 If you see an error message that isn't clear, file an issue in this repository!  Clear error messages are a priority for us, and we want to know where we're lacking.
+
+### License scanning local dependencies
+
+Fossa offers the ability to license scan your code directly. This is used primarily if a package manager is not yet supported or if you are vendoring dependencies. Using the license scanning feature will allow you to capture the licenses for dependencies that may otherwise be missed from normal fossa analysis that relies on package manager information.
+
+In order to specify a file path, modify your `fossa-deps.yml` file and add a `vendored-dependencies` section like the following:
+
+```yml
+# Example full `fossa.deps.yml` file.
+referenced-dependencies:
+- type: gem
+  name: rubyXL
+  version: 3.4.16
+
+vendored-dependencies:
+- name: Django
+  path: vendor/Django-3.4.16.zip # path can be either a file or a folder.
+  version: 3.4.16 # revision will be set to the MD5 hash of the filepath if left unspecified.
+```
+
+> Note: License scanning currently operates by uploading the files at the specified path to a secure S3 bucket. All files that do not contain licenses are then removed after 2 weeks.
+
+We also support json-formatted dependencies:
+
+```json
+{
+  "referenced-dependencies": [
+    {
+      "type": "gem",
+      "name": "iron"
+    }, {
+      "type": "pypi",
+      "name": "Django",
+      "version": "2.1.7"
+    }
+  ],
+  "custom-dependencies": [
+    {
+      "name": "foo",
+      "version": "1.2.3",
+      "license": "MIT"
+    }, {
+      "name": "foo-wrapper",
+      "version": "1.0.2",
+      "license": "MIT or Apache-2.0",
+      "description": "Provides a help wrapper for foo-related tasks",
+      "url": "https://foo-project.org/homepage"
+    }
+  ],
+  "vendored-dependencies": [
+    {
+      "name": "lodash",
+      "path": "lodash-4.17.21"
+    }, {
+      "name": "winston",
+      "path": "vendor/winston.tar.gz",
+      "version": "5.0.0-alpha"
+    }
+  ]
+}
 
 ## `fossa test`
 
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -208,6 +208,7 @@ Supported dependency types:
 - `nuget` - .NET dependencies found at [NuGet.org](https://www.nuget.org/).
 - `pypi` - Python dependencies that are typically found at [Pypi.org](https://pypi.org/).
 - `cocoapods` - Swift and Objective-C dependencies found at [Cocoapods.org](https://cocoapods.org/).
+- `url` - The URL type allows you to specify only the download location of an archive (e.g.: `.zip`, .`tar.gz`, etc.) in the `name` field and the FOSSA backend will attempt to download and scan it. Example for a github source dependency `https://github.com/fossas/spectrometer/archive/refs/tags/v2.7.2.tar.gz`. The `version` field will be silently ignored for `url` type dependencies.
 
 ### User-defined dependencies
 
@@ -222,12 +223,13 @@ custom-dependencies:
 - name: foo
   version: 1.2.3
   license: "MIT or Apache-2.0"
-# You can also provide a description and/or url
+# You can also provide a description and/or homepage
 - name: foo-wrapper
   version: 1.2.3
   license: MIT
-  url: https://www.foowrapper.com/about
-  description: Provides foo and a helpful interface around foo-like tasks.
+  metadata:
+    homepage: https://www.foowrapper.com/about
+    description: Provides foo and a helpful interface around foo-like tasks.
 ```
 
 ### Remote dependencies
@@ -243,11 +245,13 @@ remote-dependencies:
   version: 1.2.3
   url: https://www.fooarchive.tar.gz
 
-# You can also provide a description
+# You can also provide a description and/or homepage
 - name: foo-wrapper
   version: 1.2.3
   url: https://www.foowrapper.tar.gz
-  description: Provides foo and a helpful interface around foo-like tasks.
+  metadata:
+    description: Provides foo and a helpful interface around foo-like tasks.
+    homepage: https://www.foowrapper-home.com
 ```
 
 ### Errors in the `fossa-deps` file
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -234,9 +234,13 @@ custom-dependencies:
 
 ### Remote dependencies
 
-FOSSA also supports dependencies that can't be automatically discovered or identified, but the user has a remote URL that points to an archive where the dependency is hosted.
+FOSSA also supports dependencies that can't be automatically discovered or identified, but where the user has a URL where FOSSA can download the source code of the dependency.
 
-To do this, you must supply the name, version, and url of the dependency. The FOSSA backend will attempt to download and scan the archive hosted at this remote location. Example for a github source dependency: `https://github.com/fossas/spectrometer/archive/refs/tags/v2.7.2.tar.gz`. You may also supply a description that will be included inside your report, however this is optional.
+To specify a remote dependency, you must provide the name, version, and download URL of the dependency. The FOSSA backend will attempt to download and scan any source code contained in an archive hosted at this URL.
+
+For example, for a dependency released on a GitHub release, your URL might look like: `https://github.com/fossas/spectrometer/archive/refs/tags/v2.7.2.tar.gz`.
+
+You can also optionally add metadata fields ("description" and "homepage") to populate these fields in the FOSSA web UI (these fields can be displayed when generating reports).
 
 ```yaml
 remote-dependencies:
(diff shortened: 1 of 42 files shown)


docs/userguide.md
```diff
@@ -231,6 +230,26 @@ custom-dependencies:
   description: Provides foo and a helpful interface around foo-like tasks.
 ```
 
+### Remote dependencies
+
+FOSSA also supports dependencies that can't be automatically discovered or identified, but the user has a remote URL that points to an archive where the dependency is hosted.
+
+To do this, you must supply the name, version, and url of the dependency. The FOSSA backend will attempt to download and scan the archive hosted at this remote location. Example for a github source dependency: `https://github.com/fossas/spectrometer/archive/refs/tags/v2.7.2.tar.gz`. You may also supply a description that will be included inside your report, however this is optional.
```
zlav, line 237: Is version required by the backend? I think allowing users to provide the version is great to reduce scan time, but if the contents at the URL change we won't be picking up any of those changes. 

Due to the increase in scan time to hash the downloaded file, I think there's a chance this is a limitation we're ok with, but we should make this very clear to users so that they don't try to use this for that purpose.

Example: Company keeps the latest version at a URL that doesn't change (https://fossa.com/fossa-release/latest/fossa.tar) but they upload the latest version of the dependency to that URL every time a new version is released.

cnr:
Have concerns about the data model and satisfying the yarn usecase -- linking the slack thread for future folks viewing this PR: https://teamfossa.slack.com/archives/G01DUC2JW3B/p1623882456025500?thread_ts=1623880509.023300&cid=G01DUC2JW3B

liftM (dismissed):
I think it's confusing that both custom dependencies _and_ remote dependencies have a `url` field. I think users will mix them up.

Instead, I would propose:

```yml
custom-dependencies:
- name: foo
  version: 1.2.3
  license: MIT
  metadata:
    homepage: www.example.com/foo # Used to be `url`
    description: Does a Foo

remote-dependencies:
- name: bar
  version: 4.5.6
  url: www.example.com/bar/download/bar.tar.gz
  metadata:
    description: Does a Bar
```

----

Re: @cnr's concerns around data model - I think @anuccio1 is right in identifying that URLs identifying _global_ dependencies (like in Yarn) are semantically different than URLs pointing to _user-specific_ dependencies (which is more like a shorthand for "archive upload the code at this URL").

I think the underlying semantic issue here is that we don't distinguish between _global_ dependencies (dependencies that would be shared across users) from _user_ dependencies (dependencies that only one user will ever use because it's custom to them), and we need to make that distinction in the backend. (In the meantime, maybe we can hack it by separating the global-URL vs. user-URL fetcher types, much like how we've hacked the `custom` and `archive` fetchers.)

scruffystuffs:
I think that part of the problem for the yarn use-case is that we picked up the `url` fetcher for it because there was no other option at the time, and that option happened to match what we needed.  However, the semantic idea of url-based dependencies for the yarn use case is not necessarily the same as the original `url` fetcher, even though it could be expressed in some of the same terms ("Some code at some url").  

The original `url` implementation was not the complete understanding of the archive url locator/fetcher, and the yarn use case is not necessarily the same use-case now that the archive url use-case has evolved to a more semantically complete state.

anuccio1:
> I think it's confusing that both custom dependencies _and_ remote dependencies have a `url` field. I think users will mix them up.
> 
> Instead, I would propose:
> 
> ```yaml
> custom-dependencies:
> - name: foo
>   version: 1.2.3
>   license: MIT
>   metadata:
>     homepage: www.example.com/foo # Used to be `url`
>     description: Does a Foo
> 
> remote-dependencies:
> - name: bar
>   version: 4.5.6
>   url: www.example.com/bar/download/bar.tar.gz
>   metadata:
>     description: Does a Bar
> ```
> 
> Re: @cnr's concerns around data model - I think @anuccio1 is right in identifying that URLs identifying _global_ dependencies (like in Yarn) are semantically different than URLs pointing to _user-specific_ dependencies (which is more like a shorthand for "archive upload the code at this URL").
> 
> I think the underlying semantic issue here is that we don't distinguish between _global_ dependencies (dependencies that would be shared across users) from _user_ dependencies (dependencies that only one user will ever use because it's custom to them), and we need to make that distinction in the backend. (In the meantime, maybe we can hack it by separating the global-URL vs. user-URL fetcher types, much like how we've hacked the `custom` and `archive` fetchers.)

I'm moving forward with making a completely separate fetcher for the two use cases. One will be the existing `url` for global deps, the other will be `url-private`. In addition, I will change the `fossa-deps` file to include this `metadata` field

docs/userguide.md
anuccio1: It will be required, yes. I think what you're describing will be an edge case that we're ok with

src/App/Fossa/ManualDeps.hs
```diff
@@ -147,8 +164,27 @@ data CustomDependency = CustomDependency
   { customName :: Text
   , customVersion :: Text
   , customLicense :: Text
-  , customDescription :: Maybe Text
-  , customUrl :: Maybe Text
+  , customMetadata :: Maybe CustomDependencyMetadata
+  }
+  deriving (Eq, Ord, Show)
+
+data CustomDependencyMetadata = CustomDependencyMetadata
```
scruffystuffs, line 171: [nit]: Should we combine the types of metadata here?  Are `CustomDependencyMetadata` and `RemoteDependencyMetadata` going to diverge?

scruffystuffs: Also, should `VendoredDependency` also support description and homepage?

src/App/Fossa/ManualDeps.hs
```diff
@@ -180,10 +217,27 @@ instance FromJSON CustomDependency where
     CustomDependency <$> obj .: "name"
       <*> (unTextLike <$> obj .: "version")
       <*> obj .: "license"
-      <*> obj .:? "description"
-      <*> obj .:? "url"
+      <*> obj .:? "metadata"
```
scruffystuffs, line 220: Now that we've moved url away, let's add it to the `forbidMembers` list.

src/App/Fossa/ManualDeps.hs
```diff
@@ -180,10 +217,27 @@ instance FromJSON CustomDependency where
...
+    CustomDependencyMetadata <$> obj .:? "description"
+      <*> obj .:? "homepage"
+
+instance FromJSON RemoteDependency where
+  parseJSON = withObject "RemoteDependency" $ \obj ->
+    RemoteDependency <$> obj .: "name"
+      <*> (unTextLike <$> obj .: "version")
+      <*> obj .: "url"
+      <*> obj .:? "metadata"
+      <* forbidMembers "remote dependencies" ["license"] obj
```
scruffystuffs, line 234: Let's add `type` to this list.

docs/userguide.md
```diff
@@ -223,12 +223,35 @@ custom-dependencies:
...
+```
+
+### Remote dependencies
+
+FOSSA also supports dependencies that can't be automatically discovered or identified, but the user has a remote URL that points to an archive where the dependency is hosted.
+
+To do this, you must supply the name, version, and url of the dependency. The FOSSA backend will attempt to download and scan the archive hosted at this remote location. Example for a github source dependency: `https://github.com/fossas/spectrometer/archive/refs/tags/v2.7.2.tar.gz`. You may also supply a description that will be included inside your report, however this is optional.
+
+```yaml
+remote-dependencies:
```
scruffystuffs, line 242: Let's also add this section to the JSON example

src/Srclib/Types.hs
```diff
@@ -43,16 +44,27 @@ data SourceUnitDependency = SourceUnitDependency
   }
   deriving (Eq, Ord, Show)
 
-newtype AdditionalDepData = AdditionalDepData
-  {userDefinedDeps :: [SourceUserDefDep]}
+data AdditionalDepData = AdditionalDepData
+  { userDefinedDeps :: Maybe [SourceUserDefDep]
+  , remoteDeps :: Maybe [SourceRemoteDep]
```
scruffystuffs, line 49: Is there a reason these are `Maybe`'s?  Is the empty list case different than the null case?  If not, these should just be lists.

test/App/Fossa/testdata/the-works.yml
```diff
@@ -19,5 +19,17 @@ custom-dependencies:
...
   version: 3.2.1
   license: GPL-3.0
-  description: description for full
-  url: we don't validate url's
\ No newline at end of file
+  metadata:
+    description: description for full
+    homepage: we don't validate homepages
+
+remote-dependencies:
```
scruffystuffs, line 26: You'll need to add this to `the-works.json` as well.

scruffystuffs (changes requested):
Few minor questions and changes, but this is structurally sound.  I agree with the new UX as well.

zlav:
@anuccio1 after adding another fetcher `url-private` how do you expect that to translate to the `fossa-deps` file format? Does this look like separate types or do we proceed with the `remote-dependencies` section as well?

docs/userguide.md
```diff
@@ -223,12 +223,35 @@ custom-dependencies:
 - name: foo
   version: 1.2.3
   license: "MIT or Apache-2.0"
-# You can also provide a description and/or url
+# You can also provide a description and/or homepage
```
liftM, line 226:
```suggestion
# You can also provide a description and/or homepage. These values populate metadata fields in reports in the FOSSA web UI.
```

docs/userguide.md
```diff
@@ -223,12 +223,35 @@ custom-dependencies:
...
-  url: https://www.foowrapper.com/about
-  description: Provides foo and a helpful interface around foo-like tasks.
+  metadata:
+    homepage: https://www.foowrapper.com/about
+    description: Provides foo and a helpful interface around foo-like tasks.
+```
+
+### Remote dependencies
+
+FOSSA also supports dependencies that can't be automatically discovered or identified, but the user has a remote URL that points to an archive where the dependency is hosted.
```
liftM, line 237:
```suggestion
FOSSA also supports dependencies that can't be automatically discovered or identified, but where the user has a URL where FOSSA can download the source code of the dependency.
```

docs/userguide.md
liftM: (Sorry to necro-post this thread, but GitHub won't let me start a new comment on this line for some reason.)

```suggestion
To specify a remote dependency, you must provide the name, version, and download URL of the dependency. The FOSSA backend will attempt to download and scan any source code contained in an archive hosted at this URL.

For example, for a dependency released on a GitHub release, your URL might look like: `https://github.com/fossas/spectrometer/archive/refs/tags/v2.7.2.tar.gz`.

You can also optionally add metadata fields ("description" and "homepage") to populate these fields in the FOSSA web UI (these fields can be displayed when generating reports).
```

liftM (commented):
UX LGTM. I would block this merge on an implementation LGTM.

src/App/Fossa/ManualDeps.hs
anuccio1: It looks like `VendoredDependency`'s are just converted into locators. I think supporting metadata for them would involve a larger scale refactor so I'm gonna leave that alone for now

src/App/Fossa/ManualDeps.hs
anuccio1: great suggestion

anuccio1:
> @anuccio1 after adding another fetcher `url-private` how do you expect that to translate to the `fossa-deps` file format? Does this look like separate types or do we proceed with the `remote-dependencies` section as well?

`url-private` will be abstracted away from the CLI code base altogether. We'll have the new `remote-dependencies` section, and Core will receive a payload like this:

```
[
      {
          "Name": ".",
          "Type": "commonjspackage",
          "Manifest": ".",
          "Build": {
              "Artifact": "default",
              "Context": null,
              "Succeeded": true,
              "Imports": [
                  "mvn+com.github.fge:json-schema-core$1.0.2"
              ],
              "Dependencies": [
                  {
                      "locator": "mvn+com.github.fge:json-schema-core$1.0.2"
                  }
              ]
          },
          "AdditionalDependencyData": {
              "UserDefinedDependencies": [{
                  "Name": "usertest",
                  "Version": "1.2.3",
                  "License": "MIT OR APACHE-2.0",
                  "Description": "This is a desc",
                  "Homepage": "www.foo.com2"
              }],
              "RemoteDependencies": [{
                  "Name": "al-url",
                  "Version": "1.2.3",
                  "Description": "This is a desc of al-url",
                  "Url": "https://repo1.maven.org/maven2/aero/m-click/mcpdf/0.2.3/mcpdf-0.2.3-jar-with-dependencies.jar"
              }]
          }
      }
  ]
```

And construct the locators inside the build, similar to how we're doing this with user-defined dependencies

src/Srclib/Types.hs
anuccio1: They are `Maybe`'s, because they are optional, and can be null.

> Is the empty list case different than the null case?

They would produce the same results in the Build, however if there are no `Remote` or `UserDefined` deps, I would rather have `AdditionalDeps` leave that field out completely, rather than having it point at an empty list. This makes more sense to me as we will only add fields that the user has provided explicitly, and additionally the type will more closely represent reality

scruffystuffs: Good enough for me.

scruffystuffs (approved):
LGTM, just need a changelog entry and we should probably merge the backend work first.

PR #262, opened 2021-06-22
Fix scanning with no reference dependencies

zlav:
# Overview

If custom or archive dependencies are found, but no reference dependencies are found, the backend build will fail. This is due to the fact that the upload has an upload with the following format that has a `null` `Build` field. This causes a backend failure and instead should be uploaded as `[]` if we need this build to pass. I am still looking into if there is a better value to return.
```
    {
      "Build": null,
      "Manifest": "/Users/zachlavallee/Programming/Fossa-Tools/spectrometer/user-deps-test/",
      "Name": "/Users/zachlavallee/Programming/Fossa-Tools/spectrometer/user-deps-test/",
      "AdditionalDependencyData": null,
      "Type": "user-specific-yaml"
    },
```

## Acceptance criteria

Scans that do not have `referenced-dependencies` succeed.

## Testing plan

I validated that these changes work by building a new binary and testing on a few different variations of `fossa-deps.yaml` file.
- Only archive
- Only ref
- Only custom
- Arc and custom
- Arc and ref
- Ref and custom

## Risks

None, this PR only fixes something that's broken.

## Checklist

### DIFF
--- a/src/App/Fossa/ArchiveUploader.hs
+++ b/src/App/Fossa/ArchiveUploader.hs
@@ -26,7 +26,7 @@ import Data.Text (Text)
 import Data.Text qualified as T
 import Fossa.API.Types
 import Path hiding ((</>))
-import Srclib.Types (Locator (..), SourceUnit (..), SourceUnitBuild (..), SourceUnitDependency (SourceUnitDependency))
+import Srclib.Types (Locator (..))
 import System.FilePath.Posix
 
 data VendoredDependency = VendoredDependency
--- a/src/App/Fossa/ArchiveUploader.hs
+++ b/src/App/Fossa/ArchiveUploader.hs
@@ -62,7 +62,7 @@ compressAndUpload apiOpts arcDir tmpDir dependency = do
 
 -- archiveUploadSourceUnit receives a list of vendored dependencies, a root path, and API settings.
 -- Using this information, it uploads each vendored dependency and queues a build for the dependency.
-archiveUploadSourceUnit :: (Has Diag.Diagnostics sig m, Has (Lift IO) sig m) => Path Abs Dir -> ApiOpts -> [VendoredDependency] -> m SourceUnit
+archiveUploadSourceUnit :: (Has Diag.Diagnostics sig m, Has (Lift IO) sig m) => Path Abs Dir -> ApiOpts -> [VendoredDependency] -> m [Locator]
 archiveUploadSourceUnit baseDir apiOpts vendoredDeps = do
   archives <- withSystemTempDir "fossa-temp" (uploadArchives apiOpts vendoredDeps baseDir)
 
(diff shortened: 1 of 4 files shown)


src/App/Fossa/ArchiveUploader.hs
```diff
@@ -78,47 +78,22 @@ archiveUploadSourceUnit baseDir apiOpts vendoredDeps = do
...
 
-  pure $ Just $ archivesToSourceUnit archivesWithOrganization
-
+  pure $ arcToLocator <$> archivesWithOrganization
 
 -- archiveNoUploadSourceUnit exists for when users run `fossa analyze -o` and do not upload their source units.
-archiveNoUploadSourceUnit :: [VendoredDependency] -> Maybe SourceUnit
-archiveNoUploadSourceUnit deps = Just . archivesToSourceUnit $ map forceVendoredToArchive deps
+archiveNoUploadSourceUnit :: [VendoredDependency] -> [Locator]
+archiveNoUploadSourceUnit deps = arcToLocator <$> map forceVendoredToArchive deps
```
scruffystuffs, line 85: [nit]: This is essentially a `map f . map g a`, which is equivalent to `map (f . g) a`
```suggestion
archiveNoUploadSourceUnit deps = map (arcToLocator . forceVendoredToArchive) deps
```

src/App/Fossa/YamlDeps.hs
```diff
@@ -37,19 +36,16 @@ import Path
...
+    Nothing -> pure Nothing
     Just depsFile -> do
       yamldeps <- context "Reading fossa-deps file" $ readContentsYaml depsFile
-      sourceUnit <- context "Converting fossa-deps to partial API payload" $ Just <$> toSourceUnit root yamldeps
-      archiveSrcUnit <- case maybeApiOpts of
-        Nothing -> pure $ archiveNoUploadSourceUnit (vendoredDependencies yamldeps)
-        Just apiOpts -> archiveUploadSourceUnit root apiOpts (vendoredDependencies yamldeps)
-      pure $ catMaybes [sourceUnit, archiveSrcUnit]
+      -- If the file exists and we have no dependencies to report, that's a failure.
+      when (hasNoDeps yamldeps) $ fatalText "No dependencies found in fossa-deps file"
```
scruffystuffs, line 47: [Important]: This line should be a part of `toSourceUnit`.  That constructor should take the sole responsibility of deciding whether we have enough information to construct a valid source unit.

src/App/Fossa/YamlDeps.hs
```diff
@@ -78,32 +78,30 @@ toSourceUnit root yamldeps@YamlDependencies{..} = do
       , additionalData = additional
       }
 
-toBuildData :: NE.NonEmpty ReferencedDependency -> SourceUnitBuild
-toBuildData deps =
+toBuildData :: NE.NonEmpty Locator -> SourceUnitBuild
```
scruffystuffs, line 81: [no action]: For now, I'm okay with this.  Long-term, I prefer to let internal models exist as raw/unprocessed as possible until they are needed, and I think `Locator` is not the best thing to use in this position.

For example, I would prefer passing back some structure that could be turned into a locator, e.g. `Archive` and `arcToLocator`.  This ensures that we don't limit ourselves to our output representation, and instead retain full flexibility of our own internal data model.

I think in this case, there's no real concern, but it is a pattern I'd like to see us do more of.

scruffystuffs (changes requested):
One minor change necessary.

src/App/Fossa/ArchiveUploader.hs
zlav: Why is this superior?

src/App/Fossa/YamlDeps.hs
zlav: Yea I agree with you, however in this case returning something other than Locator requires another layer of transformation which feels overkill.

src/App/Fossa/YamlDeps.hs
zlav: I don't know if I agree with this. `toSourceUnit` is converting a list of locators into a sourceUnit. It feels wrong to allow that constructor to exit 1 the entire program. IMO the caller should be the one to decide whether the output is acceptable or not.

src/App/Fossa/ArchiveUploader.hs
scruffystuffs: `map f . map g` iterates over the list twice (in theory.  In practice, GHC might do some shenanigans to fix this, but we shouldn't rely on that).  `map (f .g)` iterates once, but does more transformation per iteration step.

Additionally, `map f $ map g a` is clearer than `f <$> map g a` even though they're identical, and `map (f . g) a` is an improvement on both.

zlav: Ah, yea I see that now. Thanks for pointing that out. Didn't realize for some reason that I was iterating twice.

src/App/Fossa/YamlDeps.hs
scruffystuffs: It doesn't fail the program, it just fails the `Diagnostic` carrier.  In this case, we intentionally let that diagnostic bubble up all the way to the main `analyze` handler.  `toSourceUnit` is a smart constructor, and it's only responsibility is to transform some inputs into a source unit, which is not guaranteed to succeed.  In this codebase, `fatal`, and its `Text`-specialized version `fatalText` will "throw a diagnostic error".  Which can be caught or left to flow upwards.

src/App/Fossa/YamlDeps.hs
scruffystuffs: Like I said, for now this is perfectly fine.  I think we may revisit this when we deprecate srclib.

src/App/Fossa/YamlDeps.hs
zlav: I'm still not sure I agree with the idea that the constructor should be allowed to fail for the reason we are failing it, but I'm comfortable changing this to preserve the previous behavior.

src/App/Fossa/YamlDeps.hs
```diff
@@ -37,19 +36,15 @@ import Path
...
-    -- If the file exists and we have no SourceUnit to report, that's a failure
+    Nothing -> pure Nothing
     Just depsFile -> do
       yamldeps <- context "Reading fossa-deps file" $ readContentsYaml depsFile
-      sourceUnit <- context "Converting fossa-deps to partial API payload" $ Just <$> toSourceUnit root yamldeps
-      archiveSrcUnit <- case maybeApiOpts of
-        Nothing -> pure $ archiveNoUploadSourceUnit (vendoredDependencies yamldeps)
-        Just apiOpts -> archiveUploadSourceUnit root apiOpts (vendoredDependencies yamldeps)
-      pure $ catMaybes [sourceUnit, archiveSrcUnit]
+      -- If the file exists and we have no dependencies to report, that's a failure.
```
scruffystuffs, line 46: [nit]: This should probably move to where the `when (hasNodeps deps) $ fatalText` call is.

scruffystuffs (approved):
Looks good, should probably move the comment before merging.

PR #263, opened 2021-06-23
Fix logstdout

zlav:
# Overview

Log stdout is not currently writing to output and I have traced the cause down to us not running `flushConcurrentOutput` after all `logStdout`s have been run. Adding `withConcurrentOutput` to the start of main ensures that `flush` gets called at the very end of the function.

I tested adding `withConcurrentOutput` into `logStdout` but for some reason, this caused the program to hang and I'm not entirely sure why.

## Acceptance criteria

This PR is trying to 
## Testing plan

I ran `fossa report attribution --json` and saw that with my changes, the JSON output appears.

## Risks

There shouldn't be any risks here since we are fixing something that is broken.

## Checklist

## Notes
I am going to create a followup ticket for this because we really should be testing this in some sort of integration test.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fix an issue with logStdout ([#263](https://github.com/fossas/spectrometer/pull/263))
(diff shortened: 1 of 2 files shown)


scruffystuffs:
closed in favor of #265

PR #264, opened 2021-06-23
Add support for short flags

zlav:
# Overview

Users upgrading from CLI v1 have found compatibility issues with their command when using short flags. Further overview about this ticket is located in the github issue.

## Acceptance criteria

Allow short flags that were supported by CLI v1.

## Testing plan

Build the CLI and validate that the short flags work.

## Risks

User's commands become less clear as they choose to use the short flags. If we were starting from scratch I may not support adding the short flags, but due to the need for making upgrades seamless, I believe we need to. For this reason, I am not going to omit documentation in the user guide for short flags. They show up when running `-h` on the command line.

## References

Closes https://github.com/fossas/team-analysis/issues/596.

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # Spectrometer Changelog
 
+## Unreleased
+- Adds support for short flags to mirror CLI v1 commands ([#264](https://github.com/fossas/spectrometer/pull/264))
+
 ## v2.9.2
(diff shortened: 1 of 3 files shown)


scruffystuffs (approved):
If we're agreed on the approach of implementing short flags, then this LGTM

PR #265, opened 2021-06-23
`logStdout` should always flush output by the end of the logging effect.

scruffystuffs:
Different approach to #263 

## Checklist


zlav:
Tested that this works. I like this solution better than mine so I'm going to close my PR. Waiting to approve until we chat about adding stdout back to the logger effect.

zlav (approved):
Awesome, much more solid fix.

PR #275, opened 2021-07-01
Use `Diag.logWithExit_` to prevent exiting 0 on failure.

scruffystuffs:
# Overview

During `fossa container analyze`, when syft fails, we exit with 0.  This is because our raw diagnostic handler does not control exiting values.  Using `Diag.logWithExit_`, we can safely handle errors which terminate the program, and automatically return non-zero exit codes.

This PR migrates all eligible entry points to using `Diag.logWithExit_`.

## Acceptance criteria

* Every entry point with the following pattern uses `Diag.logWithExit_`, instead of handling manually:
  * Returns `IO ()`
  * Reports error to stderr using default format/layout.
  * Diagnostic success/failure directly translates to program success/failure
  * A pre-existing logger is in use.

## Testing plan

Test all modified commands in failure mode, confirm non-zero exit code.

## References

Reported here http://fossa.zendesk.com/agent/tickets/2814
Closes fossas/team-analysis#626

## Checklist

### DIFF
--- a/src/App/Fossa/Container/Analyze.hs
+++ b/src/App/Fossa/Container/Analyze.hs
@@ -7,14 +7,10 @@ import App.Fossa.Analyze (ScanDestination (..))
 import App.Fossa.Container (ImageText (..), extractRevision, runSyft, toContainerScan)
 import App.Fossa.FossaAPIV1 (UploadResponse (uploadError, uploadLocator), uploadContainerScan)
 import App.Types (OverrideProject (..), ProjectRevision (..))
-import Control.Carrier.Diagnostics (
-  Diagnostics,
-  renderFailureBundle,
-  runDiagnostics,
- )
+import Control.Carrier.Diagnostics (Diagnostics, logWithExit_)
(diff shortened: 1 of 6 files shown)


src/App/Fossa/Container/Analyze.hs
```diff
@@ -23,10 +19,7 @@ import Srclib.Types (parseLocator)
 
 analyzeMain :: ScanDestination -> Severity -> OverrideProject -> ImageText -> IO ()
 analyzeMain scanDestination logSeverity override image = withDefaultLogger logSeverity $ do
-  result <- runDiagnostics $ analyze scanDestination override image
-  case result of
-    Left err -> logError (renderFailureBundle err)
```
zlav, line 28: Is this the culprit for why this PR was necessary?

scruffystuffs: Yep.

zlav (approved):
Looks great, good refactor on the other parts

PR #277, opened 2021-07-06
Update WIGGINS_TAG

enricozb:

PR #278, opened 2021-07-06
fix: logWithExit_ should exit on Just

liftM:
# Overview

Fixes a bug surfaced by https://github.com/fossas/spectrometer/commit/d4f6bef696f336d32134c1e15a873a17414e4377. See the unexpected test failure at https://github.com/fossas/spectrometer/pull/277/checks?check_run_id=3000639861.

Previously, programs that ran a successful `DiagnosticsC` action under `logWithExit_` would (incorrectly) continue evaluation. These are not the intended semantics. Instead, a successful run should exit successfully, much like how an unsuccessful run exits with failure.

This bug affects commands with steps after `logWithExit_` e.g. `fossa test` always exits non-zero even on success.

## Acceptance criteria

Actions run with `logWithExit_` should exit zero on success.

## Testing plan

Run a `fossa test` on a revision with no issues (I'm using the latest Spectrometer revision). See that it succeeds. Check that the exit code is zero.

```sh
cabal run fossa -- test --project https://github.com/fossas/spectrometer --revision 4f69104e2dcdb557a870c6621639750605261ca4 && echo OK
```

## Risks

This significantly changes the semantics of `logWithExit_`. I _believe_ these were the original intended semantics given that this is only ever used as a wrapper for commands (see `grep logWithExit_`), but other code that uses this will now abort surprisingly.

## Checklist

### DIFF
--- a/src/Control/Carrier/Diagnostics.hs
+++ b/src/Control/Carrier/Diagnostics.hs
@@ -49,8 +49,8 @@ logDiagnostic diag = do
     Right success -> pure $ Just success
 
 -- | Run a void Diagnostic effect into a logger, using the default error/warning renderers.
--- | Exits with non-zero if the result is a failure.
--- | Useful for setting up diagnostics from CLI entry points.
+-- Exits with zero if the result is a success, or non-zero if the result is a failure.
+-- Useful for setting up diagnostics from CLI entry points.
 logWithExit_ :: (Has (Lift IO) sig m, Has Logger sig m) => DiagnosticsC m () -> m ()
 logWithExit_ diag = logDiagnostic diag >>= maybe (sendIO exitFailure) (const (sendIO exitSuccess))
 
--- a/src/Control/Carrier/Diagnostics.hs
+++ b/src/Control/Carrier/Diagnostics.hs
@@ -29,7 +29,7 @@ import Control.Monad.Trans
 import Data.Monoid (Endo (..))
 import Data.Text (Text)
 import Effect.Logger
-import System.Exit (exitFailure)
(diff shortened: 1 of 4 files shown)


liftM:
@cnr what's a good way to test this?

Is there a convenient way to do a single file with a small `testMain`, where I can check that the program fails to reach a line that causes it to exit non-zero?

cnr:
@liftM because `exitSuccess` really just throws an IO Exception (`throwIO ExitSuccess`), you can test it with the hspec builtin `shouldThrow`:

```
  describe "foo" $ do
    it "should do the thing" $ do
      myAction `shouldThrow` (== ExitSuccess)
```

src/Control/Carrier/Diagnostics.hs
```diff
@@ -52,7 +52,7 @@ logDiagnostic diag = do
 -- | Exits with non-zero if the result is a failure.
 -- | Useful for setting up diagnostics from CLI entry points.
```
cnr, line 53: make sure to update this comment with the new behavior

cnr (approved):
LGTM - did a cursory skim of all of the places we use `logWithExit_` and didn't notice anything problematic. we only ever use it for program entrypoints

liftM:
Given that tests passed for https://github.com/fossas/spectrometer/pull/278/commits/6dd5045b05802d959384714585eba0747a28aa0a, and that the integration test failure is due to a bad release (we should really build the integration test CLI from source, and have a _separate_ test for the installer), I'm going to force-merge this into the mainline to get the release out.

PR #271, opened 2021-06-30
doc: Document high-level differences between v1 and v2

liftM:
# Overview

This PR adds a document describing high-level differences between CLI 1.x and 2.x.

This is an _overview_ document, not intended as a detailed migration guide. The use case is basically as an explainer for "why should I upgrade from CLI v1 to CLI v2?".

A more detailed migration guide will follow in another PR.

## Checklist

- [x] I added tests for this PR's change (or confirmed tests are not viable).
- [x] I linted and formatted (via `haskell-language-server`) any haskell files I touched in this PR.
- [x] If this PR introduced a user-visible change, I added documentation into `docs/`.
- [x] I updated `Changelog.md` if this change is externally facing. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.
- [x] I linked this PR to any referenced GitHub issues, if they exist.

PR #281, opened 2021-07-08
Configuration file reference doc

zlav:
# Overview

This PR adds docs for the fossa configuration file.

[Markdown link
](https://github.com/fossas/spectrometer/blob/docs/config-file/docs/reference/fossa_yml.md)

## Acceptance criteria

Clear docs

## Testing plan

Extensive integration testing

## References

_List any referenced GitHub issues. If PR references tickets in other systems (e.g. Zendesk), they should probably be mirrored as a GitHub Issue._

_Make sure to use keywords to link this PR to GitHub issues ([GitHub Docs](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))._

_Example:_ Closes org/repo#123.

## Checklist

- [X ] I added tests for this PR's change (or confirmed tests are not viable).
- [ X] I linted and formatted (via `haskell-language-server`) any haskell files I touched in this PR.
- [ X] If this PR introduced a user-visible change, I added documentation into `docs/`.

### DIFF
--- a/docs/reference/fossa_yml.md
+++ b/docs/reference/fossa_yml.md
@@ -130,7 +130,7 @@ Targets can be are listed in the following formats for both `only` and `exclude`
 The list of `only` targets that should be scanned. When used alongside `paths.only`, the list of targets to be scanned are unioned.
 
 ### `targets.exclude:`
-The list of `exclude` targets which should be excluded from scanning. The targets listed in the exclude section will override the targets listed in the only sections. This feature is used most effectively to remove specific targets from a directory that has targets which should be scanned. 
+The list of `exclude` targets which should be excluded from scanning. The targets listed in the exclude section will override the targets listed in the only sections. This feature is used most effectively to remove specific targets from a directory. 
 
 Example: You have a directory called `docker` which contains 3 different targets but you would like to omit the one with type `bundler`. You can do this with the following configuration:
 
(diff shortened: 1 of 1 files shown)


liftM (approved):
Fine to merge as a first cut (nothing is wrong, and this is better than what we currently have, which is nothing), but I'd like to make additional improvements.

In particular:

- I think we should explicitly "tag" fields for ease of readability. For example, which fields are required? Which fields are only valid on project creation? I think the example makes it look like all fields is required, which is definitely untrue.
- We should document which CLI flag each field is equivalent to.
- We should describe use cases for modifying each field - when would I ever _want_ to set the project ID?
- [nit] I would prefer top-level headings with nested names over nested headings e.g. `### project.id:` over `#### id`, because this improves greppability (you no longer need to scroll up to see what parent field you're in).

Just because it's a reference doesn't mean we can get away with not explaining how each field is meant to be used. Inaccessible docs are not much better than missing docs.

Still, this is fine to merge as a first cut - I can go back in and revise these (or we can pair on the next cut) if these comments don't make a lot of sense to you.

liftM:
@zlav once you're done with this, please also update the documentation from https://github.com/fossas/spectrometer/pull/271#discussion_r665660950.

PR #286, opened 2021-07-13
Monorepo integration

kitified:
# Overview

### Deprecates `fossa vps` commands.

The "VPS Classic" product is deprecated. Instead we're splitting into Monorepo and VSI support.

### Adds a new flag to `fossa analyze`

`experimental-enable-monorepo`: Accept a string indicating the type of monorepo scan to run, with the only currently valid value of `aosp`. If not provided, the analysis command proceeds as normal without performing a monorepo scan.
Any provided string value is passed along to the monorepo plugin for validation; spectrometer doesn't do any validation on this value.

## Acceptance criteria

* Leave `fossa analysis` invocations that are not run with `--experimental-enable-monorepo` untouched.
* When `--experimental-enable-monorepo` is provided, do not run normal analysis; instead run a scan with the monorepo plugin.

## Testing plan

* I ran `cabal run fossa -- analyze --project quick -e 'http://localhost:9578' --fossa-api-key $FOSSA_API_KEY ~/projects/quick` and validated that the command worked the same as it did before this change.
* I ran `cabal run fossa -- analyze --experimental-enable-monorepo aosp -e 'http://localhost:9578' --fossa-api-key $FOSSA_API_KEY ~/projects/aosp-mini/` and validated that the scan worked in monorepo mode.
* I tried a few different arguments in both modes and things worked as expected.

## Risks

This could potentially keep standard `fossa analyze` commands from running properly if my new [branching pattern match](https://github.com/fossas/spectrometer/compare/monorepo-integration?expand=1#diff-f2fe846f526d80e94301cb1315b5a660dea9fee7efe2fb1be098fad1112200a8R88) is not well formed. Other than that I don't think this is very risky.

## References

Pivotal tracker: https://www.pivotaltracker.com/story/show/178806134

<details><summary>Reproduced ticket text</summary>
Remove the concept of the `vps` subcommand in Spectrometer, bringing monorepo support into the standard `analyze` process.

* Add a deprecation warning to the `fossa vps` subcommand.
* Add `--experimental-enable-monorepo` to `fossa analyze`:
  * Accepts a string.
  * Forwards the string to wiggins; leave validation logic there.
* When `fossa analyze` is run with `experimental-enable-monorepo`, run `wiggins monorepo` instead of the standard analysis flow.

When running `wiggins monorepo`, pass along flags as previously implemented for `fossa vps analyze`. Additionally, pass along the value for `--experimental-enable-monorepo` to wiggins' `-type` argument.
</details>

## Checklist

_Going to work through these, but since they're doc related only the PR is ready for review now._

### DIFF
--- a/src/App/Fossa/Main.hs
+++ b/src/App/Fossa/Main.hs
@@ -12,6 +12,7 @@ import App.Fossa.Container.Analyze qualified as ContainerAnalyze
 import App.Fossa.Container.Test qualified as ContainerTest
 import App.Fossa.EmbeddedBinary qualified as Embed
 import App.Fossa.ListTargets (listTargetsMain)
+import App.Fossa.Monorepo
 import App.Fossa.Report qualified as Report
 import App.Fossa.Test qualified as Test
 import App.Fossa.VPS.AOSPNotice (aospNoticeMain)
@@ -84,6 +85,19 @@ appMain = do
           }
 
   case optCommand of
+    AnalyzeCommand AnalyzeOptions{analyzeOutput, analyzeBranch, analyzeMetadata, monorepoAnalysisOpts = (MonorepoAnalysisOpts (Just monorepoAnalysisType) monorepoFollowSymlinks monorepoScanFileFilters), analyzeBaseDir} -> do
+      dieOnWindows "Monorepo analysis is not supported on Windows"
+      if analyzeOutput
+        then die "Monorepo analysis does not support stdout scan destination"
+        else do
+          key <- requireKey maybeApiKey
+          let apiOpts = ApiOpts optBaseUrl key
+          let metadata = maybe analyzeMetadata (mergeFileCmdMetadata analyzeMetadata) fileConfig
+          let monorepoAnalysisOpts = MonorepoAnalysisOpts (Just monorepoAnalysisType) monorepoFollowSymlinks monorepoScanFileFilters
+          let analyzeOverride = override{overrideBranch = analyzeBranch <|> ((fileConfig >>= configRevision) >>= configBranch)}
+          basedir <- parseAbsDir analyzeBaseDir
+          monorepoMain (BaseDir basedir) monorepoAnalysisOpts logSeverity apiOpts metadata analyzeOverride
+    --
     AnalyzeCommand AnalyzeOptions{..} -> do
       -- The branch override needs to be set here rather than above to preserve
       -- the preference for command line options.
@@ -302,7 +316,7 @@ monorepoOpts =
   MonorepoAnalysisOpts
     <$> optional (strOption (long "experimental-enable-monorepo" <> metavar "MODE" <> help "scan the project in the experimental monorepo mode. Supported modes: aosp"))
     <*> switch (long "experimental-monorepo-follow-symlinks" <> help "if enabled, monorepo scans follow symbolic links. Does not protect against link loops.")
-    <*> (FilterExpressions <$> jsonOption (long "experimental-monorepo-file-filters" <> metavar "REGEXPS" <> help "JSON encoded array of regular expressions used to filter scanned paths" <> value []))
+    <*> (FilterExpressions <$> jsonOption (long "experimental-monorepo-file-filters" <> metavar "REGEXPS" <> help "JSON encoded array of RE2 regular expressions used to filter scanned paths during monorepo scans" <> value []))
 
 metadataOpts :: Parser ProjectMetadata
 metadataOpts =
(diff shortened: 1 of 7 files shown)


src/App/Fossa/Main.hs
```diff
@@ -4,14 +4,15 @@ module App.Fossa.Main (
   appMain,
 ) where
 
-import App.Fossa.Analyze (RecordMode (..), ScanDestination (..), UnpackArchives (..), JsonOutput (..), VSIAnalysisMode (..), analyzeMain)
+import App.Fossa.Analyze (JsonOutput (..), RecordMode (..), ScanDestination (..), UnpackArchives (..), VSIAnalysisMode (..), analyzeMain)
```
kitified, line 7: Auto formatting did this.

src/App/Fossa/Main.hs
```diff
@@ -84,14 +85,27 @@ appMain = do
           }
 
   case optCommand of
+    AnalyzeCommand AnalyzeOptions{analyzeOutput, analyzeBranch, analyzeMetadata, monorepoAnalysisOpts = (MonorepoAnalysisOpts (Just monorepoAnalysisType)), analyzeBaseDir} -> do
+      dieOnWindows "Monorepo analysis is not supported on Windows"
+      if analyzeOutput
+        then die "Monorepo analysis does not support stdout scan destination"
```
cnr, line 91: [nit] may want to specify the `--output` flag more directly in this message

src/App/Fossa/Main.hs
```diff
@@ -84,14 +85,27 @@ appMain = do
...
+        else do
+          key <- requireKey maybeApiKey
+          let apiOpts = ApiOpts optBaseUrl key
+          let metadata = maybe analyzeMetadata (mergeFileCmdMetadata analyzeMetadata) fileConfig
+          let monorepoAnalysisOpts = MonorepoAnalysisOpts (Just monorepoAnalysisType)
```
cnr, line 96: [nit] the pattern match above should already bind `monorepoAnalysisOpts` to this value, I think(?)

src/App/Fossa/Monorepo.hs
```diff
@@ -0,0 +1,42 @@
...
+import System.Exit (exitFailure)
+
+monorepoMain :: BaseDir -> MonorepoAnalysisOpts -> Severity -> ApiOpts -> ProjectMetadata -> OverrideProject -> IO ()
+monorepoMain basedir monoRepoAnalysisOpts logSeverity apiOpts projectMeta overrideProject = withDefaultLogger logSeverity $ do
+  result <- runDiagnostics $ withWigginsBinary $ monorepoScan basedir monoRepoAnalysisOpts logSeverity apiOpts projectMeta overrideProject
+  case result of
+    Left failure -> do
+      logStdout $ toText $ show $ renderFailureBundle failure
+      sendIO exitFailure
+    Right _ -> pure ()
```
cnr, line 27: [nit] use our `logWithExit_` helper here
```suggestion
  logWithExit_ $ withWigginsBinary $ monorepoScan basedir monoRepoAnalysisOpts logSeverity apiOpts projectMeta overrideProject
```

cnr (approved):
LGTM -- I don't see any mention in the code of, e.g., `experimental-monorepo-follow-symlinks` from the PR description. Is that intentional?

cnr:
General question: is there a suggested timeline for removing the deprecated commands?

kitified:
@cnr 

> I don't see any mention in the code of, e.g., experimental-monorepo-follow-symlinks from the PR description. Is that intentional?

Yes- I removed them from the code but forgot to update the PR. PR is now updated. Basically we decided to not go down that road at this time.

> General question: is there a suggested timeline for removing the deprecated commands?

There is! We currently have 3 customers using these commands: Ford, Ericsson, and Motorola. We're working on migrating them off of these commands, and once all three are migrated off we're going to remove the entire `fossa vps` subcommand.
We don't have an exact timeline but we expect we'll be able to do this by late August (the main blocker is transitioning Ericsson to our C/C++ strategy, which is pending us being ready to beta it to them).

src/App/Fossa/Main.hs
kitified: It doesn't seem to do so:
```
Couldn't match expected type ‘MonorepoAnalysisOpts’
              with actual type ‘AnalyzeOptions -> MonorepoAnalysisOpts’
```

And I can't access `AnalyzeOptions` because we're deconstructing it in the pattern match statement.

cnr: weird. there's syntax we can use to bind the `AnalyzeOptions` as part of the pattern match:

```hs
opts@AnalyzeOptions{ etc }
```

though I'm not sure that would make this clearer. oh well

PR #287, opened 2021-07-15
Adds elixir support, when mix is accessible

meghfossa:
# Overview

This PR adds elementary elixir support using mix. I wanted to keep this PR small and manageable, so would like to merge in first tactic, before adding in tactic for parsing `mix.lock` and `mix.exs` file for analysis.

## Acceptance criteria

- [x] CLI can infer direct and indirect dependencies when `mix` is accessible 
- [x] CLI can analyze dependency from git, or hex when `mix` is accessible 

## Testing plan

- I tested this locally against few open sources elixir project. I performed e2e test against results generated. 

To test, 

- run .`/fossa analyze projectDir --output` for example project and compare result against `mix deps.tree` and `mix deps`

## Risks

_Highlight any areas that you're unsure of, or want reviewers to pay particular attention to._

Mix supports multiple build managers - mix, rebar, rebar3, and make. Collected dependency could be discovered as its own project from rebar, rebar3 strategies. This may lead to noise when running the analyze command, and in the results. Currently, there is no way to filter out specific directory. 

## References

- https://github.com/fossas/team-analysis/issues/398
- https://github.com/fossas/team-analysis/issues/661
- https://github.com/fossas/team-analysis/issues/660

This does not close, but partially meets reqs of https://github.com/fossas/team-analysis/issues/398

## Checklist


src/Strategy/Elixir/MixTree.hs
```diff
@@ -0,0 +1,374 @@
...
+    sc = L.space (void $ some $ char ' ') empty empty
+
+-- | Consumes whitespace ` ` or tab `\t`
+whitespaceOrTab :: Parser Text
+whitespaceOrTab = takeWhileP (Just "whitespaceOrTab") isSpaceOrTab
+  where
+    isSpaceOrTab c = c == ' ' || c == '\t'
+
+-- | Parses `VerConstraint`
+parseVerConstraint :: Parser VerConstraint
```
meghfossa, line 326: I would like to abstract this out - it's very similar to poetry's version constraint expression parsing. But keeping this in for now, until poetry PR is merged. I can address this in separate PR.

zlav:
Can you elaborate on what this means?

>Collected dependency could be discovered as its own project from rebar, rebar3 strategies

Are you saying that there vendored dependencies that have a `rebar.config` file which will be picked up and scanned?

meghfossa:
> Are you saying that there vendored dependencies that have a `rebar.config` file which will be picked up and scanned?

Ah! As mix can use few build manager (make, rebar, rebar3, mix). Some dependency may, for example have rebar3 as build manager. 

Here is an example:

mix.lock file
```elixir
%{
...

  "hackney": {:hex, :hackney, "1.15.2", "07e33c794f8f8964ee86cebec1a8ed88db5070e52e904b8f12209773c1036085", [:rebar3], [{:certifi, "2.5.1", [hex: :certifi, repo: "hexpm", optional: false]}, {:idna, "6.0.0", [hex: :idna, repo: "hexpm", optional: false]}, {:metrics, "1.0.1", [hex: :metrics, repo: "hexpm", optional: false]}, {:mimerl, "~>1.1", [hex: :mimerl, repo: "hexpm", optional: false]}, {:ssl_verify_fun, "1.1.5", [hex: :ssl_verify_fun, repo: "hexpm", optional: false]}], "hexpm", "e0100f8ef7d1124222c11ad362c857d3df7cb5f4204054f9f0f4a728666591fc"},

...
}```

When compiling and getting dependencies, it will download build config for `hackney` in deps directory.

```
>> tree elxirEampleProject/deps/hackney
...
├── _build
│   └── default
│       └── lib
├── ebin
├── hex_metadata.config
├── rebar.config
├── rebar.lock
└── src
    ├── hackney.app.src
....
```

This is getting discovered as it's own project with analyze. I'm not sure how to exclude this. I did skip deps directory in my strategy's discovery, but since rebar strategies are independent they pick these up as independent projects.

**Comment by scruffystuffs** [comment]
The way to solve this is to adjust the rebar strategy to not search the `deps` folder if there's a `mix.exs` file with the string `":rebar3"`.

**Review by zlav — COMMENTED** [review]
This looks really good. I have a few comments that need to be addressed and a few nits but nothing that is too major.

File: src/Strategy/Elixir/MixTree.hs [hunk 2]
```diff
@@ -0,0 +1,374 @@
...
+mixDepCmd :: Command
+mixDepCmd =
+  Command
+    { cmdName = "mix"
+    , cmdArgs = ["deps", "--all"]
+    , cmdAllowErr = Never
+    }
+
+-- | mix deps.tree --format plain (--only applied to env atom if any)
+mixDepTreeCmd :: Maybe MixEnvAtom -> Command
```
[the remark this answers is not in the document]
meghfossa: No you are right. By default command reports all dependencies, I can update it so we only get prod dependencies.

src/Strategy/Elixir/MixTree.hs
```diff
@@ -59,41 +71,37 @@ missingDepVersionsMsg = "Some of dependencies versions were not resolved from `m
 analyze' :: (Has Exec sig m, Has Diagnostics sig m) => Path Abs Dir -> m (Graphing Dependency)
 analyze' dir = do
   -- Get all dependencies
-  depsAllEnvTree <- context "Identifying relationship among dependencies" $ execParser mixTreeCmdOutputParser dir $ mixDepTreeCmd (Nothing)
-  depsAllResolved <- context "Inferring dependencies versioning" $ execParser mixDepsCmdOutputParser dir mixDepCmd
+  depsAllEnvTree <- context "Identifying relationship among dependencies" $ execParser mixTreeCmdOutputParser dir mixDepTreeCmd <&> supportedMixDeps
+  depsAllResolved <- context "Inferring dependencies versioning" $ execParser mixDepsCmdOutputParser dir mixDepCmd <&> Map.filter (supportedSCM <$> depResolvedSCM)
```
scruffystuffs, line 75: I'd rather not see `<&>` used in the codebase, we don't really use it or `&` anywhere else.
Also, the `<$>` in `(supportedSCM <$> depResolvedSCM)` is just `.`.  I mentioned it on the poetry PR, but `-> a` has a functor instance, where `fmap = (.)`.  I've corrected that in my suggestion.

```suggestion
  depsAllEnvTree <- 
    context "Identifying relationship among dependencies" $
      supportedMixDeps <$> execParser mixTreeCmdOutputParser dir mixDepTreeCmd
  depsAllResolved <-
    context "Inferring dependencies versioning" $
      Map.filter (supportedSCM . depResolvedSCM)
        execParser mixDepsCmdOutputParser dir mixDepCmd
```

src/Strategy/Elixir/MixTree.hs
```diff
@@ -138,68 +166,60 @@ parseDepSCM = (try (try parseDepHex <|> parseDepSCMGit) <|> parseDepSCMLocalPath
...
-mixDepsCmdOutputParser = M.fromList <$> (parseDep `sepBy` char '*') <* eof
+mixDepsCmdOutputParser :: Parser (Map PackageName MixDepResolved)
+mixDepsCmdOutputParser = Map.fromList <$> (parseDep `sepBy` char '*') <* eof
   where
     findName :: Parser Text
     findName = takeWhileP (Just "dep") (/= ' ')
 
     findVersion :: Parser Text
-    findVersion = T.pack <$> some (alphaNumChar <|> char '.' <|> char '-')
+    findVersion = Text.pack <$> some (alphaNumChar <|> char '.' <|> char '-')
```
scruffystuffs, line 196: [nit]: Use `toText` from `Data.String.Conversion` instead of `Text.pack`.
```suggestion
    findVersion = toText <$> some (alphaNumChar <|> char '.' <|> char '-')
```

src/Strategy/Elixir/MixTree.hs
```diff
@@ -138,68 +166,60 @@ parseDepSCM = (try (try parseDepHex <|> parseDepSCMGit) <|> parseDepSCMLocalPath
...
       name <- findName <* space
 
       -- If dependencies are not resolved, version won't be printed
       -- This can occur if lock file is missing
-      version <- optional . try $ do CEq <$> findVersion
+      version <- optional . try $ CEq <$> findVersion
 
-      -- Source control manager of the pkg - e.g. git, hex
+      -- Source control manager of the package - e.g. git, hex
       scm <- space *> between (char '(') (char ')') parseDepSCM
```
scruffystuffs, line 211: We talked offline about using `space` directly, but I just wanted to leave a comment so we didn't forget: use a local `sc` or `scn` (**S**pace **C**onsumer, or **S**pace **C**onsumer with **N**ewlines), see other strategies for examples

src/Strategy/Elixir/MixTree.hs
```diff
@@ -214,15 +234,12 @@ mixTreeParser = concat <$> ((try (mixDep 0) <|> ignoredLine) `sepBy` eol) <* eof
     isEndLine '\r' = True
     isEndLine _ = False
 
-    -- ignore content until the end of the line
+    ignoredLine :: Parser [MixDep]
+    ignoredLine = ignored $> []
+
     ignored :: Parser ()
     ignored = () <$ takeWhileP (Just "ignored") (not . isEndLine)
```
scruffystuffs, line 241: [nit]: use `void` from Data.Functor` instead.
```suggestion
    ignored = void $ takeWhileP (Just "ignored") (not . isEndLine)
```

scruffystuffs: from the source for `Data.Functor`

```haskell
void :: Functor f => f a -> f ()
void x = () <$ x
```

src/Strategy/Elixir/MixTree.hs
```diff
@@ -233,62 +250,64 @@ mixTreeParser = concat <$> ((try (mixDep 0) <|> ignoredLine) `sepBy` eol) <* eof
...
+
+      -- Dependency's scm
+      scm <- between (char '(') (char ')') parseDepSCM
       _ <- takeWhileP (Just "ignored") (not . isEndLine)
 
-      deps <- many $ try $ mixRecurse $ depth + 1
-      pure [MixDep (PackageName dep) (toDependencyVersion =<< version) location (concat deps)]
+      -- Deep Dependencies
+      deps <- many $ try $ mixRecurse (depth + 1)
+      pure [MixDep (PackageName dep) (toDependencyVersion =<< version) scm (concat deps)]
```
scruffystuffs, line 265: [optional nit]: prefer `>>=` to `=<<`.  It's a little unintuitive considering that most similar operators go the other direction, but it does seem to be more idiomatic in other haskell libraries.
```suggestion
      pure [MixDep (PackageName dep) (version >>= toDependencyVersion) scm (concat deps)]
```

src/Strategy/Elixir/MixTree.hs
```diff
@@ -338,29 +356,27 @@ parseVerConstraint = do
...
+      "==" -> pure Equal
+      "!=" -> pure NotEqual
+      ">=" -> pure GreaterThanOrEqual
+      "<=" -> pure LessThanOrEqual
+      ">" -> pure GreaterThan
+      "<" -> pure LessThan
+      "~>" -> pure Compatible
+      "*" -> pure WildcardAny
+      "=" -> pure Equal
+      other -> fail ("Could not recognize mix constraint operator: (" <> Text.unpack other <> ")")
```
scruffystuffs, line 377: [nit]: prefer `toString` from `Data.String.Conversion` over `Text.unpack`
```suggestion
      other -> fail ("Could not recognize mix constraint operator: (" <> toString other <> ")")
```

docs/strategies/elixir.md
```diff
@@ -0,0 +1,136 @@
...
+
+1. Run `mix deps.tree --format plain` and generate output similar to:
+```
+├─ one─1.0.0 (https://github.com/dep/one)
+│  ├─ two─2.0.0 (hex package)
+│  │  └─ three─3.0.0 (hex package)
+│  └─ four─4.0.0 (https://github.com/dep/four)
+└─ five─5.0.0 (hex package)
+```
+2. Run `mix deps --all` and generate output similar to:
```
scruffystuffs, line 23: If we've changed our commands to handle prod vs. dev differently, we should sync these docs with those changes.  if not, then these are OK.

src/Strategy/Mix.hs
```diff
@@ -0,0 +1,44 @@
+module Strategy.Mix (
```
scruffystuffs, line 1: Love the size and organization of this file 👍

scruffystuffs (changes requested):
A few changes, but along with the feedback from @zlav, this is pretty close.

docs/strategies/elixir.md
meghfossa: No we need to explicitly say `--all`. If not done so, environment variable `MIX_ENV` can influence results returned. This looks right to me.

src/Strategy/Elixir/MixTree.hs
meghfossa: I've added this suggestion to style guide now - https://github.com/fossas/team-analysis/issues/671

src/Strategy/Elixir/MixTree.hs
meghfossa: +in style guide: https://github.com/fossas/team-analysis/issues/671

src/Strategy/Elixir/MixTree.hs
meghfossa: +style guide https://github.com/fossas/team-analysis/issues/671

src/Strategy/Elixir/MixTree.hs
meghfossa: +style guide https://github.com/fossas/team-analysis/issues/671

src/Strategy/Elixir/MixTree.hs
meghfossa: +style guide https://github.com/fossas/team-analysis/issues/671

src/Strategy/Elixir/MixTree.hs
```diff
@@ -0,0 +1,374 @@
...
+whitespaceOrTab = takeWhileP (Just "whitespaceOrTab") isSpaceOrTab
+  where
+    isSpaceOrTab c = c == ' ' || c == '\t'
+
+-- | Parses `VerConstraint`
+parseVerConstraint :: Parser VerConstraint
+parseVerConstraint = do
+  operator <- whitespaceOrTab *> parseConstraintOperator <* whitespaceOrTab
+  versionText <- findVersionText <* whitespaceOrTab
+  case operator of
```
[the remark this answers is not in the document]
meghfossa: I've modified the code a bit to collapse multiple function into one. I intent to abstract this piece of code out in future PR, once poetry PR is merged. The parsing for version constraint is fairly similar among them - and I'm presuming it will be for lots of languages. I tried the above code, but it led into test failing for irregular spacing. If you insist I can tackle it in this pr. My preference is that I address this subsequent PR.

src/Strategy/Elixir/MixTree.hs
```diff
@@ -0,0 +1,374 @@
...
+  choice
+    [ Mix <$ string "mix"
+    , Rebar3 <$ string "rebar3"
+    , Rebar <$ string "rebar"
+    , Make <$ string "make"
+    ]
+
+-- | Parses Source Control Manager
+parseDepSCM :: Parser DepSCM
+parseDepSCM = (try (try parseDepHex <|> parseDepSCMGit) <|> parseDepSCMLocalPath)
```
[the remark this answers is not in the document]
meghfossa: I've renamed to Other. I would like to keep this as a catch all - I've updated a code to filter out Others when resolving graph dependencies - that I think should take care of it.

src/Strategy/Elixir/MixTree.hs
```diff
@@ -0,0 +1,374 @@
...
+  "==" -> pure Equal
+  "!=" -> pure NotEqual
+  ">=" -> pure GreaterThanOrEqual
+  "<=" -> pure LessThanOrEqual
+  ">" -> pure GreaterThan
+  "<" -> pure LessThan
+  "~>" -> pure Compatible
+  "*" -> pure WildcardAny
+  "=" -> pure Equal
+  other -> fail ("Could not recognize mix constraint operator: (" <> toString other <> ")")
```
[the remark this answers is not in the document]
meghfossa: Yeah - I don't think it's possible, unless new operators are added to operator list without updating textToMixVersion.

docs/strategies/elixir.md
scruffystuffs: It looks like the first command changed to include prod, right?  you added `--only prod`

meghfossa: That's correct. I want to capture all resolved dependencies with `mix deps --all`. 

The code only reports dependencies from `mix deps.tree --only prod` but uses resolved versions captured from `mix deps --all`.

docs/strategies/elixir.md
```diff
@@ -0,0 +1,136 @@
...
+| -------- | ------------------ | ------------------ | ------------------ |
+| mix deps | :white_check_mark: | :white_check_mark: | :white_check_mark: |
+
+## Project Discovery
+
+In order to find elixir projects, we look for `mix.exs` file which specify the root of a elixir project. We skip `deps` and `_build` directories. 
+
+## Analysis
+
+1. Run `mix deps.tree --format plain` and generate output similar to:
```
scruffystuffs:
```suggestion
1. Run `mix deps.tree --format plain --only prod` and generate output similar to:
```

docs/strategies/elixir.md
scruffystuffs: I might not be as clear as I think I'm being.  See [this comment](https://github.com/fossas/spectrometer/pull/287/files/99cea438c20e58879ceb761d421f18f47b4e658b#r674192081) for what I was expecting.

docs/strategies/elixir.md
scruffystuffs: nvm, this is already done, we were discussing an old commit after you had fixed it, my bad.

docs/strategies/elixir.md
scruffystuffs: I just noticed that this is done in the latest commit, and we're commenting on an old commit.

src/Strategy/Elixir/MixTree.hs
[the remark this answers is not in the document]
meghfossa: I was able to make this work. This should have been `Nothing -> return $ CEq versionText` instead of `Nothing -> fail ("Could not recognize mix constraint operator:")`. This is for case when user simply does not use any operator e.g. myPkg: "2.2" vs myPkg: ">2.2"

meghfossa:
> > Are you saying that there vendored dependencies that have a `rebar.config` file which will be picked up and scanned?
> 
> Ah! As mix can use few build manager (make, rebar, rebar3, mix). Some dependency may, for example have rebar3 as build manager.
> 
> Here is an example:
> 
> mix.lock file
> 
> ```elixir
> %{
>   "hackney": {:hex, :hackney, "1.15.2", "07e33c794f8f8964ee86cebec1a8ed88db5070e52e904b8f12209773c1036085", [:rebar3], [{:certifi, "2.5.1", [hex: :certifi, repo: "hexpm", optional: false]}, {:idna, "6.0.0", [hex: :idna, repo: "hexpm", optional: false]}, {:metrics, "1.0.1", [hex: :metrics, repo: "hexpm", optional: false]}, {:mimerl, "~>1.1", [hex: :mimerl, repo: "hexpm", optional: false]}, {:ssl_verify_fun, "1.1.5", [hex: :ssl_verify_fun, repo: "hexpm", optional: false]}], "hexpm", "e0100f8ef7d1124222c11ad362c857d3df7cb5f4204054f9f0f4a728666591fc"},
> }
> ```
> 
> When compiling and getting dependencies, it will download build config for `hackney` in deps directory.
> 
> ```
> >> tree elxirEampleProject/deps/hackney
> ...
> ├── _build
> │   └── default
> │       └── lib
> ├── ebin
> ├── hex_metadata.config
> ├── rebar.config
> ├── rebar.lock
> └── src
>     ├── hackney.app.src
> ....
> ```
> 
> This is getting discovered as it's own project with analyze. I'm not sure how to exclude this. I did skip deps directory in my strategy's discovery, but since rebar strategies are independent they pick these up as projects.

I and @zlav discussed potential options in regards to this. Options discussed were, 

1) Modify Rebar3 strategy to exlude _deps directory when mix.exs is found. 
2) No Code Change. Use filter options to exclude _deps directory. 

We decided to go with (2) for now. Main reason for doing so is that approach (1) introduces a change in sibling strategy (2) It deviates the notion of independent strategies and introduces some coupling. Aforementioned concern may only be applicable to small segment of elixir project. For such project alternate solution exists as of now. I will update relevant docs for the strategy.

src/Strategy/Elixir/MixTree.hs
```diff
@@ -350,42 +339,31 @@ toDependencyVersion dt = case parse parseConstraintExpr "" dt of
...
+    Just "=" -> return $ CEq versionText
+    Just "!=" -> return $ CNot versionText
+    Just ">=" -> return $ CGreaterOrEq versionText
+    Just ">" -> return $ CGreater versionText
+    Just "<=" -> return $ CLessOrEq versionText
+    Just "<" -> return $ CLess versionText
+    Just "~>" -> return $ CCompatible versionText
+    Just "*" -> return $ CEq "*"
+    Just _ -> fail ("Could not recognize mix constraint operator:")
+    Nothing -> return $ CEq versionText
```
scruffystuffs, line 354: [nit]: Remember to use `pure` instead of `return`

src/Strategy/Elixir/MixTree.hs
```diff
@@ -350,42 +339,31 @@ toDependencyVersion dt = case parse parseConstraintExpr "" dt of
...
-      "!=" -> pure NotEqual
-      ">=" -> pure GreaterThanOrEqual
-      "<=" -> pure LessThanOrEqual
-      ">" -> pure GreaterThan
-      "<" -> pure LessThan
-      "~>" -> pure Compatible
-      "*" -> pure WildcardAny
-      "=" -> pure Equal
-      other -> fail ("Could not recognize mix constraint operator: (" <> Text.unpack other <> ")")
+    findVersionText = Text.pack <$> some (alphaNumChar <|> char '.' <|> char '-' <|> char '*' <|> char '+')
```
scruffystuffs, line 365: [nit]: Use `toText` from `Data.String.Conversion`
```suggestion
    findVersionText = toText <$> some (alphaNumChar <|> char '.' <|> char '-' <|> char '*' <|> char '+')
```

src/Strategy/Elixir/MixTree.hs
```diff
@@ -350,42 +339,31 @@ toDependencyVersion dt = case parse parseConstraintExpr "" dt of
...
+    Just "*" -> return $ CEq "*"
+    Just _ -> fail ("Could not recognize mix constraint operator:")
+    Nothing -> return $ CEq versionText
   where
-
     whitespaceOrTab :: Parser Text
     whitespaceOrTab = takeWhileP (Just "whitespaceOrTab") (\c -> c == ' ' || c == '\t')
 
+    parseConstraintOperator :: Parser (Maybe Text)
+    parseConstraintOperator = optional (asum (map symbol operatorList))
```
scruffystuffs, line 360: [nit]: Use `$` and `.` instead of brackets in easy cases like this.
```suggestion
    parseConstraintOperator = optional . asum $ map symbol operatorList
```

scruffystuffs (commented):
Few nits, I'll do a full review soon.

src/Strategy/Elixir/MixTree.hs
```diff
@@ -0,0 +1,378 @@
...
+        else Nothing
+
+-- | Parses Source Control Manager
+parseDepSCM :: Parser DepSCM
+parseDepSCM = try parseDepHex <|> parseDepSCMGit <|> parseDepSCMOther
+  where
+    parseDepHex :: Parser DepSCM
+    parseDepHex = do
+      _ <- chunk "Hex package"
+      pure (Hex)
```
scruffystuffs, line 183: [nit]: Avoid `do`-notation
```suggestion
    parseDepHex = Hex <$ chunk "Hex package"
```

scruffystuffs (approved):
Looks good to me.  You'll need to rebase on master to get CI to pass.  I left a few nits, but nothing else needs to be addressed, you can merge this when you're ready.

Make sure you fix the changelog before merging.

meghfossa:
> > I and @zlav discussed potential options in regards to this. Options discussed were,
> > 
> > 1. Modify Rebar3 strategy to exlude _deps directory when mix.exs is found.
> > 2. No Code Change. Use filter options to exclude _deps directory.
> > 
> > We decided to go with (2) for now. Main reason for doing so is that approach (1) introduces a change in sibling strategy (2) It deviates the notion of independent strategies and introduces some coupling. Aforementioned concern may only be applicable to small segment of elixir project. For such project alternate solution exists as of now. I will update relevant docs for the strategy.
> 
> The reason I suggested doing number 1 is that it's really easy to do, usually correct (as long as it's fully checked), and already has precedent in our codebase ([`npm` ignores `package.json` if it also finds `yarn.lock`](https://github.com/fossas/spectrometer/blob/b436f4fab8da37da1b1066f22b67d24932c53751/src/Strategy/Npm.hs#L24)). While most good code is dumb code, scanning for and identifying projects is one of the few things that needs more nuanced processing. We don't have to do it in this PR, but I'd like to handle it sooner rather than later.
> 
> We can't keep strategies truly independent if they don't actually stay independent in the real world. Coupling in our codebase is unavoidable if the actual tools are coupled.

Only potential issue I see with this approach is that it will be incorrect, if the projects follow umbrella structure. Albeit, this may be not as frequent.

PR #291, opened 2021-07-20
Exclusion targets: monorepo interop

kitified:
# Overview

Update monorepo code to work with the new exclusion and inclusion paths functionality.

_Blocked on monorepo plugin support landing (in PR)_

## Acceptance criteria

* Accept `--only-path` and `--exclude-path` arguments, passing them along to the monorepo plugin
* Error on combination of `--experimental-enable-monorepo aosp` and `--filter`, because monorepo scans don't have targets
* Error on combination of `--experimental-enable-monorepo aosp` and `--exclude-target` or `--include-target`, because monorepo scans don't have targets

## Testing plan

Prior to building filter support in the monorepo plugin, I ran the following sets of commands with the attached results, which were all what I wanted to see.

_Demonstrates passing paths successfully:_
```shell
$ cabal run fossa -- analyze --experimental-enable-monorepo aosp -e '<url>' --fossa-api-key <key> --only-path foo --only-path foo2 --exclude-path bar <dir>
[ INFO] Running monorepo scan
[ERROR] ----------
  An error occurred:

      Command execution failed:
          command: /private/var/folders/qj/b_t4_wcx6j50zt70r7drqbn00000gn/T/fossa-vendor/vsi-plugin
          args: [ monorepo
                , -endpoint
                , http://localhost:9578
                , -fossa-api-key
                , 2dd3cf1d3e0a76eb78f9d33d4572cbac
                , -project
                , aosp-mini
                , -revision
                , 1626742024
                , -only-paths
                , foo/
                , -only-paths
                , foo2/
                , -exclude-paths
                , bar/
                , -type
                , aosp
                , . ]
          dir: /Users/kit/projects/aosp-mini/
          exit: ExitFailure 2
          stdout:

          stderr:
            2021/07/19 17:47:05 maxprocs: Leaving GOMAXPROCS=12: CPU quota undefined
            flag provided but not defined: -only-paths
```

_Demonstrates error on `--only-target`_
```shell
$ cabal run fossa -- analyze --experimental-enable-monorepo aosp -e '<url>' --fossa-api-key <key> --only-path foo --exclude-path bar --only-target x <dir>
The --only-target and --exclude-target options are invalid for monorepo projects. Instead use --only-path and/or --exclude-path to control which paths are scanned
```

_Demonstrates error on `--filter`_
```shell
$ cabal run fossa -- analyze --experimental-enable-monorepo aosp -e '<url>' --fossa-api-key <key> --only-path foo --exclude-path bar --filter vsi@. <dir>
The --filter option is invalid for monorepo projects. Refer to the new path exclusion feature for how to filter monorepo project files. --filter will be removed by v2.20.0
```

After building support in the pluin, I used different sets of filters to exclude directories from scans and they worked as expected.

## Risks

I don't think there's much risk here, we're just reading some data already being parsed by the base PR.

## Checklist

### DIFF
--- a/src/App/Fossa/VPS/Scan/RunWiggins.hs
+++ b/src/App/Fossa/VPS/Scan/RunWiggins.hs
@@ -9,6 +9,7 @@ module App.Fossa.VPS.Scan.RunWiggins (
   generateWigginsMonorepoOpts,
   WigginsOpts (..),
   ScanType (..),
+  MonorepoFilters (..),
 ) where
 
 import App.Fossa.EmbeddedBinary
@@ -43,6 +44,11 @@ data WigginsOpts = WigginsOpts
   , spectrometerArgs :: [Text]
   }
 
+data MonorepoFilters = MonorepoFilters
+  { onlyPaths :: [Path Rel Dir]
+  , excludePaths :: [Path Rel Dir]
+  }
+
 generateWigginsScanOpts :: Path Abs Dir -> Severity -> ProjectRevision -> ScanType -> FilterExpressions -> ApiOpts -> ProjectMetadata -> WigginsOpts
 generateWigginsScanOpts scanDir logSeverity projectRevision scanType fileFilters apiOpts metadata =
   WigginsOpts scanDir $ generateSpectrometerScanArgs logSeverity projectRevision scanType fileFilters apiOpts metadata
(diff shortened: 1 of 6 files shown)


src/App/Fossa/VPS/Scan/RunWiggins.hs
```diff
@@ -43,6 +46,17 @@ data WigginsOpts = WigginsOpts
   , spectrometerArgs :: [Text]
   }
 
+data PathFilters = PathFilters
+  { onlyPaths :: [Path Rel Dir]
+  , excludePaths :: [Path Rel Dir]
+  }
+
+toText :: Path a b -> Text
+toText = T.pack . toFilePath
```
scruffystuffs, line 55: This is already defined in `Data.String.Conversion`

scruffystuffs: permalink for reference: https://github.com/fossas/spectrometer/blob/f2f56bea9a9445bb42ad4938f56564553e497879/src/Data/String/Conversion.hs#L75-L76

kitified: Oh 🤦‍♀️ thanks!

scruffystuffs: Yeah, that module is sort of new, I'm trying to help focus all of our string conversion to that module.

zlav (approved):
This looks great! I don't actually have any comments since this is essentially just piping the filters we already have into the VSI binary. If you rebase on master you will find that the dependency-analysis step is fixed.

PR #292, opened 2021-07-20
fix: moduleArtifacts is sometimes empty

liftM:
# Overview

This PR fixes an issue where `moduleArtifacts` could be empty, causing a silent `NoSuchElementException` that would cause certain Gradle configurations to not resolve dependencies.

Please pay particular attention to the documentation and comment changes. Do these help improve understandability? I would like to use the documentation I wrote here as the model for docs for all future analyzers.

## Acceptance criteria

Gradle projects relying on dependencies with no module artifacts (e.g. `jackson-bom` or `junit-bom`) should successfully resolve.

## Testing plan

Put this `build.gradle.kts` file in a new empty directory to use as a test project:

```kotlin
/*
 * This file was generated by the Gradle 'init' task.
 *
 * This generated file contains a sample Java library project to get you started.
 * For more details take a look at the 'Building Java & JVM projects' chapter in the Gradle
 * User Manual available at https://docs.gradle.org/7.1.1/userguide/building_java_projects.html
 */

plugins {
    // Apply the java-library plugin for API and implementation separation.
    `java-library`
}

repositories {
    // Use Maven Central for resolving dependencies.
    mavenCentral()
}

dependencies {
    testImplementation("org.junit.jupiter:junit-jupiter:5.7.1")
    testImplementation("org.mockito:mockito-junit-jupiter:3.11.2")
    testImplementation("commons-io:commons-io:2.10.0")

    implementation("com.google.guava:guava:30.1-jre")
    implementation("org.apache.calcite:calcite-core:1.27.0")
    implementation("org.projectlombok:lombok:1.18.8")
    implementation("com.github.vertical-blank:sql-formatter:1.0")
    implementation("org.apache.commons:commons-collections4:4.2")
    implementation("org.apache.avro:avro:1.8.2")
    implementation("org.apache.commons:commons-lang3:3.12.0")
    implementation("com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.12.3")
    annotationProcessor("org.projectlombok:lombok:1.18.8")
}

tasks.test {
    // Use JUnit Platform for unit tests.
    useJUnitPlatform()
}
```

(Run `gradle :dependencies` in the directory to see that this is a valid project, and to make sure your `gradle` is working.)

Run `fossa analyze --output` using a build from `master` on the directory containing this file. Notice that no dependencies are found.

Then, switch to this PR, and try again. You should now get working dependencies.

## Risks

I'm not sure how to actually unit test this change. @cnr, any ideas on unit testing this script?

If not, I think I'll add this to our integration tests once they merge back into Spectrometer (i.e. once https://github.com/fossas/team-analysis/issues/676 lands).

## References

Resolves support conversation in Slack [here](https://teamfossa.slack.com/archives/CJBPV3Z0E/p1626129892369200).

## Checklist


docs/strategies/gradle.md
```diff
@@ -1,29 +1,190 @@
...
+
+1. Gradle build plugin
+2. Parsing `gradle :dependencies` (not yet implemented)
+
+### Gradle build plugin
+
+|                    |                                                                                          |
+| ------------------ | ---------------------------------------------------------------------------------------- |
+| :heavy_check_mark: | This tactic reports dependencies for all subprojects.                                    |
+| :heavy_check_mark: | This tactic provides a graph for subproject dependencies.                                |
```
cnr, line 105: I'm not sure I would understand the importance of this as an end-user

scripts/jsondeps.gradle
```diff
@@ -21,31 +37,79 @@
 //     Project Name -- first-party (sub)projects
 //   | Package Name Version [Dependency]
 
+
+// TODO: Only print debug logging when running in debug mode?
```
cnr, line 41: I suspect the excessive debug logging of each module artifact for each dependency of each configuration of each subproject will tank performance for particularly-large projects.

Notably, I've seen the output of a customer's project that was over 100 MB(!), and I think this extensive debug logging would substantially slow it down

scripts/jsondeps.gradle
```diff
@@ -21,31 +37,79 @@
...
                     def childResults = []
                     if (!resolvedDep.children.isEmpty()) {
+                        // childResolvedDep: https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/ResolvedDependency.html
                         resolvedDep.children.each { childResolvedDep ->
-                            def result = depToJSON childResolvedDep
-                            childResults << result
+                            // If we don't do this check, then calling
+                            // `resolvedDep.moduleArtifacts.iterator().next()`
+                            // above explodes because the `moduleArtifacts`
+                            // iterator is empty.
```
cnr, line 89: It would be nice to move this check closer to the invocation of `iterator().next()` -- seems error-prone to keep them separate. Could return `null` from the recursive `depToJSON` call and do a null-check below before adding to `childResults`

src/Strategy/Gradle.hs
```diff
@@ -72,35 +101,47 @@ discover ::
...
-discover dir = context "Gradle" $ do
-  projects <- context "Finding projects" $ findProjects dir
-  pure (map mkProject projects)
-
-pathToText :: Path ar fd -> Text
-pathToText = T.pack . toFilePath
-
+discover dir =
+  context "Gradle" $
+    (fmap . fmap) mkProject $ context "Finding projects" $ findProjects dir
```
cnr, line 106: I know the do-notation version isn't pleasant, but this feels less readable to me

docs/strategies/gradle.md
```diff
@@ -1,29 +1,190 @@
-# Gradle Analysis
+# Gradle
 
-When using gradle, users specify dependencies and plugins with a groovy DSL. This makes accurate dependency analysis particularly difficult: gradle projects tend to contain several (sub)projects with their own buildscripts, and plugins that add dependencies of their own.
+[Gradle](https://gradle.org/) is a polyglot tool mostly used by JVM projects. It's popular with Java and Android projects.
 
-To accurately determine the entire project/dependency structure, we use an
-initscript to inject our own `jsonDeps` task.
+Gradle users generally specify their builds using a `build.gradle` file, written in Groovy. These builds are specified as programs that must be dynamically evaluated.
```
cnr, line 5:
```suggestion
Gradle users generally specify their builds using a `build.gradle` file, written in Groovy, or a `build.gradle.kts` file, written in Kotlin. These builds are specified as programs that must be dynamically evaluated.
```

cnr (approved):
Looks good generally; left a lot of mostly-minor comments

---

> any ideas on unit testing this script
No, short of breaking the script out into a separate repo (which I'd certainly like to do).

> Please pay particular attention to the documentation and comment changes. Do these help improve understandability?
Yes, these docs are great and seem approachable.

It might be helpful to occasionally link to key terms in a glossary or other relevant documentation -- especially for the warnings like "This strategy requires dynamic analysis", which probably wouldn't mean anything to me as a user (both what "dynamic analysis" means, and the impact of requiring it). I also likely wouldn't understand the ramifications of "This tactic provides a graph for subproject dependencies" without a link to additional documentation.

Additionally, linking to relevant documentation the first time we use our own defined terms like "tactic" or "strategy" or "analysis target" might be helpful, as those terms likely won't mean anything to a user directly navigating to the gradle docs page

docs/strategies/gradle.md
```diff
@@ -1,29 +1,190 @@
...
+
+1. Unpacks our init script to a temporary directory.
+2. Invokes the init script with `gradle jsonDeps -Ipath/to/init.gradle`.
+3. Parses the JSON output of the init script.
+
+This init script is implemented [here](https://github.com/fossas/spectrometer/blob/master/scripts/jsondeps.gradle) and bundled into the CLI during compilation.
+
+The script works by iterating through configurations, resolving their dependencies, and then serializing those dependencies into JSON.
+
+Currently, this script only reports dependencies in the `default` configuration of each subproject.
```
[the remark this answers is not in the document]
liftM: Yeah, I blindly copied this from older docs. I realized this was no longer true later, but forgot to fix it here.

docs/userguide.md
```diff
@@ -1,23 +1,49 @@
...
-    - [`fossa test`](#fossa-test)
-        - [Specifying a timeout](#specifying-a-timeout)
-        - [Print issues as json](#print-issues-as-json)
-4. [Common FOSSA Project Flags](#common-fossa-project-flags)
-5. [Frequently-Asked Questions](#frequently-asked-questions)
-    - [`fossa analyze`: Why wasn't my project found?](#fossa-analyze-why-wasnt-my-project-found)
-    - [When are you adding support for (some buildtool/language)](#when-are-you-adding-support-for-some-buildtoollanguage)
-    - [What are these experimental monorepo flags?](#what-are-these-experimental-monorepo-flags)
+<!-- omit in toc -->
+## Table of contents
```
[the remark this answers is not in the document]
liftM: I'm generating them using VSCode right now, but ideally I'd like to generate them with some tool and lint for them in CI. Not sure what the right candidate tool here is.

docs/strategies/gradle.md
```diff
@@ -1,29 +1,190 @@
...
+If you'd like to make a bug report about incorrect dependencies, make sure to include the list of incorrect dependencies, as well as the commands you ran to obtain that list.
+
+### Debugging the "Gradle build plugin" tactic
+
+The Gradle build plugin is a Gradle [init script](https://docs.gradle.org/current/userguide/init_scripts.html) implemented [here](../../scripts/jsondeps.gradle).
+
+If this tactic doesn't appear to be working (e.g. is giving you incorrect dependencies or is missing dependencies), you can run the init script directly using:
+
+```
+gradle -I$PATH_TO_SCRIPT $SUBPROJECT1:jsonDeps $SUBPROJECT2:jsonDeps ...
```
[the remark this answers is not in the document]
liftM: We should probably roll this up into a single debug-bundle-generating subcommand, like the other ideas we've been batting around for https://github.com/fossas/team-analysis/issues/610.

scripts/jsondeps.gradle
liftM: I'm going to try this for now to unblock the fix, and come back to figure out how to selectively debug log later. If we need to back this out in a hurry, we can always comment out the log lines.

liftM: I've logged the issue in https://github.com/fossas/team-analysis/issues/689.

docs/strategies/gradle.md
liftM: I think power users will understand this. I'm also open to a better explanation, but I can't think of one right now.

scripts/jsondeps.gradle
liftM: Good call, not sure what I was thinking here.

liftM:
Thanks for the thorough docs read!

PR #293, opened 2021-07-20
Use types to enforce that the user supplies both releaseGroupName and releaseGroupRelease

cnr:
Rather than parsing the release group `name` and `release` separately as `(Maybe Text, Maybe Text)`, combine the parse into `Maybe (Text, Text)` to enforce that the user supplies both or none. This removes the manual check that both are supplied from here: https://github.com/fossas/spectrometer/compare/release-group-config-cleanup?expand=1#diff-f2fe846f526d80e94301cb1315b5a660dea9fee7efe2fb1be098fad1112200a8L113-L116

When the user only supplies one, they receive an error message on top of the usage/help text:
```
$ fossa analyze --release-group-name foo
Missing: --release-group-release ARG

Usage: fossa analyze [-o|--output] [--unpack-archives] [--json]
                     [-b|--branch ARG] [-t|--title ARG] [-P|--project-url ARG]
                     [-j|--jira-project-key ARG] [-L|--link ARG] [-T|--team ARG]
                     [--policy ARG]
                     [--release-group-name ARG --release-group-release ARG]
                     [--filter ANALYSIS-TARGET]
                     [--experimental-enable-monorepo MODE] [DIR]
  Scan for projects and their dependencies

Available options:
[...]
```

### DIFF
--- a/src/App/Fossa/Configuration.hs
+++ b/src/App/Fossa/Configuration.hs
@@ -8,7 +8,6 @@ module App.Fossa.Configuration (
   ConfigFile (..),
   ConfigProject (..),
   ConfigRevision (..),
-  ConfigReleaseGroup (..),
   ConfigTargets (..),
   ConfigPaths (..),
 ) where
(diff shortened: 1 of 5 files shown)


zlav (approved):
This looks really good, with a much better error message. I don't think you need to update the changelog here but it might be nice to add something. Also you should make sure your PR description is in the format of the PR template.

PR #294, opened 2021-07-20
Run hlint and fourmolu in CI

scruffystuffs:
# Overview

Check for linter and formatter errors in CI, prevent merge if they fail.

## Testing plan

- Use existing errors caught by `make lint` and `make fmt`
  - Confirm that GH prevents merge.
- Fix errors
  - Confirm that CI allows merge

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -125,7 +125,7 @@ jobs:
 
     - name: Run hlint
       run: |
-        make hlint
+        make lint
   
   format-check:
     name: formatter-check
@@ -135,9 +135,15 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    # Run the formatter
     - name: run fourmolu
       run: |
-        make check-fmt
+        make fmt
+
+    # If git has chages, then the formatter check fails.
+    - name: check git status
+      run: |
+        git diff --name-only --exit-code
 
   create-release:
     name: create-release
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -114,6 +114,31 @@ jobs:
         name: ${{ runner.os }}-binaries
         path: release
 
+  # Run linter and format checker indpendently, so you see errors from both.
+  linter-check:
+    name: linter-check
+    runs-on: ubuntu-latest
+    container: ghcr.io/fossas/haskell-dev-tools:8.10.4
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Run hlint
+      run: |
+        make hlint
+  
+  format-check:
+    name: formatter-check
+    runs-on: ubuntu-latest
+    container: ghcr.io/fossas/haskell-dev-tools:8.10.4
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: run fourmolu
+      run: |
+        make check-fmt
+
   create-release:
     name: create-release
     if: ${{ startsWith(github.ref, 'refs/tags/v') }}
--- a/HACKING.md
+++ b/HACKING.md
@@ -59,6 +59,7 @@ In the base directory, run `cabal test`
 ### Installing haskell-language-server
 
 In VSCode:
+
 - Install the "Haskell Language Server" (`haskell.haskell`) plugin in VSCode.
 - In the Haskell extension settings, under `Haskell: Formatting Provider`, choose `fourmolu`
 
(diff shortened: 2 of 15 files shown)


.github/workflows/build.yml
```diff
@@ -114,6 +114,37 @@ jobs:
...
+
+    steps:
+    - uses: actions/checkout@v2
+
+    # Run the formatter
+    - name: run fourmolu
+      run: |
+        make fmt
+
+    # If git has chages, then the formatter check fails.
```
liftM, line 143:
```suggestion
    # If git has changes, then the formatter check fails.
```

.github/workflows/build.yml
```diff
@@ -114,6 +114,37 @@ jobs:
         name: ${{ runner.os }}-binaries
         path: release
 
+  # Run linter and format checker indpendently, so you see errors from both.
```
liftM, line 117:
```suggestion
  # Run linter and format checkers independently, so you see errors from both.
```

HACKING.md
```diff
@@ -79,7 +81,9 @@ In VSCode, the formatter can be activated with the standard format-file and form
 
 Make sure to run the formatter on any files you modify. Enabling `Editor: Format On Save` in VSCode can make satisfying this requirement easier.
 
-We do not recommend using formatters outside of HLS, as there are some issues with configuration (see [FAQ](#FAQ) for more info).
+You can run `make fmt` to run the formatter on the entire codebase, or `make check-fmt` to dry-run the formatter.
+
+`make check-fmt` is run in CI, so any formatting errors will revent merging.
```
liftM, line 86:
```suggestion
`make fmt` is run in CI, so any formatting errors will prevent merging.
```

HACKING.md
```diff
@@ -68,8 +69,9 @@ If you installed HLS in the old, complicated way, you can safely remove it.  HLS
 
 `hlint` is built into HLS, and is enabled by default. hlint suggestions appear as warnings in the editor.
 
-`hlint` errors may become required changes in pull requests, but we do not currently run `hlint` in CI, as there are a few outstanding lint errors that have not yet been fixed.
-You do not need to enforce that `hlint` passes to submit a PR, but it does help greatly, for both the author and reviewer.
+You can also use `make lint` to run the linter.
+
+`make lint` is run in CI, so any outlying errors will prevent merging.
```
liftM, line 74:
```suggestion
`make lint` is run in CI. Any errors will prevent merging.
```

ci/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# Build the correct binaries
+FROM fossa/haskell-static-alpine:ghc-8.10.4 as builder
+RUN cabal update
+RUN cabal install hlint
+RUN cabal install fourmolu
```
liftM, line 5: Not pinning these spooks me a little. This build is not reproducible. Can we pin the `hlint` and `fourmolu` versions, at least?

src/App/Fossa/Main.hs
```diff
@@ -111,9 +111,9 @@ appMain = do
           let apiOpts = ApiOpts optBaseUrl key
           let metadata = maybe analyzeMetadata (mergeFileCmdMetadata analyzeMetadata) fileConfig
           case (projectReleaseGroupName metadata, projectReleaseGroupRelease metadata) of
-                  (Just _, Just _) -> pure ()
-                  (Nothing, Nothing) -> pure ()
-                  _ -> die "releaseGroup.release and releaseGroup.name must both be specified if you want to associate this project to a release group."
+            (Just _, Just _) -> pure ()
```
liftM, line 114: FYI @meghfossa, the reason we use parentheses _here_ is because the value on the LHS being deconstructed is actually a _tuple_, and the parentheses are part of the expression itself.

liftM (approved):
LGTM, but please make the build reproducible.

I haven't run through the testing plan myself, but if you've tried with test PRs, I trust that this works (code seems correct - most likely failure mode is whether the `make X` commands actually exit 1 on failure properly).

ci/Dockerfile
scruffystuffs: We just pull the built image, we don't rebuild this every time.  It's reproducible for that reason.  I think pinning here is also not the right move, I'd prefer to always use the latest, I just didn't want to pay the cost of installing these every time.

In this specific case, what does reproducibility gain us?

scruffystuffs:
Addressing comments in #297

ci/Dockerfile
liftM: > we don't rerun this dockerfile every time

This is _exactly_ the problem. Someone who runs these commands locally may end up with a different set of tool versions than the tools being used in CI.

Pinning here makes it _extremely_ obvious which tool versions are being used.

I think we should pin because:

1. Pinning provides value - the tool versions become obvious.
2. Not pinning causes subtle problems - the tool versions can drift silently.
3. Not pinning does not provide significantly more value than pinning - you can simulate "always use latest" by just moving the pin.
4. There's really no cost here to pinning. The only cost I can think of is "we're not always on latest", but as you've mentioned, we're _already_ not on latest even without pinning. The only difference is that now we're on a _known_ version of not-latest instead of an _arbitrary depending on when the Docker build runs_ version of not-latest.

Being explicit is better than being magic.

PR #297, opened 2021-07-21
Fix CI issues

scruffystuffs:
- Address comments from #294
- Fix installer script to work with current GH action environment.


install.sh
```diff
@@ -47,12 +47,14 @@ execute() {
   # hash_sha256_verify "${tmpdir}/${TARBALL}" "${tmpdir}/${CHECKSUM}"
   srcdir="${tmpdir}"
   (cd "${tmpdir}" && unzip -o "${TARBALL}")
-  install -d -m 775 "${BINDIR}" 2> /dev/null || install -d "${BINDIR}"
+  log_debug "setting up bindir: $BINDIR"
+  mkdir -p $BINDIR
```
scruffystuffs, line 51: This `install` was the main problem.  We could use `sudo`, it'll work in GH, but I don't know if we want to do that for everyone else.

install.sh
```diff
@@ -47,12 +47,14 @@ execute() {
...
   for binexe in "fossa" ; do
     if [ "$OS" = "windows" ]; then
       binexe="${binexe}.exe"
     fi
-    install "${srcdir}/${binexe}" "${BINDIR}/" 2> /dev/null || sudo install "${srcdir}/${binexe}" "${BINDIR}/"
+    log_debug "installing binary: $binexe"
+    install "${srcdir}/${binexe}" "${BINDIR}/"
```
zlav, line 57: Is there a reason we're removing the `sudo` option?

scruffystuffs: No, that was for testing, I will re-add it.

install.sh
zlav: This install makes sense that it's the issue. If the user's CI doesn't have a `usr/local/bin` is the whole script going to break? I don't know if that's a valid issue, but we should be aware of it before merging.

scruffystuffs: The issue is that it already exists, and the `install` is trying to control it.  In CI, we could just `sudo install`, but I don't want to ship that as the installation script.  We just need the directory to exist, so we can simply use `mkdir -p`

The change to GH was that the directory used to not exist, but now it does.  Or ownership of the directory changed, it's hard to tell.

If the user doesn't have the directory, we will just make it.  We should probably fallback to `sudo mkdir`.

I'm going to refactor this script anyway, just so we can clear up the existing issue, then we can talk in detail about what we should be doing.

.github/workflows/dependency-scan.yml
```diff
@@ -20,8 +20,9 @@ jobs:
           cabal update
 
       - name: Install spectrometer from github
+        # To get this to work, we have to run the local script and then change it after.
         run: |
-          curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/spectrometer/master/install.sh | bash
+          ./install.sh -d
```
zlav, line 25: Is this still required? Shouldn't the script change fix the installation issue?

scruffystuffs: No, it has to be merged first, it's picking up the script from master, which hasn't changed and is still seeing the issue.

zlav: Gotcha, I just wanted to make sure that it wouldn't be an issue for others still.

zlav (approved):
Looks good! Thanks for diving into this.

.github/workflows/build.yml
```diff
@@ -82,9 +82,14 @@ jobs:
...
     - name: Build
+      # Occasionally, we run out of memory on the build process.
+      # Since cabal uses incremental compilation, we can retry from where we left off
+      #   by simply re-running cabal if we fail.
+      env:
+        RUN_CMD: cabal build --project-file=${{ matrix.project-file }} all
       run: |
         cabal update
-        cabal build --project-file=${{ matrix.project-file }} all
+        $RUN_CMD || $RUN_CMD
```
liftM, line 92: This is really spooky. I've opened a ticket to track this hack at https://github.com/fossas/team-analysis/issues/682.

PR #299, opened 2021-07-22
Use local installer for dep-scan CI job

scruffystuffs:
I don't see any benefit in downloading the latest master installer in our CI, since we can test the upcoming installer just as easy.

### DIFF
--- a/.github/workflows/dependency-scan.yml
+++ b/.github/workflows/dependency-scan.yml
@@ -20,7 +20,6 @@ jobs:
           cabal update
 
       - name: Install spectrometer from github
-        # To get this to work, we have to run the local script and then change it after.
         run: |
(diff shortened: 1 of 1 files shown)


zlav:
I can see some benefit in the fact that this is basically integration testing the install script. Ideally that would be a separate check like in CLI v1, but I'm personally a fan of keeping the curl bash logic in our CI.

scruffystuffs:
There's no benefit.  This is still integration testing the installer, but now it's actually testing the changes.  Previously, it was only testing **after** merging to master (which in the case of that file is the same as deploying), now it tests before merge/deploy.

zlav (approved):
Your point makes sense to me. This allows us to easier test changes to the install script. Looks good.

PR #308, opened 2021-07-28
Partial Graph Uploads

anuccio1:
# Overview

[Tech plan](https://docs.google.com/document/d/1q0TC0jR_wT4rxi2DMhkJA9UYyMNAs98WpaV5tAAKTAM/edit#heading=h.em4xrcdbhsh5)

Right now, FOSSA assumes that all dependencies from `srclib` analyses are direct dependencies, and that all dependencies from the CLI are a full graph, with no in-between. Because of this, we take the results from a `srclib` scan, and analyze each dependency, and then analyze the results from each of those scans for dependencies, and so on and so forth. This is often referred to as the “recursive step”, and is the strategy we use to construct dependency graphs for non-CLI builds. We do this, because we are unable to construct a full graph due to a lack of information provided from the source code.

In order to replace `srclib` with the `CLI`, we will need to add this concept to `CLI` output. This PR introduces the concept of `GraphBreadth` to the graph constructed in each strategy. This will be either `Complete`, or `Partial`, and signal to Core if it needs to do further dependency analysis or not

## Acceptance criteria

- Adds a `GraphBreadth` section to each SourceUnit in the output payload of `analyze`

## Testing plan

- Run analysis on several project types
- Check each strategy and confirm that it is in fact `Complete` or `Partial`

## Risks

- Low risk, as this is an addition to the payload, and will be unused for now

## References

[Tech plan](https://docs.google.com/document/d/1q0TC0jR_wT4rxi2DMhkJA9UYyMNAs98WpaV5tAAKTAM/edit#heading=h.em4xrcdbhsh5)

## Checklist

~- [ ] I added tests for this PR's change (or confirmed tests are not viable).~
~- [ ] If this PR introduced a user-visible change, I added documentation into `docs/`.~


src/Strategy/Googlesource/RepoManifest.hs
```diff
@@ -71,11 +71,14 @@ mkProject project =
...
+getDeps :: (Has ReadFS sig m, Has Diagnostics sig m) => RepoManifestProject -> m (Graphing Dependency, GraphBreadth)
 getDeps = analyze' . repoManifestXml
 
-analyze' :: (Has ReadFS sig m, Has Diagnostics sig m) => Path Abs File -> m (Graphing Dependency)
-analyze' file = buildGraph <$> nestedValidatedProjects (parent file) file
+-- TODO confirm this is Partial @scott
+analyze' :: (Has ReadFS sig m, Has Diagnostics sig m) => Path Abs File -> m (Graphing Dependency, GraphBreadth)
+analyze' file = do
+  graph <- buildGraph <$> nestedValidatedProjects (parent file) file
+  pure (graph, Partial)
```
scruffystuffs, line 81: cc @spatten

spatten: I haven't touched this code in a long time, and it didn't look like this when I wrote it :). 

I think @cnr changed this to make it output a `Graphing Dependency` back in October.

scruffystuffs: The underlying strategy probably hasn't changed, just the scaffolding around it.  Was it originally outputting a partial graph or a complete graph.  i.e.: would it benefit from recursive analysis of the deps discovered?

src/Strategy/Scala.hs
```diff
@@ -53,7 +53,7 @@ mkProject basedir closure =
     , projectPath = parent $ PomClosure.closurePath closure
     , projectBuildTargets = mempty
     , -- only do static analysis of generated pom files
-      projectDependencyGraph = \_ -> pure (Pom.analyze' closure)
+      projectDependencyGraph = \_ -> pure (Pom.analyze' closure, Complete)
```
scruffystuffs, line 56: [cleanup]: While you're here, can you just make this `const`?
```suggestion
      projectDependencyGraph = const $ pure (Pom.analyze' closure, Complete)
```

src/Strategy/VSI.hs
```diff
@@ -80,10 +80,12 @@ transformLocator :: ValidVSILocator -> Dependency
...
   Dependency (validType locator) (validName locator) (CEq <$> validRevision locator) [] [] mempty
 
-analyze :: (Has (Lift IO) sig m, MonadIO m, Has Exec sig m, Has Diagnostics sig m) => WigginsOpts -> m (Graphing Dependency)
+-- TODO: confirm this is Partial @kit
+analyze :: (Has (Lift IO) sig m, MonadIO m, Has Exec sig m, Has Diagnostics sig m) => WigginsOpts -> m (Graphing Dependency, GraphBreadth)
 analyze opts = context "VSI" $ do
   vsiLocators <- context "Running VSI binary" $ withWigginsBinary (execWigginsJson opts)
-  context "Building dependency graph" $ fromEither (buildGraph vsiLocators)
+  graph <- context "Building dependency graph" $ fromEither (buildGraph vsiLocators)
+  pure (graph, Partial)
```
scruffystuffs, line 88: cc @kitified

We're trying to determine if this strategy should be applied recursively.  `Partial` strategies are recursive, `Complete` strategies are non-recursive.

src/Types.hs
```diff
@@ -56,10 +57,28 @@ data DiscoveredProject m = DiscoveredProject
...
+  , projectDependencyGraph :: FoundTargets -> m (Graphing Dependency, GraphBreadth)
   , projectLicenses :: m [LicenseResult]
   }
 
+-- | The exhaustiveness or completeness of the graph found during analysis.
+--   Complete - indicates that the dependencies in question are a full, transitive graph, requiring no additional analysis
+--      Ex -> yarn.lock, Podfile.lock, Gemfile.lock
+--   Partial - indicates that the dependencies in question do *NOT* represent the fully resolved graph, i.e. because of a limitation
+--             in the package manager in use, it wasn't possible for us to obtain a full list of deps
+--      Ex -> stand-alone `package.json`, stand-alone Podfile
```
scruffystuffs, line 69: [nit]: It might be worth mentioning that the ramification of a `Partial` graph is that it is re-run recursively.

test/Go/GoListSpec.hs
```diff
@@ -69,7 +70,9 @@ spec = do
               & run
       case result of
         Left err -> expectationFailure ("analyze failed: " <> show (renderFailureBundle err))
-        Right graph -> graph `shouldBe` expected
+        Right (graph, graphType) -> do
+          graph `shouldBe` expected
+          graphType `shouldBe` Complete
```
scruffystuffs, line 75: [nit]: I don't think it's very useful to test the graph breadth, it's just a constant value
```suggestion
        Right (graph, _) -> graph `shouldBe` expected
```

test/Go/GoListSpec.hs
```diff
@@ -80,4 +83,6 @@ spec = do
 
       case result of
         Left err -> fail $ "failed to build graph" <> show (renderFailureBundle err)
-        Right graph -> length (graphingDirect graph) `shouldBe` 12
+        Right (graph, graphType) -> do
+          length (graphingDirect graph) `shouldBe` 12
+          graphType `shouldBe` Complete
```
scruffystuffs, line 88: [nit]: Same as above
```suggestion
        Right (graph, _) -> length (graphingDirect graph) `shouldBe` 12
```

scruffystuffs (approved):
Once you get confirmation from @kitified and @spatten, I'd say this is ready.

Love how simple this change is, left a few nits, but it's up to you if you want to do them.

src/Strategy/VSI.hs
kitified: The idea behind the VSI strategy is that it discovers a single target which is the project scan root. It then uploads all of the fingerprints to the VSI service, and reports a single flat dep graph of the direct dependencies found by the VSI service for the singular target, which encompasses the entire scan directory.

So if I'm understanding the change set properly, this strategy should be considered `Complete`.

src/Strategy/Googlesource/RepoManifest.hs
spatten: I think it's outputting a complete graph.

@kitified is that correct?

kitified: @spatten I am not familiar with the `Googlesource` strategy 😬

spatten: Interesting. I think this was originally built as part of AOSP parsing, and it might be dead code. I'm definitely not sure on that, though

I'll take a look and refresh my memory and report back here

spatten: I did some reading and remembering, and I can confirm that this is a partial graph.

I'm 90% sure that this is dead code, but that doesn't have anything to do with this PR.

So :+1: from me on the `Partial` being correct

src/Strategy/VSI.hs
anuccio1: thank you @kitified

PR #310, opened 2021-07-29
Add `mvn dependency:tree` tactic

scruffystuffs:
# Overview

Adds a `mvn dependency:tree`  tactic to the maven strategy.  This is a fallback strategy if the maven plugin does not succeed.

Simple overview:

* Run `mvn dependency:tree -DoutputType=dot -DoutputFile=<filename>`
  * `<filename>` is a random integer, and the constant string `fossa-deptree.dot`.  The random integer prevents multiple runs from picking up files from previous runs.  The constant string allows humans to understand what the file is, if they see it.
* Locate all `dot` file matching the expected file name.
* Read and parse the files in `dot` syntax.
* Construct a graph with the info in the `dot` syntax.

Unfinished business:

* What is the ideal graph of any multi-module project?  Does the graph produced by this strategy match that?  If not, is it even similar?
  * We should consider documenting the ideal "we can magically know and deduce everything about your project" graph for each strategy.  Without that, any new tactic is basically working from scratch (especially in JVM tactics).
* How are maven components addressed?  I'm pretty sure the current version is not correct.
* We drop files every time we run this tactic, they need to be cleaned up.

## Acceptance criteria

- [x] Graph produced makes sense with given Pom files
- [x] Files produced by the command are cleaned up (if we need them for debugging after, we can use record/replay)
- [ ] Docs are added reflecting the given overview

## Testing plan

Find a multi-module maven project (I used [apache/commons-rng](https://github.com/apache/commons-rng)), run analysis on it with the plugin disabled (hacked that in, see `Strategy.Maven.PluginStrategy`)

## References

Fixes fossas/team-analysis#669

## Checklist


scruffystuffs:
@liftM Here's the initial PR.  I'm still working on this and might have it ready to go today

scruffystuffs:
Re-assigning since zach is out.

src/Strategy/Maven/DepTree.hs
```diff
@@ -0,0 +1,180 @@
...
+  pure . DotGraph root $ concatMap sequenceEdges edgeLists
+
+-- | Produce tuples of every adjacent pair in the list.
+-- 1-item lists contain no edges, and are discarded.
+--
+-- >>> sequenceEdges [1, 2, 3, 4]
+-- [(1, 2), (2, 3), (3, 4)]
+-- >>> sequenceEdges [1]
+-- []
+sequenceEdges :: [a] -> [(a, a)]
```
cnr, line 141: drive-by clever snippet: we can implement this with `zip`

```hs
sequenceEdges :: [a] -> [(a, a)]
sequenceEdges xs = zip xs (drop 1 xs)
```

..because `drop 1` will align the next element in the list with the current element for the zip:
```
xs        = [1, 2, 3, 4]
drop 1 xs = [2, 3, 4]
```

scruffystuffs: 👏 👏 👏

liftM:
Still need to do docs, but the logic should be ready for review.

src/Strategy/Maven/DepTree.hs
```diff
@@ -0,0 +1,174 @@
...
+analyze :: (Has Exec sig m, Has ReadFS sig m, Has Diagnostics sig m, Has (Lift IO) sig m) => Path Abs Dir -> m (Graphing Dependency, GraphBreadth)
+analyze dir = do
+  -- We generate files with a random prefix so we don't pick up files from previous runs.
+  -- If we did pick them up, they're likely not enough of a conflict to worry about, but
+  -- it's better than not doing it.
+  randIdent <- abs <$> sendIO randomIO
+  _ <- context "Running maven 'dependency:tree' plugin" $ execThrow dir $ deptreeCmd randIdent
+  graphFiles <- context "Locating maven output files" $ findDepTreeOutputs dir randIdent
+  graphs <- context "Parsing output files" $ traverse (readContentsParser parseDotGraph) graphFiles
+  -- TODO: Will this work correctly in the face of exceptions?
```
cnr, line 67: an exception here will abort analysis / later files in the `graphFiles` list wouldn't be deleted

src/Strategy/Maven/DepTree.hs
```diff
@@ -0,0 +1,174 @@
...
+findDepTreeOutputs dir ident = execState @[Path Abs File] [] $
+  flip walk dir $ \_ _ files -> do
+    case findFileNamed (show ident <> outputFileName) files of
+      Nothing -> pure WalkContinue
+      Just file -> do
+        modify (file :)
+        pure WalkContinue
+
+buildGraph :: [DotGraph] -> Graphing Dependency
+buildGraph = gmap toDependency . foldr ((<>) . toGraph) mempty
```
cnr, line 81: [nit]
```suggestion
buildGraph = gmap toDependency . foldMap toGraph
```

src/Strategy/Maven/DepTree.hs
```diff
@@ -0,0 +1,174 @@
...
+  _ <- context "Running maven 'dependency:tree' plugin" $ execThrow dir $ deptreeCmd randIdent
+  graphFiles <- context "Locating maven output files" $ findDepTreeOutputs dir randIdent
+  graphs <- context "Parsing output files" $ traverse (readContentsParser parseDotGraph) graphFiles
+  -- TODO: Will this work correctly in the face of exceptions?
+  sendIO $ traverse_ removeFile graphFiles
+  pure (buildGraph graphs, Complete)
+
+findDepTreeOutputs :: (Has ReadFS sig m, Has Diagnostics sig m) => Path Abs Dir -> Int -> m [Path Abs File]
+findDepTreeOutputs dir ident = execState @[Path Abs File] [] $
+  flip walk dir $ \_ _ files -> do
```
cnr, line 73: [nit] we have a helper for this, `walk'`, that could help here

src/Strategy/Maven/DepTree.hs
```diff
@@ -0,0 +1,174 @@
...
+parseName :: forall m. MonadFail m => Text -> m PackageId
+parseName input = combine parts
+  where
+    parts :: [Text]
+    parts = Text.splitOn ":" input
+
+    combine :: [Text] -> m PackageId
+    combine [a, b, c, d, e] = pure . PackageId a b c d $ Just e
+    combine [a, b, c, d] = pure $ PackageId a b c d Nothing
+    combine items = fail $ toString ("invalid identifier" <> Text.intercalate ":" items)
```
cnr, line 162:
```suggestion
    combine items = fail $ toString ("invalid identifier " <> Text.intercalate ":" items)
```

src/Strategy/Maven/DepTree.hs
```diff
@@ -0,0 +1,174 @@
...
+  takeWhile1P,
+  try,
+ )
+import Text.Megaparsec.Char (space1)
+import Text.Megaparsec.Char.Lexer qualified as Lexer
+import Types (GraphBreadth (Complete))
+
+newtype DepTreeLabel
+  = BuildTag DepEnvironment
+  deriving (Eq, Ord, Show)
```
cnr, line 53: This looks unused

src/Strategy/Maven/DepTree.hs
```diff
@@ -0,0 +1,174 @@
...
+    , dependencyVersion = Just $ CEq artifactVersion
+    , dependencyLocations = []
+    , dependencyEnvironments = maybe [EnvProduction] ((: []) . toBuildTag) buildTag
+    , dependencyTags = mempty
+    }
+
+toGraph :: DotGraph -> Graphing PackageId
+toGraph DotGraph{rootNode, edgeList} = run . evalGrapher $ do
+  direct rootNode
+  for_ edgeList $ uncurry edge
```
cnr, line 97: [no action] ugh this makes me want a better interface for building `Graphing` s -- similar to `AdjacencyMap`, where you could use, e.g., `edges` here, and `edge` has a better type

cnr (approved):
LGTM generally; some nits

src/Strategy/Maven/DepTree.hs
liftM: What's the right thing to do here to make sure the files get cleaned up? `catch`?

cnr: something like
```hs
import Control.Exception.Extra (safeCatch)

ignoreExceptions :: Has (Lift IO) sig m => m a -> m ()
ignoreExceptions act = void act `safeCatch` (\(_ :: SomeException) -> pure ())
```

`traverse (ignoreExceptions . removeFile) ...`

(maybe even adding this to `Control.Exception.Extra`)

src/Strategy/Maven/DepTree.hs
liftM: I'm not sure how much we could really improve this by. There's always the trivial:

```hs
edges :: Ord t => [(t, t)] -> Graphing t -> Graphing t
```

Is that what you're talking about? Or do you mean a more drastic interface change?

liftM: Or maybe a graph-shaped helper like this?

```hs
makeGraphing :: [t] -> [(t, t)] -> Graphing t
makeGraphing directs edges = ...
```

cnr: I think `edge` and `direct` need to be simplified:
```hs
edge :: Ord a => a -> a -> Graphing a

direct :: a -> Graphing a

-- maybe with extra helpers like AdjacencyMap has, like you mentioned
directs :: Ord a => [a] -> Graphing a

edges :: Ord a => [(a,a)] -> Graphing a
```

..which would allow us to simplify `toGraph` here:
```hs
toGraph DotGraph{rootNode, edgeList} = direct rootNode <> edges edgeList
```

---

I think `Graphing.direct`/`Graphing.edge`/etc have near-zero use because they're not ergonomic at all, and I'd like to approximately mirror the design of `Algebra.Graph.AdjacencyMap` 's operations

src/Strategy/Maven/DepTree.hs
liftM: Ah, I see. I think I'm asking a different question - what's the right way to handle exceptions in L66 (i.e. in `readContentsParser`)?

In particular, I know `ReadFS` is an effect - is there a specific catching effect that I should use?

src/Strategy/Maven/DepTree.hs
liftM: > I'd like to approximately mirror the design of Algebra.Graph.AdjacencyMap 's operations

Yeah, that's a good idea. Semantically, we're basically an "AM with directs" anyway.

liftM:
I'm gonna cut the release first so we can get the fix out, and then work on docs in a follow-up.

src/Strategy/Maven/DepTree.hs
cnr: Yep exactly. `Graphing` is isomorphic to `AdjacencyMap (Node a)` where:

```hs
data Node a = Root | Node a
  deriving (Eq, Ord, Show)

type Graphing = AdjacencyMap (Node a)

direct :: Ord a => a -> Graphing a
direct = AM.edge Root (Node a)

edge :: Ord a => a -> a -> Graphing a
edge a b = AM.edge (Node a) (Node b)

..etc
```

but when writing it I preferred the clarity of it having its own distinct type and explicit set of direct deps

src/Strategy/Maven/DepTree.hs
scruffystuffs: ReadFS catches IO exceptions and reports them as diagnostics.  [See here](https://github.com/fossas/spectrometer/blob/72d79a21505522c84478a86a811f76eb9db59e44/src/Effect/ReadFS.hs#L199-L218) for the code that does this.

cnr: If `readContentsParser` fails, it'll abort with a diagnostics error as wes mentions. To ensure cleaning up files, you can use the suggestion above + chain `finally` off of line 66:

```hs
import Control.Effect.Exception (finally)

[...]
graphs <- (context "..." $ ... ) `finally` deleteAllTheFiles
[...]
```

or maybe create a helper for this too
```hs
withCleanup :: Has (Lift IO) sig m -> m b -> m a -> m a
withCleanup = flip finally
```

```hs
[...]
graphFiles <- ...
withCleanup (deleteAllThe graphFiles) $ do
  graphs <- ...
  pure (buildGraph ...)

PR #314, opened 2021-08-03
docs: Normalize Changelog style

liftM:
This PR normalizes the style conventions of the changelog:

1. Where reasonable, prefer sentences that read sensibly when the version is the first word of the sentence. For example, in a release for v1.2.3, you might write "Adds support for foo.", which would be read as "v1.2.3 adds support for foo.".
2. Only include changes that are user-visible. Users don't care about refactors, tech debt, style changes, etc.
3. Write entries as if they are being read by users, which usually means avoiding internal names and concepts. Users don't know what "analysis target configuration" is, but they do understand "we now support picking which subset of folders in CWD to analyze".
4. Some entries are "scoped" e.g. for languages or cross-team projects (e.g. containers, VPS).
5. The syntax for individual entries is is `- $SCOPE:? $DESCRIPTION_AS_A_SENTENCE. ($PR)`.

(Hmm, should I add these into `CONTRIBUTING.md`?)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -65,7 +65,7 @@
 - Fixes an issue where stdout doesn't always flush to the console. ([#265](https://github.com/fossas/spectrometer/pull/265))
 - Fixes an issue when referenced-dependencies are not being uploaded. ([#262](https://github.com/fossas/spectrometer/pull/262))
 - Adds support for `fossa-deps.json`. ([#261](https://github.com/fossas/spectrometer/pull/261))
-- Adds support for `vendored-dependencies` to be licensed scanned. ([#257](https://github.com/fossas/spectrometer/pull/257))
+- Adds support for `vendored-dependencies` to be license scanned. ([#257](https://github.com/fossas/spectrometer/pull/257))
 
 ## v2.8.0
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,276 +2,258 @@
 
 ## v2.13.0
 
-- Adds support for elixir projects using `mix` command ([#287](https://github.com/fossas/spectrometer/pull/287))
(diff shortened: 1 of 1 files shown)


meghfossa (approved):
looks good to me

Changelog.md
```diff
@@ -2,276 +2,258 @@
...
 ## v2.9.0
 
-- Fix an issue where stdout doesn't always flush to the console ([#265](https://github.com/fossas/spectrometer/pull/265))
-- Fix an issue when referenced-dependencies are not being uploaded ([#262](https://github.com/fossas/spectrometer/pull/262))
-- Adds support for `fossa-deps.json` ([#261](https://github.com/fossas/spectrometer/pull/261))
-- Adds support for `vendored-dependencies` to be licensed scanned ([#257](https://github.com/fossas/spectrometer/pull/257))
+- Fixes an issue where stdout doesn't always flush to the console. ([#265](https://github.com/fossas/spectrometer/pull/265))
+- Fixes an issue when referenced-dependencies are not being uploaded. ([#262](https://github.com/fossas/spectrometer/pull/262))
+- Adds support for `fossa-deps.json`. ([#261](https://github.com/fossas/spectrometer/pull/261))
+- Adds support for `vendored-dependencies` to be licensed scanned. ([#257](https://github.com/fossas/spectrometer/pull/257))
```
scruffystuffs, line 68: [nit]: Let's fix this while we're here.
```suggestion
- Adds support for `vendored-dependencies` to be license scanned. ([#257](https://github.com/fossas/spectrometer/pull/257))
```

scruffystuffs (approved):
This looks good.  We can almost automate this now.

PR #315, opened 2021-08-03
doc: Release v2.13.1 and add to changelog

liftM:
Adds the release notes for v2.13.1. I'll tag this commit for release.


liftM:
I'm overriding the CR requirement since this doesn't have any source code in it.

PR #316, opened 2021-08-03
Add origin paths to source unit output

anuccio1:
# Overview

Part of the replacing `srclib` with the CLI includes collecting what's referred to as "Origin paths" for each dependency. This maps to the manifest file in which each dependency was found. This PR adds support for origin paths, and adds a new section named `OriginPaths` to the source unit output.

## NOTE

This strategy isn't 100% accurate, as the *correct* solution would that each dependency would have an origin path, not each SourceUnit, as dependencies within a SourceUnit can have different origin paths for some cases (`gradle` for example). This solution was the quickest, and is going to unblock me in the meantime, however a more correct solution will be considered as well.

A more accurate solution might be a new field on `SourceUnit`, such as:
```
OriginPaths: {
  'foo.nuspec': ['mvn+foo$1.2.3', 'mvn+bar$1.2.3'],
  'project.json': ['mvn+baz$1.2.3']
}
```
in which we map the path to all locators found within that path

## Acceptance criteria

- ran CLI on multiple projects, confirming results

## Testing plan

- run CLI on multiple projects, confirming results

## Risks

none, as this is not used in production at the moment

## References

TODO

## Checklist


src/App/Fossa/ManualDeps.hs
```diff
@@ -88,13 +89,17 @@ toSourceUnit root manualDeps@ManualDependencies{..} maybeApiOpts = do
       referenceLocators = refToLocator <$> referencedDependencies
       additional = toAdditionalData (NE.nonEmpty customDependencies) (NE.nonEmpty remoteDependencies)
       build = toBuildData <$> NE.nonEmpty (referenceLocators <> archiveLocators)
+      originPath = case depsFile of
+        (ManualJSON path) -> toRelativePath root path
+        (ManualYaml path) -> toRelativePath root path
```
anuccio1, line 94: is there a way to do something similar to `(_ path)`? I couldn't figure this one out

src/App/Fossa/Analyze/Project.hs
```diff
@@ -25,14 +26,17 @@ mkResult project graphResults =
...
-    (graph, graphBreadth) = graphResults
+    graph = dependencyGraph dependencyResults
+    relativeManifestFiles = map (toRelativePath basedir) $ dependencyManifestFiles dependencyResults
 
 data ProjectResult = ProjectResult
   { projectResultType :: Text
   , projectResultPath :: Path Abs Dir
   , projectResultGraph :: Graphing Dependency
   , projectResultGraphBreadth :: GraphBreadth
+  , projectResultManifestFiles :: [FilePath]
```
cnr, line 41: the `path` library we use introduced a new type called `SomeBase`:

```hs
data SomeBase t
  = Abs (Path Abs t)
  | Rel (Path Rel t)
```

..which would allow us to more directly specify what's in this field: absolute and relative files.

```suggestion
  , projectResultManifestFiles :: [SomeBase File]
```

src/Path/Extra.hs
```diff
@@ -0,0 +1,16 @@
...
+import Path
+
+-- toRelativePath returns the path of a file (Path Abs File) relative to a directory (Path Abs Dir).
+-- If the file is not within the directory, then the full file path will be returned
+toRelativePath :: Path Abs Dir -> Path Abs File -> FilePath
+toRelativePath baseDir file =
+  fromMaybe (fullPath) $ stripPrefix (fullDir) (fullPath)
+  where
+    fullPath = fromAbsFile file
+    fullDir = fromAbsDir baseDir
```
cnr, line 16: there's [a function called `stripProperPrefix` (doc link)](https://hackage.haskell.org/package/path-0.9.0/docs/Path-Posix.html#v:stripProperPrefix) that's meant for this purpose:

```hs
stripProperPrefix :: MonadThrow m => Path b Dir -> Path b t -> m (Path Rel t)
```

`MonadThrow` has [a bunch of instances (doc link)](https://hackage.haskell.org/package/exceptions-0.10.4/docs/Control-Monad-Catch.html#t:MonadThrow) we can choose from. the one I'll choose for here is `MonadThrow Maybe`.

for the other types I'll choose:
- `b` ~ `Abs`
- `t` ~ `File`

so we get a concrete type:
```hs
stripProperPrefix :: Path Abs Dir -> Path Abs File -> Maybe (Path Rel File)
```

---

given my other comment about using `SomeBase File` instead of `FilePath`, we can change this here too:
```hs
toRelativePath :: Path Abs Dir -> Path Abs File -> SomeBase File
toRelativePath baseDir file = fromMaybe (Abs file) (Rel <$> stripProperPrefix baseDir file)
```

or a `maybe` variant if you prefer:
```hs
toRelativePath :: Path Abs Dir -> Path Abs File -> SomeBase File
toRelativePath baseDir file = maybe (Abs file) Rel (stripProperPrefix baseDir file)
```

also: a name like `tryMakeRelative` might be more descriptive? and add some docs

src/Strategy/Elixir/MixTree.hs
```diff
@@ -66,13 +67,14 @@ import Text.Megaparsec (
...
 import Text.Megaparsec.Char.Lexer qualified as Lexer
-import Types (GraphBreadth (..))
+import Types (DependencyResults (..), GraphBreadth (..))
 
 missingDepVersionsMsg :: Text
 missingDepVersionsMsg = "Some of dependencies versions were not resolved from `mix deps` and `mix deps.tree`. Has `mix deps.get` and `mix compile` been executed?"
 
-analyze :: (Has Exec sig m, Has Diagnostics sig m, Has Logger sig m) => Path Abs Dir -> m (Graphing Dependency, GraphBreadth)
-analyze dir = do
+analyze :: (Has Exec sig m, Has Diagnostics sig m, Has Logger sig m) => MixProject -> m (DependencyResults)
```
cnr, line 75: [nit] there are lots of redundant parentheses in this PR, e.g. here with `(DependencyResults)`

cnr (commented):
General approach seems sound. Left a couple of comments suggesting to use `SomeBase File` instead of `FilePath`

src/Strategy/Haskell/Cabal.hs
```diff
@@ -109,47 +109,47 @@ cabalGenPlanCmd =
...
-        CabalProject
-          { cabalDir = dir
-          }
+  let dotCabalFiles = filter isDotCabal files
+  let cabalDotFiles = filter isCabalDotFile files
+  let allDotFiles = dotCabalFiles ++ cabalDotFiles
 
-  if any isCabalFile files
-    then pure ([project], WalkSkipAll)
+  -- TODO: @wes is one preferred over the other if they are *both* found?
```
anuccio1, line 129: @scruffystuffs - question for you since it looks like you may have built this. If there is both a `*.cabal` file AND a `cabal.project` file, I *assume* we should just be setting the origin path here to the `*.cabal` file (since we specify our deps in the `spectrometer.cabal` file. Wanted to confirm this behavior with you.

For more context, this field will be used to signal to users *where* we found deps (i.e. what manifest file)

src/Path/Extra.hs
anuccio1: @cnr what about: 
```
tryMakeRelative :: Path Abs Dir -> Path Abs File -> SomeBase File
tryMakeRelative absDir absFile =
  case stripProperPrefix absDir absFile of
    Left _ -> Abs absFile
    Right relFile -> Rel relFile
```

cnr: yeah that's equivalent 👍

src/Strategy/Haskell/Cabal.hs
```diff
@@ -109,47 +109,47 @@ cabalGenPlanCmd =
...
 findProjects = walk' $ \dir _ files -> do
-  let project =
-        CabalProject
-          { cabalDir = dir
-          }
-
-  if any isCabalFile files
-    then pure ([project], WalkSkipAll)
-    else pure ([], WalkContinue)
+  -- TODO: @wes is this the right approach?
```
anuccio1, line 125: @scruffystuffs - question for you since it looks like you may have built this. If there is both a *.cabal file AND a cabal.project file, I assume we should just be setting the origin path here to the *.cabal file (since we specify our deps in the spectrometer.cabal file. Wanted to confirm this behavior with you.

For more context, this field will be used to signal to users where we found deps (i.e. what manifest file)

scruffystuffs: I believe that's correct.  @cnr actually knows cabal better than me, so I'd double-check with him instead.

src/App/Fossa/ManualDeps.hs
scruffystuffs: No, this is actually intentional, since the different constructors might have different arities (arity: number of arguments needed).  There's likely a proposal/extension for allowing something like this, but I don't know about it.

src/Strategy/Haskell/Cabal.hs
cnr: it's important that we continue to look for `cabal.project`. cabal supports multi-module projects, where `cabal.project` denotes the root of the project, and subdirectories can contain `.cabal` files

notably the folder containing `cabal.project` is **not required to contain a `whatever.cabal` file**

the short-term less-accurate version here is to use the whole list from `filter isCabalFile` as manifest files in the project

the long-term more-accurate version is parse `cabal.project` and look for relevant cabal files to mark as manifests

src/App/Fossa/ManualDeps.hs
anuccio1: good to know, thanks

src/Strategy/Haskell/Cabal.hs
anuccio1: sounds good, I've updated it to use the short-term less-accurate version

src/Strategy/Rebar3.hs
```diff
@@ -35,10 +35,17 @@ mkProject project =
...
+    , projectDependencyResults = const $ getDeps project
     , projectPath = rebarDir project
     , projectLicenses = pure []
     }
 
-getDeps :: (Has Exec sig m, Has ReadFS sig m, Has Diagnostics sig m) => RebarProject -> m (Graphing Dependency, GraphBreadth)
-getDeps project = Rebar3Tree.analyze' (rebarDir project)
+getDeps :: (Has Exec sig m, Has ReadFS sig m, Has Diagnostics sig m) => RebarProject -> m DependencyResults
+getDeps project = do
+  (graph, graphBreadth) <- Rebar3Tree.analyze' (rebarDir project)
```
cnr, line 45: [nit] can we move the creation of the `graphBreadth` out of `Rebar3Tree.analyze'`? it's always `Complete`
```suggestion
  graph <- Rebar3Tree.analyze' (rebarDir project)
```

src/Strategy/Python/Setuptools.hs
```diff
@@ -42,14 +42,19 @@ findProjects = walk' $ \dir _ files -> do
...
     context "Setuptools" $
       Diag.combineSuccessful
         "Analysis failed for all requirements.txt/setup.py in the project"
         [analyzeReqTxts project, analyzeSetupPy project]
-  pure (graph, Partial)
+  pure $
+    DependencyResults
+      { dependencyGraph = graph
+      , dependencyGraphBreadth = Partial
+      , dependencyManifestFiles = (maybe [] pure (setuptoolsSetupPy project)) ++ setuptoolsReqTxt project
```
cnr, line 56: [nit] this is also known as `maybe [] pure` is also known as `Data.Maybe.maybeToList`
```hs
      , dependencyManifestFiles = maybeToList (setuptoolsSetupPy project) ++ setuptoolsReqTxt project
```

src/Strategy/Python/Poetry.hs
```diff
@@ -111,18 +111,24 @@ analyze ::
...
 analyze PoetryProject{pyProjectToml, poetryLock} = do
   pyproject <- readContentsToml pyProjectCodec (pyProjectTomlPath pyProjectToml)
-  case poetryLock of
+  (graph, graphBreadth, manifestFile) <- case poetryLock of
     Just lockPath -> do
       poetryLockProject <- readContentsToml poetryLockCodec (poetryLockPath lockPath)
       _ <- logIgnoredDeps pyproject (Just poetryLockProject)
       graph <- context "Building dependency graph from pyproject.toml and poetry.lock" $ pure $ setGraphDirectsFromPyproject (graphFromLockFile poetryLockProject) pyproject
-      pure (graph, Complete)
+      pure (graph, Complete, poetryLockPath lockPath)
```
cnr, line 122: [nit] create the `DependencyResults` here directly instead of the 3-tuple

cnr (approved):
a few nits; otherwise LGTM

src/Strategy/Rebar3.hs
anuccio1: If it always complete, then why give developers the option of marking it otherwise?

src/Strategy/Python/Setuptools.hs
anuccio1: Oh amazing, that's good to know

src/Strategy/Rebar3.hs
anuccio1: I'm going to move forward and merge. If you have a strong preference to take `graphBreadth` out, I will come back and change it

cnr: yeah looking back on this it's not important. It just felt weird to have `graphBreadth` be constant here, because we could inline it in the `DependencyResults` creation below

but if we add more tactics here it'll make sense for `Rebar3Tree.analyze'` to return `GraphBreadth`

PR #318, opened 2021-08-04
fix: Improve `mvn dependency:tree` parsing stability

liftM:
# Overview

This PR removes the outdated Maven plugin tactic, and improves the existing `mvn dependency:tree` tactic.

Improvements:

- Output files are now properly cleaned up when exceptions occur during exec or parsing.
- We now attempt to parse output files even in non-zero execs (which is possible due to only some builds in a multi-module build failing).
- We now pass `--fail-at-end` to maximize the number of succeeding builds.
- We now create a single output file in `/tmp`, instead of output in every project folder.
- We now attempt to use `settings.xml` is one is found next to the root `pom.xml`.

## Acceptance criteria

1. Continue parsing on non-zero `mvn dependency:tree` exit.
2. Always clean up temporary files.
3. Try to use `settings.xml`.

## Testing plan

It works on apache/commons-rng, but I'm trying to come up with a more satisfying testing plan here.

## Risks

@cnr has convinced me that parsing `mvn dependency:tree` is strictly superior to the plugin, but removing the plugin is still a big change.

## References

This is a follow up to #310, and part of the resolution of [ZD-2896](https://fossa.zendesk.com/agent/tickets/2896).

## Checklist


src/Strategy/Maven/DepTree.hs
```diff
@@ -123,8 +179,8 @@ data DotGraph = DotGraph
   }
   deriving (Eq, Ord, Show)
 
-parseDotGraph :: Parser DotGraph
-parseDotGraph = do
+parseDotGraphs :: Parser [DotGraph]
+parseDotGraphs = some $ do
```
scruffystuffs, line 183: We should probably unit-test this.

src/Strategy/Maven/DepTree.hs
```diff
@@ -47,27 +42,100 @@ import Text.Megaparsec.Char (space1)
...
-    Just file -> pure ([file], WalkContinue)
+  where
+    execAndParse ::
+      (Has Exec sig m, Has ReadFS sig m, Has Diagnostics sig m) =>
+      Maybe (Path Abs File) ->
+      Path Abs File ->
+      m [DotGraph]
+    execAndParse settingsFile tmpFile = do
+      _ <- context "Running 'mvn dependency:tree'" $ exec dir $ deptreeCmd settingsFile tmpFile
+      context "Parsing 'mvn dependency:tree' output" $ readContentsParser parseDotGraphs tmpFile
```
scruffystuffs, line 133: [optional]: Putting this function inside a `where` block seems weird.  I'd prefer to see it at the top level, but that's just preference.

scruffystuffs (approved):
This looks fine, but I'd really like to see a unit test for multiple graphs in a single file.  Also, notably, the docs are still TODO.

We should hold until we test against more real-world projects.  Like you said, removing the plugin is a big deal.

src/Strategy/Maven/DepTree.hs
liftM: This is intentional to make the `finally` apply to the whole action. I've added a comment explaining this.

liftM:
Yeah okay, I've added the plugin back in. Let's remove the plugin _after_ we've merged integration tests back in.

PR #319, opened 2021-08-04
Use new `shrink` function to filter sourceunit dependencies

cnr:
# Overview

When converting our internal graph representation to a `SourceUnit` we, among other things, [filter out unsupported dependency types](https://github.com/fossas/spectrometer/blob/546826367d9041a5e68e787ef5152014d584ab6c/src/Srclib/Converter.hs#L43)

This is problematic, because some analyzers (looking at you, `Gradle`) produce graphs containing only `SubprojectType` dependencies as direct. When we filter out those dependencies, we're left with a graph without any direct dependencies -- only deep dependencies and edges between them.

`shrink` is a new operation like `filter` that removes vertices, but incoming and outgoing edges from removed vertices are rewritten to preserve the overall graph structure. This means that, e.g., direct deps that are filtered out will have their successors marked as direct after a `shrink`

Unrelatedly, this refactors `Graphing` into its isomorphic representation, `AdjacencyMap (Node a)`, where `data Node a = Root | Node a`. This gains us much more code reuse in general, and reduces the surface area for incorrectly implementing `Graphing` operations. In particular, it was non-trivial to correctly implement `shrink` otherwise. I also refactor the operations of `Graphing` to more-closely mirror those of `AdjacencyMap`

## Acceptance criteria

- SourceUnits generated from Gradle projects contain direct dependencies as appropriate
- `shrink` is semantically sound for multi-node removal, rewiring several incoming/outgoing edges, and rewriting "direct" relationships

## Testing plan

1. Manual testing
  - scan a Gradle project (with at least one dependency) prior to these changes; notice that the resulting `SourceUnit.Build.Imports` array is empty
  - scan a Gradle project (with at least one dependency) after these changes; notice that the resulting `SourceUnit.Build.Imports` array is non-empty
2. Unit tests to validate the properties of `shrink`/`shrinkSingle`

## Risks

`shrink` needs to be semantically sound, because we run it on every SourceUnit we generate in the CLI.

## References

ZD3001 / ZD3006

### DIFF
--- a/src/Graphing.hs
+++ b/src/Graphing.hs
@@ -63,7 +63,7 @@ import Prelude qualified
 -- automatically deduplicated using the 'Ord' instance of @ty@.
 --
 -- Typically, when consuming a Graphing, we only care about nodes in the graph
--- reachable from 'graphingDirect'
+-- reachable from 'directList'
 newtype Graphing ty = Graphing {unGraphing :: AdjacencyMap (Node ty)}
   deriving (Eq, Ord, Show)
 
--- a/src/Graphing.hs
+++ b/src/Graphing.hs
@@ -14,7 +14,7 @@
 module Graphing (
   -- * Graphing type
   Graphing (..),
-  Node(..),
+  Node (..),
   empty,
   size,
   direct,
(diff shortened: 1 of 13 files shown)


src/Graphing.hs
```diff
@@ -46,99 +64,149 @@ import Data.Set qualified as Set
 --
 -- Typically, when consuming a Graphing, we only care about nodes in the graph
 -- reachable from 'graphingDirect'
-data Graphing ty = Graphing
-  { graphingDirect :: Set ty
-  , graphingAdjacent :: AdjacencyMap ty
-  }
+newtype Graphing ty = Graphing {unGraphing :: AdjacencyMap (Node ty)}
```
scruffystuffs, line 67: The `graphingDirect` comment in the haddock no longer makes sense.

scruffystuffs (approved):
One issue with a comment, otherwise looks good.  `shrink` makes sense to me, semantically, and the test confirm the behavior.

giphub

PR #323, opened 2021-08-06
IAT: Register user defined binary

kitified:
# Overview

Support registering user-defined IAT binaries as described in [the IAT RFC](http://go/iat-design-doc).

## Acceptance criteria

We can register the fingerprints of the contents of a directory in the FOSSA server using a set of data provided the user that mirrors the data provided for user-defined custom dependencies.

## Testing plan

1. Run a local instance of `fossa`.
2. Run a command to register binary fingerprints as a user-defined dependency. Example:

```shell
cabal run fossa -- experimental-vsi-register-binary-custom-dependency -e 'http://localhost:9578' --fossa-api-key <key> --name <name> --version <version> --license <license>  <dir containing binaries>
```

3. Verify that the binaries appear in the database.

## Risks

Since this is a totally new subcommand I think this is not very risky.

## Checklist

I don't think that tests are super useful for this; everything that I'd normally test is compiler enforced via types.


src/App/Fossa/IAT/Types.hs
```diff
@@ -0,0 +1,34 @@
...
+
+instance ToJSON (Fingerprint t) where
+  toJSON = toJSON . toText
+
+toText :: Fingerprint t -> Text
+toText (Fingerprint x) = x
+
+-- | Raw describes a fingerprint calculated from the full unmodified content of a file.
+-- Raw fingerprints are implemented as SHA256 hashes.
+data Raw deriving (Typeable)
```
scruffystuffs, line 25: I'm not sure what the purpose of this type is.  I've seen it used in `Fingerprint Raw`, but it's not clear to me why we're using this type

src/App/Fossa/IAT/Types.hs
```diff
@@ -0,0 +1,34 @@
...
+  UserDefinedBinaryAssertion (..),
+) where
+
+import Data.Aeson
+import Data.Text
+import Data.Typeable
+
+-- | Fingerprint uniquely idenfies a file, derived from its content.
+-- Fingerprints are backed by base16 representations of underlying data.
+newtype Fingerprint t = Fingerprint Text
```
scruffystuffs, line 15: Why does this type need a phantom type parameter?

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -452,3 +456,45 @@ uploadContributors apiOpts locator contributors = fossaReq $ do
...
+
+-- | The set of necessary data to describe a user defined binary IAT assertion to the FOSSA API.
+data UserDefinedIATBinaryAssertion = UserDefinedIATBinaryAssertion
+  { userDefinedBinaryAssertionName :: Text
+  , userDefinedBinaryAssertionVersion :: Text
+  , userDefinedBinaryAssertionLicenseIdentifier :: Text
+  , userDefinedBinaryAssertionDescription :: Maybe Text
+  , userDefinedBinaryAssertionUrl :: Maybe Text
+  , userDefinedBinaryAssertionFingerprints :: [IATFingerprintSet]
+  }
```
scruffystuffs, line 480: This name and the prefix are a little heavy.  I'll leave it to you to decide the correct name, but here's an idea:
```suggestion
data BinaryAssertion = BinaryAssertion
  { assertionName :: Text
  , assertionVersion :: Text
  , assertionLicenseId :: Text
  , assertionDescription :: Maybe Text
  , assertionUrl :: Maybe Text
  , assertionFingerprints :: [IATFingerprintSet]
  }
```

You should consider removing the IAT acronym everywhere as well.  I'm not sure it adds to the type names and/or function names.

A lot of the details are part of the name, but I'd rather see them as part of the haddocks (doc comments).

src/App/Fossa/IAT/Fingerprint.hs
```diff
@@ -0,0 +1,39 @@
...
+import Discovery.Walk
+import Effect.ReadFS
+import Path
+
+sha256 :: BL.ByteString -> Digest SHA256
+sha256 = hashlazy
+
+hashFile :: (Has (Lift IO) sig m, Has Diagnostics sig m) => FilePath -> m Text
+hashFile file = do
+  content <- context "Read file" $ sendIO $ BL.readFile file
```
scruffystuffs, line 21: I don't think we should be doing raw, lazy IO in production code.   I think that the `ReadFS` effect needs to be able to produce some sort of streaming type.  @cnr what do you think here?

scruffystuffs (changes requested):
Haven't done a full review yet, but I have a few questions I'd like to block on.

src/App/Fossa/IAT/Types.hs
kitified: My intent was to support other kinds of fingerprints if we want them in the future. I'm happy to avoid the extra complexity for now though, upon thinking on it more.

src/App/Fossa/FossaAPIV1.hs
kitified: I think I generally cleaned up the new identifiers and made them fit more of an overall theme as well!

src/App/Fossa/IAT/Fingerprint.hs
cnr: Yeah frustratingly here, `BL.readFile` isn't a safe operation

For streaming elsewhere we use [conduit](https://github.com/snoyberg/conduit#readme), and it looks like there's a package `conduit-cryptonite` that [offers a hashFile function](https://hackage.haskell.org/package/cryptonite-conduit-0.2.2/docs/Crypto-Hash-Conduit.html#v:hashFile) -- which can be used as `sendIO $ hashFile ...`

The body of that function is [pretty small](https://hackage.haskell.org/package/cryptonite-conduit-0.2.2/docs/src/Crypto-Hash-Conduit.html) if we'd like to just reimplement it as well

The other piece to note here is that errors aren't going to be great, because failures will just throw `IOException` and entirely skip our `Diagnostics` effect. We could resolve that by using `catch` from `Control.Effect.Exception` and catching `IOException` specifically to rethrow with `fatal`:

```hs
hashFile = ...
  `catch` (\(e :: IOException) -> fatalText ("blah blah failed to hash file " <> toText (show e)))
```

src/App/Fossa/IAT/Fingerprint.hs
```diff
@@ -0,0 +1,39 @@
...
+fingerprintRaw file = do
+  fp <- context "Fingerprint file" $ hashFile (toFilePath file)
+  pure (Fingerprint fp)
+
+fingerprintRawMany :: (Has (Lift IO) sig m, Has Diagnostics sig m) => [Path Abs File] -> m [Fingerprint]
+fingerprintRawMany [] = pure []
+fingerprintRawMany (x : xs) = do
+  fp <- fingerprintRaw x
+  fps <- fingerprintRawMany xs
+  pure (fp : fps)
```
cnr, line 34: [nit] This is good. Quick nit to simplify it: you've unknowingly reimplemented `traverse` on lists here.

```hs
traverse :: (Traversable t, Applicative f) => (a -> f b) -> t a -> f (t b)

-- or under the list instance
traverse :: Applicative f => (a -> f b) -> [a] -> f [b]
```

`traverse` is essentially "run an effectful function over everything in the structure, collecting results back into the structure". It's kind of like a specialized for loop in imperative languages, and it turns out that [`for` is another provided function](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-Traversable.html#v:for) that's just `traverse` with its arguments reversed. Makes it convenient to do things with a `do` block like:

```hs
example = for someListOfThings $ \oneElement -> do
  foo
  pure bar
```

---

### Intuition-building for `traverse`

1. Note that we get to choose the effect (anything with an `Applicative` instance -- and any `Monad` necessarily has an `Applicative` instance)

The `Maybe` monad/applicative has short-circuiting behavior, so we can use it for, e.g., data validation:
```hs
data Unvalidated = ...
data Validated = ...

validateSingle :: Unvalidated -> Maybe Validated
validateSingle = ...

-- validate everything in the list, returning `Nothing` if any elements are invalid
validateAll :: [Unvalidated] -> Maybe [Validated]
validateAll = traverse validateSingle
```

Similarly the `Either e` monad has short-circuiting behavior, but returns the first error it encounters:

```hs
data ValidationError = ...
data Unvalidated = ...
data Validated = ...

validateSingle :: Unvalidated -> Either ValidationError Validated
validateSingle = ...

-- validate everything in the list, returning `Nothing` if any elements are invalid
validateAll :: [Unvalidated] -> Either ValidationError [Validated]
validateAll = traverse validateSingle
```

and there's another `Applicative` (but not `Monad`) [called `Validation`](https://hackage.haskell.org/package/either-5.0.1.1/docs/Data-Either-Validation.html) that would return _all_ of the validation errors, and not just the first one

2. Traverse is essentially an `fmap` followed by `sequence`

```hs
-- `fmap f` gives us `t (f a)` -- e.g., `fmap putStrLn [1..5] :: [IO ()]`
--
traverse :: (Traversable t, Applicative f) => (a -> f b) -> t a -> f (t b)
traverse f = sequence . fmap f

-- "swap" the position of t and f
sequence :: (Traversable t, Applicative f) => t (f a) -> f (t a)
sequence = ...
```

3. It's common to not care about the results of the effectful action, like in a typical imperative language for loop

For this reason, there are `traverse_` and `for_` variants that "ignore the results":
```hs
traverse_ :: (Foldable t, Applicative f) => (a -> f b) -> t a -> f ()
for_ :: (Foldable t, Applicative f) => t a -> (a -> f b) -> f ()

-- e.g.,

```hs
main :: IO ()
main = traverse_ print [1..100]
```

Changelog.md
```diff
@@ -1,5 +1,10 @@
 # Spectrometer Changelog
 
+## Unreleased
+
+- Add new hidden subcommand `experimental-vsi-register-binary-custom-dependency`.
+  This subcommand is intended to be used with direct support from FOSSA engineering when working with specific projects.
```
scruffystuffs, line 6: Not 100% sure on this, but I think that if we don't want people using things, we should probably leave them out of the changelog.

kitified: Fair enough, it just seemed odd to have a release with "nothing" in it (if I end up doing a release with just my changes). Happy to go either way.

scruffystuffs: Yeah, I'm really not 100% on this.  Let's leave it in.  Also make sure to add the PR numbers to the changelog.

src/App/Fossa/IAT/Fingerprint.hs
kitified: This was super helpful- and also gives me the building block for the future. I was trying to use just `fmap` and was running into the issue where the monad was inside the result (`[IO ()]`) and couldn't figure out how to get it out, so fell back to recursion- now I know `sequence` is what gets it out, and to do both it's just `traverse`. Thanks!

src/App/Fossa/IAT/Fingerprint.hs
kitified: @cnr 

> The body of that function is pretty small if we'd like to just reimplement it as well

What is the typical practice within Spectrometer? I'm happy to go either route.
Also the library you're referring to is [this one](https://hackage.haskell.org/package/cryptonite-conduit), right?

cnr: I'd re-implement it rather than pulling in the dep; the function is really small. And yes, that's the one

src/App/Fossa/IAT/Fingerprint.hs
```diff
@@ -0,0 +1,51 @@
...
+
+fingerprintRaw :: (Has (Lift IO) sig m, Has Diagnostics sig m) => Path Abs File -> m Fingerprint
+fingerprintRaw file = do
+  fp <- hashFile' $ toFilePath file
+  let base16 = toText . show $ fp
+  pure (Fingerprint base16)
+  where
+    -- Monomorphize to SHA256 hash
+    hashFile' :: (Has (Lift IO) sig m, Has Diagnostics sig m) => FilePath -> m (Digest SHA256)
+    hashFile' = hashFile
```
cnr, line 43: [nit] we can eliminate this given either:

1. monomorphizing `hashFile` to return `Digest SHA256`
2. using a type annotation on `fp` above (`(fp :: Digest SHA256) <- hashFile ...`)

src/App/Fossa/IAT/Fingerprint.hs
```diff
@@ -0,0 +1,51 @@
...
+
+fingerprintRawMany :: (Has (Lift IO) sig m, Has Diagnostics sig m) => [Path Abs File] -> m [Fingerprint]
+fingerprintRawMany = traverse fingerprintRaw
+
+fingerprintContentsRaw :: (Has ReadFS sig m, Has Diagnostics sig m, Has (Lift IO) sig m) => Path Abs Dir -> m [Fingerprint]
+fingerprintContentsRaw = walk' $ \_ _ files -> do
+  fps <- fingerprintRawMany files
```
cnr, line 50: [nit] maybe inline `fingerprintRawMany`?
```suggestion
  fps <- traverse fingerprintRaw files
```

src/App/Fossa/IAT/Types.hs
```diff
@@ -0,0 +1,26 @@
...
+  Fingerprint (..),
+  UserDefinedAssertionMeta (..),
+) where
+
+import Data.Aeson
+import Data.Text
+
+-- | Fingerprint uniquely idenfies a file, derived from its content.
+-- Fingerprints are backed by base16 representations of underlying data.
+newtype Fingerprint = Fingerprint Text
```
cnr, line 11: [nit] we often add a record selector to wrapper newtypes, which would replace the need for `toText` below:
```suggestion
newtype Fingerprint = Fingerprint { unFingerprint :: Text }
```

and now we can use `unFingerprint` everywhere we were using `toText` before

Changelog.md
```diff
@@ -1,5 +1,10 @@
 # Spectrometer Changelog
 
+## Unreleased
+
+- Add new hidden subcommand `experimental-vsi-register-binary-custom-dependency`. ([#323](https://github.com/fossas/spectrometer/pull/323))
```
cnr, line 5:
```suggestion
- Adds a new hidden subcommand `experimental-vsi-register-binary-custom-dependency`. ([#323](https://github.com/fossas/spectrometer/pull/323))
```

cnr (approved):
LGTM! left some nits

spectrometer.cabal
```diff
@@ -142,6 +142,9 @@ library
     App.Fossa.Container
     App.Fossa.Container.Analyze
     App.Fossa.Container.Test
+    App.Fossa.IAT.AssertUserDefinedBinaries
```
meghfossa, line 145: [optional] [minor] Place`App.Fossa.IAT*` below `App.Fossa.FossaAPIV1` to preserve alphabetical order.

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -24,9 +24,11 @@ module App.Fossa.FossaAPIV1 (
   getSignedURL,
   archiveUpload,
   archiveBuildUpload,
+  assertUserDefinedBinaries,
 ) where
 
 import App.Fossa.Container (ContainerScan (..))
+import App.Fossa.IAT.Types
```
meghfossa, line 31: [nit] We are moving towards, [not having implicit blanket imports](https://github.com/fossas/spectrometer/blob/style-guide/STYLE-GUIDE.md#dont-use-implicit-blanket-imports). 

```suggestion
import App.Fossa.IAT.Types (Fingerprint (..), UserDefinedAssertionMeta(..))
```

meghfossa (approved):
I've left few nits, but other than that, LGTM. Prefer if you can avoid implicit imports on new code. 

Style guide is not merged in yet, but here is the relevant section in regards to the imports https://github.com/fossas/spectrometer/blob/style-guide/STYLE-GUIDE.md#dont-use-implicit-blanket-imports

PR #325, opened 2021-08-09
Style guide

scruffystuffs:
# Overview

Add a style guide to the repo.  Also add some lints to `.hlint.yaml`

## Acceptance criteria

Style guide is agreeable for the team.

## References

Closes fossas/team-analysis#671

## Checklist

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -9,6 +9,8 @@
   - {name: Data.Map, as: Map, message: "Use complete name for qualified import of Map"}
   - {name: Data.Map.Strict, as: Map, message: "Use complete name for qualified import of Strict Map"}
   - {name: Data.Text, as: Text, message: "Use complete name for qualified import of Text"}
+  # Don't allow unsafe IO
+  - {name: System.Unsafe.IO, within: []}
 
 # Styling Preferences
 - suggest: {lhs: "asum [x, y]", rhs: x <|> y}
@@ -17,3 +19,8 @@
 - suggest: {lhs: x <&> f, rhs: f <$> x}
 - suggest: {lhs: Data.Text.pack x, rhs: Data.String.Conversion.toText x}
 - suggest: {lhs: Data.Text.unpack x, rhs: Data.String.Conversion.toString x}
+
+# Forbidden items
+- functions:
+  - {name: error, within: [Data.String.Conversion]}
+  - {name: undefined, within: []}
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -10,7 +10,7 @@
   - {name: Data.Map.Strict, as: Map, message: "Use complete name for qualified import of Strict Map"}
   - {name: Data.Text, as: Text, message: "Use complete name for qualified import of Text"}
   # Don't allow unsafe IO
-  - {name: System.Unsafe.IO, within: []}
+  - {name: System.IO.Unsafe, within: []}
 
 # Styling Preferences
 - suggest: {lhs: "asum [x, y]", rhs: x <|> y}
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -1,0 +1,186 @@
+# Spectrometer style guide
+
+Through a number of pull requests, reviews, and discussions, a codebase style has emerged.  This document is an attempt to capture that, and prevent further discussions that slow down work unnecessarily.
+
+This style guide is forked from Github's [semantic](https://github.com/github/semantic).
+
+## General guidelines
+
+Make your code look like the code around it. Consistency is the name of the game.
+
+Use `fourmolu` for formatting.  Our CI setup enforces that all changes pass a `fourmolu` run with no differences.
+
+Our CI systems ensure that all patches pass `hlint`'s muster. We have our own set of rules in `.hlint.yaml`.
+
+We strongly recommend adding Haddock documentation to any function/data type, unless its purpose is immediately apparent from its name.
+Comments should describe the "why", type signatures should describe the "what", and the code should describe the "how".
+
+## Formatting
+
+2 spaces everywhere. Tabs are forbidden. Haskell indentation can be unpredictable, so generally stick with what your editor suggests.
+There is no hard line-length limit, though if you go beyond 110 or 120 you should generally split it up, especially for type signatures.
+
+### Use applicative notation when constructing simple data types
+
+``` haskell
+thing :: Parser Foo
+
+-- Broke:
+thing = do
+  a <- bar
+  b <- baz
+  pure (Foo a b)
+
+-- Woke:
+thing = Foo <$> bar <*> baz
+```
+
+Code written with applicative notation is less powerful than code written in `do`-notation.  As a result, it can be easier to reason about, as long as it's also readable.
+
+Overreliance on applicative notation can create code that is difficult to read. Don't use applicative notation in combination with operator sections. If in doubt, write it with `do` notation and see if it's more immediately comprehensible.
+
+### Use leading commas for records, exports, and lists
+
+Leading commas make it easy to add and remove fields without introducing syntax errors, and properly aligned records are easy to read:
+
+``` haskell
+data Pos = Pos
+  { posLine   :: Int
+  , posColumn :: Int
+  }
+```
+
+This rule is enforced by CI.
+
+## Naming
+
+Locally bound variables (such as the arguments to functions, or helpers defined in a `where` clause) can have short names, such as `x` or `go`. Globally bound functions and variables should have descriptive names.
+
+You'll often find yourself implementing functions that conflict with Prelude definitions. If this is the case, avoid adding a prefix to these functions, and instead import them qualified.
+
+``` haskell
+-- Bad
+foo = heapLookup thing
+-- Better
+foo = Heap.lookup thing
+```
+
+## Functions
+
+### Don't go buckwild with infix operators
+
+Sensible use of infix operators can provide serious readability benefits, but often the best tool is just a named function. If you're defining new operators, make sure that you have a solid justification for doing so.
+
+### Avoid list comprehensions
+
+In almost all cases, `map`, `filter`, `fold`, and the `[]` monad are more flexible and readable.
+
+### Don't go buckwild with point-free definitions
+
+Point-free style can help or hinder readability, depending on the context. If a function is expressed naturally with the `.` operator, then do so, but if you have to gyrate the definition to write it point-free, then you should probably just write out the variable names. If you are reviewing someone else's PR and find a point-free definition hard to read, ask them to simplify/clarify it.
+
+### Prefer `.` and `$` to parentheses
+
+Parentheses can make a function harder to edit, since parentheses have to be balanced. The composition and application operators (`.` and `$`) can reduce clunkiness.
+
+``` haskell
+-- Bad
+f (g (h x))
+-- Better
+f $ g $ h x
+-- Hlint will reject this
+f . g . h $ x
+```
+
+Parentheses are useful when there are other operators in the mix, since they don't have conflicting fixity like `$` and `.` have.  `<>` doesn't play well with
+`.` and `$`, for example.  Use your best judgement here.
+
+### Do not use partial functions
+
+Never use a partial function when a safe version will work instead.  Do not use `error` or `undefined`.
+
+In some cases, partial functions are necessary, like indexing lists.  In this case, you must prove the safety of the function before using, ideally with an
+accompanying comment explaining the safety.
+
+### Prefer `map` to `fmap` or `<$>`
+
+When operating on a list using `map` tells the reader "I'm transforming a list", where `fmap` tells them "I'm transforming *some* functor, and it doesn't matter which one.
+
+```haskell
+appendFoo t = t <> "-foo"
+txts = ["Hello", "darkness", "my", "old", "friend"]
+
+-- Ambiguous
+appendFoo <$> 
+```
+
+## Data Types
+
+### Prefer `newtype`s to `type`s
+
+`newtype` values are zero-cost to construct and eliminate, and provide more informative error messages than `type` synonyms. Only use `type` for convenience aliases to existing types.
+
+### Don't use `String`
+
+`String` is almost always the wrong choice. If your type represents human-readable strings, use `Text`; if you have a blob of bytes, use `ByteString`. `-XOverloadedStrings` is enabled globally to make this easy.
+
+### Only use record selectors on single-constructor types
+
+The following code generates two partial functions, which is bad:
+
+``` haskell
+data Bad = Evil { getInt :: Int }
+         | Bad  { getFloat :: Float }
+
+-- getFloat (Evil 1) throws an error
+-- so does getInt (Bad 1)
+```
+
+Often, this is incorrect deisgn.  You usually want something like the following three examples instead:
+
+```haskell
+-- Alternative 1
+data Good
+  = Okay Int
+  | NotBad Float
+
+-- Alternative 2
+data Alright = Alright
+  { getInt :: Maybe Int
+  , getFloat :: Maybe Float
+  }
+
+-- Alternative 3
+-- This is a little bit heavyweight, but is really strongly typed.
+newtype SpecialInt = SpecialInt { getInt :: Int }
+newtype SpecialFloat = SpecialFloat { getFloat :: Float }
+data Great
+  = NiceInt SpecialInt
+  | CoolFloat SpecialFloat
+```
+
+## Imports
+
+### Don't qualify imports with single letters
+
+Qualified imports deserve intelligent names, like any other variable.  Some common instances are caught by `hlint`:
+
+```haskell
+import Data.Text qualified as T -- bad; hlint will catch this specific case
+import Data.Text qualified as Text -- Good!
+```
+
+### Don't use implicit blanket imports
+
+Import items explicitly:
+
+```haskell
+import Control.Effect.Diagnostics -- bad, which items do you need from this module?
+import Control.Effect.Diagnostics (fatal, context)  -- Good
+```
+
+There are notable exceptions, but they are few and far between.  The `Path` module is a common exception, but there's no good justification for this.
+
+## Miscellaneous
+
+* Don't use `{-# ANN … #-}` to disable hlint warnings, as it can slow down compilation. If you need to disable lints in a file, do so in `.hlint.yaml`.
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -12,7 +12,7 @@ Use `fourmolu` for formatting.  Our CI setup enforces that all changes pass a `f
 
 Our CI systems ensure that all patches pass `hlint`'s muster. We have our own set of rules in `.hlint.yaml`.
 
-We strongly recommend adding Haddock documentation to any function/data type, unless its purpose is immediately apparent from its name.
+We strongly recommend adding Haddock documentation to any top-level declaration, unless its purpose is immediately apparent from its name.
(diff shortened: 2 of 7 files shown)


meghfossa:
[Minor] Not sure if the style guide is the right place for this, but I suggest we add bullet point for:

- Use of logDebug, logWarn, and it's preference over context
- Preference of not parsing metadata or irrelevant content if it's not to be included in the final dependency graph

scruffystuffs:
I'm not sure if those are really "Style guide" points.  For the most part, everything written is for people who got information from outside of the codebase (like the idea to use `&`), and instructing on whether to use that knowledge here.

scruffystuffs:
I think that instead, that kind of thing belongs in development docs, which are either in HACKING or don't really exist.

STYLE-GUIDE.md
```diff
@@ -0,0 +1,285 @@
...
+
+## General guidelines
+
+Make your code look like the code around it. Consistency is the name of the game.
+
+Use `fourmolu` for formatting.  Our CI setup enforces that all changes pass a `fourmolu` run with no differences.
+
+Our CI systems ensure that all patches pass `hlint`'s muster. We have our own set of rules in `.hlint.yaml`.
+
+We strongly recommend adding Haddock documentation to any function/data type, unless its purpose is immediately apparent from its name.
```
liftM, line 15:
```suggestion
We strongly recommend adding Haddock documentation to any top-level declaration, unless its purpose is immediately apparent from its name.
```

STYLE-GUIDE.md
```diff
@@ -0,0 +1,285 @@
...
+Comments should describe the "why", type signatures should describe the "what", and the code should describe the "how".
+
+## Formatting
+
+2 spaces everywhere. Tabs are forbidden. Haskell indentation can be unpredictable, so generally stick with what your editor suggests.
+There is no hard line-length limit, though if you go beyond 110 or 120 you should generally split it up, especially for type signatures.
```
liftM, line 21: I propose that comments should wrap at 80 (except obvious exceptions like long URLs) and lines should try to wrap at 80, but not at the expense of readability. This makes split-view code review easier on normal laptop screen sizes.

STYLE-GUIDE.md
```diff
@@ -0,0 +1,285 @@
...
+``` haskell
+thing :: Parser Foo
+
+-- Broke:
+thing = do
+  a <- bar
+  b <- baz
+  pure (Foo a b)
+
+-- Woke:
```
liftM, line 34: Broke/woke is hilarious, but I might switch this to bad/good for the sake of non-native/non-dank readers.

STYLE-GUIDE.md
```diff
@@ -0,0 +1,285 @@
...
+
+Sensible use of infix operators can provide serious readability benefits, but often the best tool is just a named function. If you're defining new operators, make sure that you have a solid justification for doing so.
+
+### Avoid list comprehensions
+
+In almost all cases, `map`, `filter`, `fold`, and the `[]` monad are more flexible and readable.
+
+### Don't go buckwild with point-free definitions
+
+Point-free style can help or hinder readability, depending on the context. If a function is expressed naturally with the `.` operator, then do so, but if you have to gyrate the definition to write it point-free, then you should probably just write out the variable names. If you are reviewing someone else's PR and find a point-free definition hard to read, ask them to simplify/clarify it.
```
liftM, line 80:
```suggestion
Point-free style can help or hinder readability, depending on the context. If a function is expressed naturally with the `.` operator, then do so, but if you have to over-complicate the definition to write it point-free, then you should probably just write out the variable names. If you are reviewing someone else's PR and find a point-free definition hard to read, ask them to simplify/clarify it.
```

STYLE-GUIDE.md
```diff
@@ -0,0 +1,285 @@
...
+``` haskell
+-- Bad
+foo = heapLookup thing
+-- Better
+foo = Heap.lookup thing
+```
+
+## Functions
+
+### Don't go buckwild with infix operators
```
liftM, line 70:
```suggestion
### Don't go crazy with infix operators
```

STYLE-GUIDE.md
```diff
@@ -0,0 +1,285 @@
...
+
+Sensible use of infix operators can provide serious readability benefits, but often the best tool is just a named function. If you're defining new operators, make sure that you have a solid justification for doing so.
+
+### Avoid list comprehensions
+
+In almost all cases, `map`, `filter`, `fold`, and the `[]` monad are more flexible and readable.
+
+### Don't go buckwild with point-free definitions
```
liftM, line 78:
```suggestion
### Don't go crazy with point-free definitions
```

STYLE-GUIDE.md
```diff
@@ -0,0 +1,285 @@
...
+### Don't use implicit blanket imports
+
+Import items explicitly:
+
+```haskell
+import Control.Effect.Diagnostics -- bad, which items do you need from this module?
+import Control.Effect.Diagnostics (fatal, context)  -- Good
+```
+
+There are notable exceptions, but they are few and far between.  The `Path` module is a common exception, but there's no good justification for this.
```
liftM, line 202: > The `Path` module is a common exception, but there's no good justification for this.

Does this mean "`Path` is a good exception" or "`Path` is a bad exception and we haven't gotten around to fixing it?". This sentence could read either way.

STYLE-GUIDE.md
```diff
@@ -0,0 +1,285 @@
...
+- `lexeme` - Parser for any basic unit of the language.  While `symbol` is for verbatim text parsing, `lexeme` is used with any parser that should consume space
+  after finishing.  For example, `Lexer.symbol` is implmented via `Lexer.lexeme`.
+  - Use [`Lexer.lexeme`](https://hackage.haskell.org/package/megaparsec-9.1.0/docs/Text-Megaparsec-Char-Lexer.html#v:lexeme) to define this helper.
+  - Use your `sc` or `scn` function for the whitespace consumer.
+
+### Every parser must consume something to succeed
+
+This is not a style concern, but a parsing rule.
+
+An example file should look somthing like this:
```
liftM, line 242:
```suggestion
An example file should look something like this:
```

STYLE-GUIDE.md
```diff
@@ -0,0 +1,285 @@
...
+  - Use your `sc` or `scn` function for the whitespace consumer.
+  - You can use `Lexer.symbol'` if your text is case-insensitive.
+- `lexeme` - Parser for any basic unit of the language.  While `symbol` is for verbatim text parsing, `lexeme` is used with any parser that should consume space
+  after finishing.  For example, `Lexer.symbol` is implmented via `Lexer.lexeme`.
+  - Use [`Lexer.lexeme`](https://hackage.haskell.org/package/megaparsec-9.1.0/docs/Text-Megaparsec-Char-Lexer.html#v:lexeme) to define this helper.
+  - Use your `sc` or `scn` function for the whitespace consumer.
+
+### Every parser must consume something to succeed
+
+This is not a style concern, but a parsing rule.
```
liftM, line 240:
```suggestion
This is not a style concern, but a common bug when writing parsers.
```

STYLE-GUIDE.md
```diff
@@ -0,0 +1,285 @@
...
+- `symbol` - Parser for verbatim text strings.
+  - Use [`Lexer.symbol`](https://hackage.haskell.org/package/megaparsec-9.1.0/docs/Text-Megaparsec-Char-Lexer.html#v:symbol) to create this helper.
+  - Use your `sc` or `scn` function for the whitespace consumer.
+  - You can use `Lexer.symbol'` if your text is case-insensitive.
+- `lexeme` - Parser for any basic unit of the language.  While `symbol` is for verbatim text parsing, `lexeme` is used with any parser that should consume space
+  after finishing.  For example, `Lexer.symbol` is implmented via `Lexer.lexeme`.
+  - Use [`Lexer.lexeme`](https://hackage.haskell.org/package/megaparsec-9.1.0/docs/Text-Megaparsec-Char-Lexer.html#v:lexeme) to define this helper.
+  - Use your `sc` or `scn` function for the whitespace consumer.
+
+### Every parser must consume something to succeed
```
liftM, line 238: Another tip is that parsers should generally also force `eof` at the end, to make sure they've consumed all input.

liftM (commented):
Some edits for clarity, but mostly LGTM.

STYLE-GUIDE.md
scruffystuffs: Actually, I'm thinking we leave this section and that rule out of this guide.  It has nothing to do with style, and is more about "How to write good spectrometer code".  Maybe we should add this and other relevant stuff to `HACKING`.

.hlint.yaml
```diff
@@ -9,11 +9,18 @@
...
 - suggest: {lhs: return x, rhs: pure x}
 - suggest: {lhs: () <$ x, rhs: Control.Monad.void x}
 - suggest: {lhs: x <&> f, rhs: f <$> x}
 - suggest: {lhs: Data.Text.pack x, rhs: Data.String.Conversion.toText x}
-- suggest: {lhs: Data.Text.unpack x, rhs: Data.String.Conversion.toString x}
\ No newline at end of file
+- suggest: {lhs: Data.Text.unpack x, rhs: Data.String.Conversion.toString x}
+
+# Forbidden items
+- functions:
```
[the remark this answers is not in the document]
scruffystuffs: I think hlint might already do this for us automatically.  I'd need to go investigate.

STYLE-GUIDE.md
scruffystuffs: I don't love marking off at 80, specifically.  I prefer 100/120, since 80 actually wastes a decent amount of space on my monitors.

In reality, not making a line limit is really **my** preferred option.  The formatter will do a lot of work for us here, and beyond that we can adjust it in code review.  As long as we tell people to write vertically instead of horizontally, we are probably fine.  I certainly don't want to set a 80-char limit without significant agreement from the team (i.e. majority believe that 80 is the *best* option, rather than just *an* option).

STYLE-GUIDE.md
scruffystuffs: Leftovers from the original guide.  I'll fix it.

STYLE-GUIDE.md
scruffystuffs: It means "It's an exception we make regularly, but we don't have a good reason to make it an exception."  I'm wide open to rephrasing suggestions here, but that seems pretty clear to me.  Not sure how I would improve it.

.hlint.yaml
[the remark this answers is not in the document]
scruffystuffs: `hlint` doesn't automatically do this, but doing it here would explode this PR (we use head at least 10 times).  I'll ticket that separately.

STYLE-GUIDE.md
cnr: Agreed on moving it out of the style guide and into a separate "parsing best practices" doc or similar

.hlint.yaml
```diff
@@ -9,11 +9,18 @@
   - {name: Data.Map, as: Map, message: "Use complete name for qualified import of Map"}
   - {name: Data.Map.Strict, as: Map, message: "Use complete name for qualified import of Strict Map"}
   - {name: Data.Text, as: Text, message: "Use complete name for qualified import of Text"}
+  # Don't allow unsafe IO
+  - {name: System.Unsafe.IO, within: []}
```
cnr, line 13: https://hackage.haskell.org/package/base-4.15.0.0/docs/System-IO-Unsafe.html
```suggestion
  - {name: System.IO.Unsafe, within: []}
```

STYLE-GUIDE.md
```diff
@@ -0,0 +1,285 @@
...
+
+-- Woke:
+thing = Foo <$> bar <*> baz
+```
+
+Code written with applicative notation is less powerful than code written in `do`-notation.  As a result, it can be easier to reason about, as long as it's also readable.
+
+Overreliance on applicative notation can create code that is difficult to read. Don't use applicative notation in combination with operator sections. If in doubt, write it with `do` notation and see if it's more immediately comprehensible.
+
+### Use leading commas for records, exports, and lists
```
cnr, line 42: I'm conflicted on including some of these sections because the formatter enforces this behavior. The same is true with the section mentioning "always use 2 spaces"

STYLE-GUIDE.md
```diff
@@ -0,0 +1,285 @@
...
+foo = Heap.lookup thing
+```
+
+## Functions
+
+### Don't go crazy with infix operators
+
+Sensible use of infix operators can provide serious readability benefits, but often the best tool is just a named function. If you're defining new operators, make sure that you have a solid justification for doing so.
+
+### Avoid list comprehensions
```
cnr, line 74: 😢

STYLE-GUIDE.md
```diff
@@ -0,0 +1,285 @@
...
+### Prefer `.` and `$` to parentheses
+
+Parentheses can make a function harder to edit, since parentheses have to be balanced. The composition and application operators (`.` and `$`) can reduce clunkiness.
+
+``` haskell
+-- Bad
+f (g (h x))
+-- Better
+f $ g $ h x
+-- Hlint will reject this
```
cnr, line 91: I'm confused at this comment/example

cnr (approved):
Looks good generally, and it's a good starting point to build from

STYLE-GUIDE.md
scruffystuffs: In my mind, it's just reinforcement of those rules.  I added bits to show that they were enforce by the formatter.

STYLE-GUIDE.md
scruffystuffs: Yeah, I know you like them, but I think they're impossible to read most of the time.  This also comes directly from the original guide from `semantic`.

STYLE-GUIDE.md
scruffystuffs: I re-read it, not sure why you're confused.  Can you elaborate?

STYLE-GUIDE.md
scruffystuffs: Moved.

STYLE-GUIDE.md
cnr: I'm confused why hlint will reject it, or why we're showing this example after "Better". If the style we prefer is the mix of `.` and `$` -- e.g.  `f . g . h $ i x` -- I'd expect to see that example here instead

scruffystuffs: Ah, yeah, that makes sense.  I'll rework it.

scruffystuffs: @cnr I reworked the example, is that better?

STYLE-GUIDE.md
zlav: I don't personally care what the line length is because I use wordwrap since a lot of our codebases are inconsistent but I do think we should set a hard limit. Whether it's 80, 100, 120, or nothing I think we should enforce it. I hate asking myself "is this ok?" if I have a long line and then having it be based on judgment.

STYLE-GUIDE.md
```diff
@@ -0,0 +1,237 @@
...
+  b <- baz
+  pure (Foo a b)
+
+-- Better:
+thing = Foo <$> bar <*> baz
+```
+
+Code written with applicative notation is less powerful than code written in `do`-notation (since that code is, by definition, monadic).  As a result, it can be easier to reason about, as long as it's also readable.
+
+Over-reliance on applicative notation can create code that is difficult to read. Don't use applicative notation in combination with operator sections. If in doubt, write it with `do` notation and see if it's more immediately comprehensible.
```
zlav, line 44: >Don't use applicative notation in combination with operator sections

Can you include an example for this?

STYLE-GUIDE.md
```diff
@@ -0,0 +1,237 @@
...
+```
+
+## Imports
+
+### Don't qualify imports with single letters
+
+Qualified imports deserve intelligent names, like any other variable.  Some common instances are caught by `hlint`:
+
+```haskell
+import Data.Text qualified as T -- bad; hlint will catch this specific case
```
zlav, line 207: I thought we said we were fine with certain imports like Text and ByteString which have widely accepted short qualified imports being left alone? `T`, `BS`, and some others. I'm ok with changing what we said, I just want to make sure that we're no longer ok with short imports for `Text` or `Bytestring`.

STYLE-GUIDE.md
zlav: If there isn't a great reason to make an exception I don't think we should. Is it overkill to maintain a list of the notable exceptions? I think that would help clear up ambiguity and I don't think there are too many imports that we would have to list.

STYLE-GUIDE.md
scruffystuffs: I think if you're debating whether it's okay, you should split it up.  The nice thing is that you can split it up yourself, and the formatter will handle the alignment and exact separation.  What the formatter seem to NOT do is decide when to split things up.  This is where judgement comes in.  Our tool isn't going to do this for us, so we just need to do it ourselves, and in that case, there's no benefit to setting a hard limit.

That said, I'm OK with a set of soft limits, for example:
* Write code vertically, rather than horizontally
* If it's trivial, stay under 80
* If it's reasonable, stay under 100
* If it's possible, stay under 120
* Don't go past 120 for anything except long literal strings without spaces, like URL's.
  * Prefer the cost of joining in memory over the readability problems of lines longer than 120.

STYLE-GUIDE.md
scruffystuffs: Yeah, let's remove this note.  I'd like to check this statically, but I'm not sure if it's a GHC warning or hlint rule, or neither.

STYLE-GUIDE.md
scruffystuffs: No, this should be reworded so that an example isn't necessary.

STYLE-GUIDE.md
scruffystuffs: I was told that we decided on names rather than letters.  It was decided during a retro that I was not present for, and we have already begun checking it with `hlint`

zlav: Ahhh, gotcha, yea we decided to use full words except for a few common exceptions like `ByteString`. @cnr do you have any input on this?

cnr: exceptions might be nice for:

- `ByteString` - `BS`
- lazy `ByteString` - `BL`

we don't really use lazy `Text` anywhere -- otherwise I'd want `T` and `TL` for those

STYLE-GUIDE.md
liftM: I think we should have just 1 rule, otherwise it just consumes way too much brainpower. `go fmt` has the right idea here.

The reason I think we should set the ruler at 80 is that 80 characters is the size of a half-page in the split-view code review panel on a 1920x1680 monitor (the standard size for most monitors and laptops). Otherwise, code lines will wrap for 1920x1680s (e.g. Macbooks).

STYLE-GUIDE.md
cnr: I'd kill the last example that hlint rejects -- I think it just adds confusion

STYLE-GUIDE.md
cnr: ghc's `-Wmissing-import-lists` will do this

liftM: Why _do_ we make `Path` an exception? Let's just fix it up. Is there a technical reason why that's difficult, or is it just something we haven't done yet?

cnr: Just something we haven't done. Someone needs to submit a PR that enables `-Wmissing-import-lists`, and do the manual work of adding explicit imports to every module (through the code action in haskell-language-server, probably)

PR #328, opened 2021-08-13
IAT: Resolve user-defined assertions

kitified:
# Overview

Resolves user-defined assertions matched via the VSI process into user dependencies.

## Acceptance criteria

* Results without IAT assertions are untouched.
* Results with IAT assertions are converted into `user` dependencies and uploaded.

## Testing plan

1. I registered a user-defined dependency via `cabal run fossa -- experimental-vsi-register-binary-custom-dependency -e '<endpoint>' --fossa-api-key <key> --name <name> --version <version> --license 'BSD-3-Clause' <dir>`
2. I put that file in the `quick` directory structure (from http://go/vps-demo-projects).
3. I then ran a scan on that directory in staging ([here](https://staging.int.fossa.io/projects/custom%2B1%2Fquick-vsi-test-assertion/refs/branch/master/with-assertion-bsd3clause/browse/dependencies)).
4. The result looks as expected.

## Risks

This does mess with some core pieces of the `Analyze` function.

## Checklist


src/App/Fossa/Analyze.hs
```diff
@@ -371,6 +381,27 @@ filterProjects rootDir = filter (isProductionPath . dropPrefix rootPath . fromAb
...
+-- We want to treat IAT dependencies separately so that we can resolve them into user-defined custom dependencies.
+-- Remove any deps from the graph of type IATType.
+-- Evaluates to the list of dependencies so removed and the list of project results after removal.
+extractIATDeps :: [ProjectResult] -> ([Dependency], [ProjectResult])
+extractIATDeps results = do
+  -- hack: process the results twice to get the two pieces of infomation desired.
+  -- I'm happy to do this a better way but I'm not sure how to combine these into one action.
+  let iatDepGraphs = Graphing.filter isIATDep <$> (projectResultGraph <$> results)
+  let iatOnlyDeps = concat $ Graphing.toList <$> iatDepGraphs
+  (iatOnlyDeps, filterIATDeps <$> results)
```
kitified, line 393: I'd love anyone's assistance on this- I'm sure there's a cleaner way to do this but I wasn't sure how to do it non-imperatively.

src/App/Fossa/Analyze.hs
```diff
@@ -371,6 +381,27 @@ filterProjects rootDir = filter (isProductionPath . dropPrefix rootPath . fromAb
...
+    isIATDep :: Dependency -> Bool
+    isIATDep d = (dependencyType d) == IATType
+    filterIATDeps :: ProjectResult -> ProjectResult
+    filterIATDeps ProjectResult{..} =
+      ProjectResult
+        projectResultType
+        projectResultPath
+        (Graphing.filter (not . isIATDep) projectResultGraph)
+        projectResultGraphBreadth
+
```
scruffystuffs, line 404: There's quite a few things I'd like to do about this block.  I don't know if we can prevent processing twice with our filter, but we may be able to do a few other fixes.

* Primarily, we're `map`-ping, and therefore iterating, multiple times, we can fix that.
* We can use record update syntax instead of manually creating another `ProjectResult`.
* We can avoid using `let` and `where` together, which can be confusing.
* We don't need do notation here.

```suggestion
-- FIXME: we do Graphing.filter twice, but one is inverted.  Can we just split the Graphing?
extractIATDeps :: [ProjectResult] -> ([Dependency], [ProjectResult])
extractIATDeps results = (iatOnlyDeps, map filterIATDeps results)
  where
    iatOnlyDeps :: [Dependency]
    iatOnlyDeps = concatMap pullDeps results

    pullDeps :: ProjectResult -> [Dependency]
    pullDeps = Graphing.toList . Graphing.filter isIATDep . projectResultGraph

    isIATDep :: Dependency -> Bool
    isIATDep d = (dependencyType d) == IATType

    filterIATDeps :: ProjectResult -> ProjectResult
    filterIATDeps result = result {projectResultGraph = Graphing.filter (not . isIATDep) $ projectResultGraph result}

```

One piece of advice from the style guide (#325): use `map` instead of `fmap` or `<$>`, since lists are different in performance than a lot of other functors.  Prefer explicit operations on all collection-like functors (or functor-like types, e.g.: `Set`).  Let me know if anything here doesn't make sense, or if it's not clear how we got here from the original code.

scruffystuffs (changes requested):
Left one major change, as well as a comment about formatting.  Let's make sure we agree on the major before merging.

src/App/Fossa/Analyze.hs
cnr: I think this highlights a couple of technical design issues:

- strategies typically don't have access to `ScanDestination`, which is required for the API requests. This is already kind of solved for `VSI`, because we already pass in the `ScanDestination`. I think the `resolveAssertedDeps` could actually happen within the `VSI` strategy
- strategies don't have a way to emit `AdditionalDepData`.. which is a big contributor for why `analyzeFossaDepsFile` isn't a strategy anymore

There are two possible resolutions here:

1. Extract out the VSI strategy and run it separately as part of `analyze`, similar to `analyzeFossaDepsFile`
2. Add a field to `DependencyResults` that would allow us to emit `AdditionalDependencyData`

The first option is somewhat painful, and I'm not sure how to feel about the second option

cc @scruffystuffs

scruffystuffs: I really don't like the second option, but I think it's because our model of what we find to what source units we output is wrong.  The entire `fossa-deps` file is wrapped into a single source unit, which was impossible in the current strategy model.

I'm less interested in how we convert to source units (and therefore, how we emit "additional dep data"), and more interested in "How do we abstractly model *every single thing we found*, during all kinds of analyses?"  I don't know if we can feasibly change this for this PR, but I think it shows we're still a little bit off the mark.

kitified: > Extract out the VSI strategy and run it separately as part of `analyze`, similar to `analyzeFossaDepsFile`

I would prefer to not do this as part of this PR, but this might be the best path forward for future maintainability- I'm envisioning having the VSI strategy output its own `SourceUnit` directly instead of a `Graphing Dependency`. Is that what you were thinking as well?

In particular I have another somewhat substantial change coming down the pipe next, I could do this there.

cnr: > I'm envisioning having the VSI strategy output its own SourceUnit directly instead of a Graphing Dependency. Is that what you were thinking as well?

Yes, exactly

kitified: @cnr I ended up refactoring the VSI logic out as part of this PR after all. I didn't see a real reason to wait. Would appreciate if you want to take a look!

src/App/Fossa/IAT/Types.hs
```diff
@@ -21,3 +28,20 @@ data UserDefinedAssertionMeta = UserDefinedAssertionMeta
...
+    UserDefinedAssertionMeta
+      <$> obj .: "name"
+      <*> obj .: "version"
+      <*> obj .: "license"
+      <*> obj .: "description"
+      <*> obj .: "url"
+
+-- | Extract user defined binary dependencies from a list of dependencies.
+-- Evaluates to the list of user defined binaries and the remaining graph.
+extractUserDefinedBinaries :: [Dependency] -> ([Dependency], [Dependency])
```
cnr, line 43: That this exists is a code smell to me; we're still constructing some combined thing ("all of the dependencies") and doing post-processing to split out a subset we care about -- why not keep them separate as part of the generating function?

Further: `IATType` is only used as a marker for "filter me out", which betrays the purpose of the (admittedly-already-broken) `DependencyType`.

Proposal: the type of `pluginAnalyze` can change to emit `m ([ValidVSILocator], [VSIUserDep])`, where `VSIUserDep` is a new type along the lines of

```hs
data VSIUserDep
  { vsiUserDepName :: Text
  , vsiUserDepVersion :: Text
  } 
```

`ValidVSILocator`s can be later processed into `Dependency`s, and `VSIUserDep`s are marshalled into additionaldepdata

---

More generally: it's best to keep our data in well-behaved, correct-by-construction internal types for as long as we can, delaying the conversion to something much less well-behaved (`Dependency`, which has fewer constraints on its shape) until the very last moment

After something is made into a `Dependency`, we should never touch it again

src/App/Fossa/IAT/Resolve.hs
```diff
@@ -0,0 +1,61 @@
...
+    , srcUserDepLicense = assertedLicense
+    , srcUserDepDescription = assertedDescription
+    , srcUserDepHomepage = assertedUrl
+    }
+
+-- | toLocator is a special case of Dependency to Locator.
+-- All asserted dependencies are of type IATType, have a name, and have a non-empty version which is constrained to a specific version.
+-- Anything else shouldn't even get here, but if it does it should be ignored.
+toLocator :: Dependency -> Maybe Locator
+toLocator Dependency{dependencyType = IATType, dependencyName, dependencyVersion = Just (CEq v)} =
```
cnr, line 78: [related to main comment] This is an example of something that makes me uneasy.

> All asserted dependencies are of type IATType, have a name, and have a non-empty version

..is not actually specified in the types at all. This behavior we're relying on can change underneath our feet and we'd have no idea until we started seeing failures at runtime. By using stronger domain types, this is something that gets verified at compile time

src/App/Fossa/IAT/Resolve.hs
```diff
@@ -0,0 +1,61 @@
...
+
+toSourceUnit :: ProjectResult -> Maybe [SourceUserDefDep] -> SourceUnit
+toSourceUnit project deps = do
+  let unit = Srclib.toSourceUnit project
+  unit{additionalData = fmap toDepData deps}
+  where
+    toDepData d = AdditionalDepData (Just d) Nothing
+
+resolveUserDefined :: (Has (Lift IO) sig m, Has Diagnostics sig m) => ApiOpts -> [Dependency] -> m (Maybe [SourceUserDefDep])
+resolveUserDefined _ [] = pure Nothing
```
cnr, line 57: [nit] this case turns out to be redundant; `traverse` on the empty list will return an empty list, which will match the `null assertions` check below
```suggestion
```

cnr (commented):
Looks a lot better already after that refactor

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -501,10 +502,10 @@ assertUserDefinedBinaries apiOpts UserDefinedAssertionMeta{..} fingerprints = fo
   _ <- req POST (assertUserDefinedBinariesEndpoint baseUrl) (ReqBodyJson body) ignoreResponse baseOpts
   pure ()
 
-resolveUserDefinedBinaryEndpoint :: Url scheme -> Locator -> Url scheme
-resolveUserDefinedBinaryEndpoint baseurl locator = baseurl /: "api" /: "iat" /: "resolve" /: "user-defined" /: renderLocator locator
+resolveUserDefinedBinaryEndpoint :: Url scheme -> IAT.UserDep -> Url scheme
```
cnr, line 505: [no action] ooh this is even better. I like that we skip `Locator` entirely

src/App/Fossa/VSI/Types.hs
```diff
@@ -0,0 +1,79 @@
...
+      <> pretty (locatorFetcher locator)
+      <> " to known dependency type. Locator: "
+      <> viaShow locator
+
+validateLocator :: Srclib.Locator -> Either LocatorParseError Locator
+validateLocator loc =
+  Locator
+    <$> Right (Srclib.locatorFetcher loc)
+    <*> Right (Srclib.locatorProject loc)
+    <*> validateRevision loc
```
cnr, line 68: [nit] love the applicative syntax here -- makes it immediately clear to me that we only care about whether `validateRevision` fails

there's another way we can do this -- because it turns out that `f <$> pure a <*> pure b === pure (f a b)`. In the case of `Either e`, `pure = Right`, so:

```hs
validateLocator :: Srclib.Locator -> Either LocatorParseError Locator
validateLocator loc =
  Locator
    <$> Right (Srclib.locatorFetcher loc)
    <*> Right (Srclib.locatorProject loc)
    <*> validateRevision loc
```

is the same as 
```hs
validateLocator :: Srclib.Locator -> Either LocatorParseError Locator
validateLocator loc =
  Right (Locator (Srclib.locatorFetcher loc) (Srclib.locatorProject loc))
    <*> validateRevision loc
```

---

further, `pure f <*> x` is the same as `f <$> x`:

```hs
-- same type as fmap, but using only applicative methods
fmapDefault :: Applicative f => (a -> b) -> f a -> f b
fmapDefault f x = pure f <*> x
```

so we can make one final change:
```hs
validateLocator :: Srclib.Locator -> Either LocatorParseError Locator
validateLocator loc =
  Locator (Srclib.locatorFetcher loc) (Srclib.locatorProject loc)
    <$> validateRevision loc
```

src/App/Fossa/VSI/Types.hs
```diff
@@ -0,0 +1,79 @@
...
+import Srclib.Types qualified as Srclib
+
+-- | VSI supports a subset of possible Locators.
+-- | Specifically, all VSI locators must have a valid revision.
+data Locator = Locator
+  { locatorFetcher :: Text
+  , locatorProject :: Text
+  , locatorRevision :: Text
+  }
+  deriving (Eq, Ord, Show)
```
cnr: [no action] 🚀

src/App/Fossa/VSI/Types.hs
```diff
@@ -0,0 +1,79 @@
...
+
+toDependency :: Locator -> Either ToDependencyError Dependency
+toDependency locator =
+  Dependency
+    <$> validateDepType locator
+    <*> Right (locatorProject locator)
+    <*> (Right . Just . CEq $ locatorRevision locator)
+    <*> Right []
+    <*> Right []
+    <*> Right mempty
```
cnr, line 46: [no action] I like the use of the explicit `Right` constructor here over `pure` -- I wonder if we should add that to the style guide for simple applicatives like `Either e` and `Maybe`. Not sure how often this comes up, though

cc: @scruffystuffs

src/App/Fossa/VSI/Types.hs
```diff
@@ -0,0 +1,79 @@
...
+  { locatorFetcher :: Text
+  , locatorProject :: Text
+  , locatorRevision :: Text
+  }
+  deriving (Eq, Ord, Show)
+
+parseLocator :: Text -> Either LocatorParseError Locator
+parseLocator a = do
+  let plain = Srclib.parseLocator a
+  validateLocator plain
```
cnr, line 29: [nit] use function composition

```hs
parseLocator = validateLocator . Srclib.parseLocator
```

cnr (approved):
This is great. A couple more nits then LGTM

src/App/Fossa/VSI/Types.hs
scruffystuffs: It's worth noting, though I wouldn't make it strict.  I'll add this.

PR #329, opened 2021-08-13
fix(maven): Parse platform specifiers in Maven package IDs

liftM:
# Overview

This PR fixes a parsing issue where package IDs with platform specifiers would not parse correctly.

It also adds a unit test as well as an output fixture from the customer who reported this bug - I'm working with them to figure out whether we can actually commit the anonymized customer fixture.

## Acceptance criteria

The Maven `dependency:tree` tactic should work on Maven package IDs that include platforms.

## Testing plan

Run `cabal test` on the unit test, and with the fixture (DM @liftM for the fixture).

## References

- [ZD-2896](https://fossa.zendesk.com/agent/tickets/2896)

## Checklist


scruffystuffs (approved):
LGTM, but that package format is so stupid.  Optional middle components with identical separators?

PR #333, opened 2021-08-16
IAT: Register and resolve revision assertions

kitified:
# Overview

Support registering and resolving revision assertions.

Revision assertions are binaries that are asserted to a specific revision in FOSSA during another analysis run.
* Registering them entails running project analysis, then uploading the fingerprints of files in the specified directory to FOSSA pointing at the revision that was just analyzed.
* Resolving them entails using the FOSSA dependencies API to get a list of dependencies belonging to the project, then treating those as deep dependencies in our graph.

## Acceptance criteria

* Support asserting binaries when analyzing a project.
* Support later discovering that project as a dependency of another project.

## Testing plan

1. I selected a simple project to analyze, [utfutil](https://github.com/TomOnTime/utfutil). This project was selected because it has a small number of dependencies and those dependencies don't overlap with those from my base project.
2. I [ran analysis over `utfutil`](https://staging.int.fossa.io/projects/custom%2B1%2Futfutil-iat/refs/branch/master/1629050393/browse/dependencies), registering a binary on my system as the binary for `utfutil` during analysis.
3. I then put that registered binary in the `quick` VPS demo project directory.
4. I then ran analysis on the `quick` demo project.
5. Finally, [I observed](https://staging.int.fossa.io/projects/custom%2B1%2Fquick-vsi-test-assertion/refs/branch/master/1629146250/browse/dependencies) that the resulting analysis run lists `utfutil` as a dependency, and its dependencies as deep dependencies.

## Risks

Again this touches core project analysis code, so would appreciate a review of that area especially.

## Checklist


src/App/Fossa/Analyze.hs
```diff
@@ -257,12 +267,18 @@ analyze (BaseDir basedir) destination override unpackArchives jsonOutput enableV
...
-      OutputStdout -> logStdout . decodeUtf8 . Aeson.encode $ buildResult manualSrcUnits filteredProjects
-      UploadScan opts metadata -> uploadSuccessfulAnalysis (BaseDir basedir) opts metadata jsonOutput override sourceUnits
+      OutputStdout -> do
+        _ <- logStdout . decodeUtf8 . Aeson.encode $ buildResult manualSrcUnits filteredProjects
+        pure Nothing
+      UploadScan opts metadata -> do
+        locator <- uploadSuccessfulAnalysis (BaseDir basedir) opts metadata jsonOutput override sourceUnits
+        pure $ Just locator
+
+  doAssertRevisionBinaries iatAssertion apiOpts uploadLocator
```
cnr, line 281: [nit] let's move this up into the `UploadScan` case branch above -- prevents us from having to pass out an `uploadLocator`, which only happens in that branch anyway

in that branch we have the concrete `opts` and `locator` too, so we can remove the `Maybe` wrappers from the args to `doAssertRevisionBinaries`

src/App/Fossa/Analyze.hs
```diff
@@ -327,9 +348,12 @@ uploadSuccessfulAnalysis (BaseDir basedir) apiOpts metadata jsonOutput override
   -- Warn on contributor errors, never fail
   void . Diag.recover . runExecIO $ tryUploadContributors basedir apiOpts (uploadLocator uploadResult)
 
-  if fromFlag JsonOutput jsonOutput
-    then logStdout . decodeUtf8 . Aeson.encode $ buildProjectSummary revision (uploadLocator uploadResult) buildUrl
-    else pure ()
+  _ <-
```
cnr, line 351: does this compile without this line?

cnr (approved):
LGTM - couple of nits

src/App/Fossa/Analyze.hs
kitified: So it does! Assumptions haunt me yet again.

PR #339, opened 2021-08-19
Output the real project ID and revision with `--json`

scruffystuffs:
# Overview

FOSSA strips protocols (and .git suffixes) from project ID's in locators.  When this happens, the project ID we return under the `--json` flag does not match the project component of the actual locator.  This PR uses the return value of the build upload call to determine the correct locator.  This way, even though FOSSA may not be consistent with the _requested_ project ID, the CLI will be correctly determine the _true_ project ID.

## Acceptance criteria

* Without `--json`, no change.
* With `--json`:
  * fail if no revision was sent (with good error message).
  * use the API return value instead of the supplied project name

## Testing plan

Run with ` --json --project https://foo.bar/baz.git`, check the following (using psuedo-`jq` syntax):
- `$.project` is an exact infix of `$.id`
- `$.project` is NOT equal to `https://foo.bar/baz.git` (only true when original starts with `https?:`)

## Risks

No known risks.

## References

https://teamfossa.slack.com/archives/CJBPV3Z0E/p1629392377095100
https://fossa.zendesk.com/agent/tickets/3114

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,8 +1,12 @@
 # Spectrometer Changelog
 
+## v2.15.5
+
+- Fixes an issue where `--json` would output the raw project ID, instead of a normalized ID ([#339](https://github.com/fossas/spectrometer/pull/339))
+
 ## v2.15.4
 
-- Gradle: Search parent directories for gradlew and gradlew.bat ([#336](https://github.com/fossas/spectrometer/pull/336)) 
+- Gradle: Search parent directories for gradlew and gradlew.bat ([#336](https://github.com/fossas/spectrometer/pull/336))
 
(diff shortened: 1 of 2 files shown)


xinding33:
Does this fix the opposite case where a `.git` suffix is appended to the core project ID when one is not specified by the cli `--projects` flag?

scruffystuffs:
FOSSA normalizes anything we send through `--project`.  This does nothing to change that (and we can't change that).  But if FOSSA adds a `.git` suffix for whatever reason, as long as it reports back correctly, we'll pick it up and output it correctly here.

cnr (approved):

PR #340, opened 2021-08-20
Refractors Cocoapods lock tactic to use Data.Yaml parser

meghfossa:
# Overview

- It correctly parses when Podlock file has quotations in entry for dependency and pod section.  
- It uses Data.Yaml to do lockfile parsing

## Acceptance criteria

- Refractors Parser to use Yaml Parsing
- Addressing @cnr's PR comments
- Parse Entry in Pod correctly when enclosed with quotation
- Parse Entry in Dependency correctly when enclosed with quotation

## Testing plan

- I've included unit tests to of those effect

For reproducible example, 

- Setup simple pod project
- Add `pod 'GoogleUtilities' ` in `Podfile`
- Execute `pod install`
- Analyze Pod project with applied fixes
- Analyze Pod project without applied fixes

## Risks

N/A

## References

Works on https://github.com/fossas/team-analysis/issues/716

## Checklist


src/Strategy/Cocoapods/PodfileLock.hs
```diff
@@ -1,34 +1,42 @@
...
+  Diagnostics,
+  Has,
+  context,
+  run,
+ )
+import Data.Aeson (FromJSON (parseJSON), (.:))
 import Data.Char qualified as C
 import Data.Foldable (traverse_)
 import Data.Functor (void)
+import Data.HashMap.Lazy qualified as HM
```
meghfossa, line 21: - [ ] Fix to `HashMap`

src/Strategy/Cocoapods/PodfileLock.hs
```diff
@@ -114,99 +125,50 @@ data Remote = Remote
...
+parseNameAndVersion = do
+  name <- parseName
+  version <- parseVersion
+  pure (name, version)
+
+instance FromJSON PodLock where
+  parseJSON = Yaml.withObject "Podfile.lock content" $ \o -> do
+    pods <- o .: "PODS"
+    deps <- o .: "DEPENDENCIES"
+    pure $ PodLock pods deps
```
scruffystuffs, line 144: Since this is a parser, we should prefer applicative parsing to monadic parsing:
```suggestion
instance FromJSON PodLock where
  parseJSON = Yaml.withObject "PodLock" $ \obj ->
    PodLock <$> obj .: "PODS"
      <*> obj .: "DEPENDENCIES"
```

Two nits as well:

- Prefer `obj` name to `o`.  It's codebase convention for almost every json and yaml parser we have.
- For the error string, it's sufficient and conventional to just use the name of the data structure.

src/Strategy/Cocoapods/PodfileLock.hs
```diff
@@ -114,99 +125,50 @@ data Remote = Remote
...
+
+instance FromJSON Pod where
+  parseJSON (Yaml.String p) = parserPod p Nothing
+  parseJSON (Yaml.Object o) = case exactlyOne $ HashMap.toList o of
+    Just (podEntry, podDepsListing) -> parserPod podEntry $ Just podDepsListing
+    Nothing -> fail $ "Expected list of dependencies, but received: " <> show o
```
scruffystuffs, line 150: [nit]: We can probably inline the pattern-match here:
```suggestion
  parseJSON (Yaml.Object o) = case HashMap.toList o of
    [(podEntry, podDepsListing)] -> parserPod podEntry $ Just podDepsListing
    _ -> fail $ "Expected list of dependencies, but received: " <> show o
```

src/Strategy/Cocoapods/PodfileLock.hs
```diff
@@ -114,99 +125,50 @@ data Remote = Remote
...
+parseName = lexeme $ takeWhile1P (Just "dep") (not . C.isSpace)
+
+parseVersion :: Parser Text
+parseVersion = between (char '(') (char ')') $ lexeme (takeWhileP (Just "version") (/= ')'))
+
+parseNameAndVersion :: Parser (Text, Text)
+parseNameAndVersion = do
+  name <- parseName
+  version <- parseVersion
+  pure (name, version)
```
scruffystuffs, line 138: [optional]: I'm on the fence about this one, because I hate using tuple constructors as functions (I prefer tuple literals, or not using tuples at all), but ideally this should be applicative instead of monadic:
```suggestion
parseNameAndVersion :: Parser (Text, Text)
parseNameAndVersion = (,) <$> parseName <*> parseVersion
```

Another option is to do something like this:
```haskell
data NameAndVersion = NameAndVersion Text Text

parseNameAndVersion :: Parser NameAndVersion
parseNameAndVersion = NameAndVersion <$> parseName <*> parseVersion

doThing :: Text -> Foo
doThing txt = case parse parseNameAndVersion "" txt of
  Left err -> handleError err
  Right (NameAndVersion name version) -> do
    -- do stuff with `name` and `version`
```

scruffystuffs (approved):
Left a few comments, looks good otherwise.

PR #341, opened 2021-08-24
Don't attempt to upload analysis results to monorepo projects

kitified:
# Overview

Monorepo scans are not valid scans for `fossa analyze` (without `--experimental-enable-monorepo`), but currently there is no way for users to know this- their scans run and claim to be successful, but when they open the project in the UI nothing looks different and the new revision is nowhere to be found.

## Acceptance criteria

* Fail the standard analysis flow for projects that are marked monorepo, providing a helpful error message.
* Do not fail standard analysis for new projects.
* Do not fail standard analysis for existing non-monorepo projects.

## Testing plan

1. Scan a monorepo project using `fossa vps analyze` or `fossa analyze --enable-experimental-monorepo aosp` (both of these create a monorepo project, so they are equivalent for the purpose of this test).
2. Immediately try to use `fossa analyze` on the same project. Observe that the error message appears and nothing changes in the UI.
3. Scan a new project that doesn't exist. Observe that it works and the new project appears in the UI.
4. Scan a new revision of the project in step 3. Observe that it works and the new revision appears in the UI.

_Aside:_ Also tested the changes in #321 at the same time against these projects, which works great!

## Risks

No major risk here. Worst case scenario if the `getProject` API hasn't been updated on the server to return monorepo status we assume all projects are _not_ monorepo, so we get the same functionality as before (ignored analysis runs).

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,8 +2,8 @@
 
 ## v2.15.7
 
-- Resolve an issue where errors running `fossa report` and `fossa test` would be made more confusing when the project isn't a monorepo project ([#321](https://github.com/fossas/spectrometer/pull/321))
-- Prevent uploading standard analysis results to monorepo projects, where they'd be silently ignored ([#341](https://github.com/fossas/spectrometer/pull/341))
+- Resolves an issue where errors running `fossa report` and `fossa test` would be made more confusing when the project isn't a monorepo project ([#321](https://github.com/fossas/spectrometer/pull/321))
+- Prevents uploading standard analysis results to monorepo projects, where they'd be silently ignored ([#341](https://github.com/fossas/spectrometer/pull/341))
 
 ## v2.15.6 
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,10 @@
 # Spectrometer Changelog
 
+## v2.15.7
+
+- Resolve an issue where errors running `fossa report` and `fossa test` would be made more confusing when the project isn't a monorepo project ([#321](https://github.com/fossas/spectrometer/pull/321))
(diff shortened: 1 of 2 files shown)


src/App/Fossa/Analyze.hs
```diff
@@ -330,6 +337,7 @@ uploadSuccessfulAnalysis ::
 uploadSuccessfulAnalysis (BaseDir basedir) apiOpts metadata jsonOutput override units = do
   revision <- mergeOverride override <$> (inferProjectFromVCS basedir <||> inferProjectDefault basedir)
   saveRevision revision
+  dieOnMonorepoUpload apiOpts revision
```
scruffystuffs, line 340: [nit]: Can we move this before the `saveRevision` call?  I think it's OK either way, but seems ideal to only save if we're planning to use it.

Changelog.md
```diff
@@ -1,5 +1,10 @@
 # Spectrometer Changelog
 
+## v2.15.7
+
+- Resolve an issue where errors running `fossa report` and `fossa test` would be made more confusing when the project isn't a monorepo project ([#321](https://github.com/fossas/spectrometer/pull/321))
+- Prevent uploading standard analysis results to monorepo projects, where they'd be silently ignored ([#341](https://github.com/fossas/spectrometer/pull/341))
```
cnr, line 6: [nit] we've been wording changelog entries as, e.g., "_Resolves_ an issue where ..." or "_Prevents_ ..."

It's as if the sentence implicitly starts like "Version v2.15.7 resolves an issue where ..." or "Version v2.15.7 prevents ..."

scruffystuffs (approved):
One question, but your choice on what to do with it, LGTM

cnr (approved):
LGTM; nit on changelog entries

PR #349, opened 2021-08-30
Add better logging and diffing

scruffystuffs:
This should be a transparent change, only adding logs to the CI lint and fmt jobs.  Also, when the formatter fails, a git diff is shown between the expected and current states.


meghfossa (approved):
LGTM

PR #350, opened 2021-08-30
Also apply to `fmt` target

scruffystuffs:
Follow-up to #349, also apply the `fourmolu` logging to the `fmt` target in the makefile.


meghfossa (approved):
LGTM

PR #353, opened 2021-09-01
Detect binaries as dependencies

kitified:
# Overview

Add feature flagged functionality to detect binaries as dependencies, using the [new `discovered-binaries` feature](https://github.com/fossas/FOSSA/pull/6443).

## Acceptance criteria

This was initially a feature request from one customer, but we think this is generally useful as an expansion to C/C++ support at FOSSA. As such we've gone ahead and enabled their specific request behind a feature flag, with the plan to roll this in to broader C/C++ support as that functionality matures.

## Testing plan

I ran a FOSSA instance on the `discovered-binary-fetcher` branch and ran a scan against the CPP demo project:
```shell
# Run a VSI command
# Assuming https://github.com/fossas/cpp-vsi-demo is at ~/projects/cpp-demo
cabal run fossa -- analyze ~/cpp-demo/example-internal-project --endpoint '<endpoint>' --fossa-api-key <key> --project discover-binaries --revision $(date +%s) --exclude-path vendor
```

The binaries located in the project under the `include/` directory were detected and added to the project as unlicensed dependencies. 

## Risks

This is feature flagged, so not super risky. 

This feature is only enforced client side; if someone recompiles Spectrometer with this feature flag force enabled they will use this feature. We don't consider that an issue because we plan to make it generally available as part of our C/C++ analysis once it is out of beta.

## Checklist


cnr:
Some high-level concerns before I dive into code:

---

I'm concerned about the noise of false-positive deps

e.g. running this on the spectrometer repo, we get an additional 3900 dependencies:

- build products, like `**/.hi`, `**/.o`, ...
- `.git/objects/**`
- `**/.DS_Store`

Looking at `.gitignore` (and whatever the similar thing is for SVN) could solve part of this issue -- keeping in mind intracacies around submodule gitignores, etc -- but what about scanning directories without VCS?

On the `gradle/gradle` repo, we get another 274 binary dependencies, even ignoring the files ignored by `.gitignore` -- some `.gpg` files, images, example jars, asset zip files, etc

On the fossa-core repo, we get another 100 binary dependencies -- some testing jars, some legitimate binaries (go-multirunner, nomos), nomos build products (e.g., `.o`), images, fonts

There are a bunch of types of files I wouldn't want to see as dependencies but are classified as "binary" by git

---

Reading every file in the filetree is very very slow, especially on large repositories

---

We'd need to be resilient to file-read-errors: for example, if there are any orphaned symlinks, we abort analysis entirely:

```
[ERROR] ----------
  An error occurred:

      Error reading file /Users/connor/Desktop/tmp5/home-assistant/.Python:
          /Users/connor/Desktop/tmp5/home-assistant/.Python: openFile: does not exist (No such file or directory)

      Traceback:
        - Walking the filetree
```

cnr:
I can't see the related pivotal tracker ticket, but I wonder if there's a way to support the customers' use-cases with a bit of scripting and the use of [`custom-dependencies` in a `fossa-deps` yaml/json file](https://github.com/fossas/spectrometer/blob/4770c05ed58f9d93ba6878d955ea2e55dd65c877/docs/userguide.md#custom-dependencies)

kitified:
> I'm concerned about the noise of false-positive deps
> There are a bunch of types of files I wouldn't want to see as dependencies but are classified as "binary" by git

This is a fair concern. To explain, our beta partner for these features cares more about finding all the binaries than they do about false positives. Since it is technically possible for a user to build and use a binary with an uncommon extension (eg `libssl.png`), I didn't put any initial filtering in place, instead thinking that we'd wait for user feedback.

Since this feature works with include and exclude filters, users may use those to customize which directories are scanned- for example, they can run `--exclude-path .git` to prevent git objects from appearing in the result. 
I would be open to configuring a built-in allowlist of file extensions and built-in prebuilt filter set (eg, "allow .o, .so, ...; deny .git") in an attempt to give the user a more narrowly scoped set of binaries to start. I'm open to the idea that the likelihood of uncommon file extensions is low enough that we should provide some reasonable defaults.

Long term I think our goal would be to develop a heuristic with which we can inspect binary files to determine if they're likely to be C/C++ binaries rather than relying on file extensions, but it may be sensible to start with file extensions.

Note however, some of the things you've flagged as false positives aren't false positives from a C/C++ perspective:
* `build products, like **/.hi, **/.o, ...`

These are examples of what C/C++ users are after. Specifically, build products are potentially libraries in C/C++, so we want to highlight these for the user to explicitly ignore or license.

> Reading every file in the filetree is very very slow, especially on large repositories

I'd expect similar performance from Haskell as we get from Go; in our experience from fingerprinting large file trees in Go this is a relatively quick process, although admittedly it's slower than simply walking the file tree for sure. Since this is feature flagged this wouldn't cause issues with other users today, only for our beta partner for this feature.

Our plan at this time is that once this feature is mature enough to be out of beta, we'd still only enable it by default in C/C++ contexts so this will be opt-in until C/C++ scanning is on by default, which has a lot larger performance hurdles to overcome so won't be any time soon. The only reason why the flag doesn't require C/C++ to be enabled is because today many users are using archive upload for C/C++, and we want to have the ability to test this feature with them without making them swap to the new C/C++ flow entirely.

However if you have cause to believe that this would be a very slow process in Haskell specifically I'd love to hear it- this probably wouldn't stop the feature from being made at all, but would mean I'd pull it out into a plugin.

> We'd need to be resilient to file-read-errors: for example, if there are any orphaned symlinks, we abort analysis entirely

Unless I'm misunderstanding something, this is already a concern today, right? This PR reuses the `walk'` function used elsewhere so I'd expect this issue to appear during scans of such environments before this PR.

> I can't see the related pivotal tracker ticket, but I wonder if there's a way to support the customers' use-cases with a bit of scripting and the use of custom-dependencies in a fossa-deps yaml/json file

I suppose we could theoretically create a script that finds all binaries in the tree and creates a fossa-deps file for them, but we think that after refinement this feature will be generally useful for C/C++ users. We will probably want to implement for example additional checks on the binaries, or prebuilt filters, or something similar before making this generally used in C/C++, but we're hoping to get that feedback from this customer (and have this feature available to test with other FOSSA customers). For this reason we'd rather have this functionality in the CLI rather than available via scripting, as we can more easily support and control this feature when we haven't essentially handed it off to the user.

kitified:
@cnr sorry for the messy git history, but this is now ready to review with the changes we talked about offline!

src/App/Fossa/VSI/IAT/Types.hs
```diff
@@ -15,12 +15,16 @@ import Data.Aeson (
...
+import Data.String.Conversion (toString)
 import Data.Text (Text)
 
 -- | Fingerprint uniquely idenfies a file, derived from its content.
 -- Fingerprints are backed by base16 representations of underlying data.
 newtype Fingerprint = Fingerprint {unFingerprint :: Text}
 
+instance Show Fingerprint where
+  show f = toString $ unFingerprint f
+
```
cnr, line 27: [nit] let's `deriving Show` on `Fingerprint` instead

(hopefully we're not using this `Show` instance anywhere directly -- `Show` by convention is primarily for debug)

src/Effect/ReadFS.hs
```diff
@@ -114,6 +119,13 @@ readContentsBS' path = send (ReadContentsBS' path)
 readContentsBS :: (Has ReadFS sig m, Has Diagnostics sig m) => Path b File -> m ByteString
 readContentsBS = fromEither <=< readContentsBS'
 
+fileIsBinary :: (Has ReadFS sig m, Has Diagnostics sig m) => Path b File -> m Bool
+fileIsBinary file = do
+  -- git defines binaries as "a file that contains a 0 byte in the first 8000 bytes of the file".
+  attemptedContent <- send (ReadContentsBSLimit' file 8000)
+  content <- fromEither attemptedContent
+  pure $ BS.elem 0 content
```
cnr, line 127: A few suggested changes:

- move `fileIsBinary` into `App.Fossa.BinaryDeps`
- move the git comment to a function-level `-- | Doc string` on `fileIsBinary`
- export the `readContentsBSLimit` operation from `Effect.ReadFS`, similar to the existing `readContentsBS` operation:

```hs
-- name clashes with an existing function you defined; might have to rename it
readContentsBSLimit' :: Has ReadFS sig m => Path b File -> Int -> m (Either ReadFSErr ByteString)
readContentsBSLimit' file limit = send (ReadContentsBSLimit' file limit)

readContentsBSLimit :: (Has ReadFS sig m, Has Diagnostics sig m) => Path b File -> Int -> m ByteString
readContentsBSLimit file limit = readContentsBSLimit' file limit >>= fromEither
```

src/App/Fossa/VSI/IAT/Fingerprint.hs
```diff
@@ -40,7 +41,7 @@ hashFile fp =
   sendIO (runConduitRes (sourceFile fp .| sinkHash))
     `catch` (\(e :: IOException) -> fatalText ("unable to hash file: " <> toText (show e)))
 
-fingerprintRaw :: (Has (Lift IO) sig m, Has Diagnostics sig m) => Path Abs File -> m Fingerprint
+fingerprintRaw :: (Has (Lift IO) sig m, Has Diagnostics sig m) => Path x File -> m Fingerprint
```
cnr, line 44: [nit] I'd like to keep this as `Path Abs File` if possible -- ensures that we're not accidentally using relative filepaths if, e.g., we run with a different `BaseDir` like `fossa analyze path/to/dir`

src/App/Fossa/Analyze.hs
```diff
@@ -245,19 +263,19 @@ analyze ::
...
   capabilities <- sendIO getNumCapabilities
 
   let apiOpts = case destination of
         OutputStdout -> Nothing
         UploadScan opts _ -> Just opts
 
   manualSrcUnits <- analyzeFossaDepsFile basedir apiOpts
-  vsiResults <- analyzeVSI enableVSI apiOpts basedir filters
+  vsiResults <- analyzeVSI modeVSIAnalysis apiOpts basedir filters
+  binarySearchResults <- analyzeDiscoverBinaries modeBinaryDiscovery basedir filters
```
cnr, line 278: [no action] Might be useful to pair at some point on how to make these tasks run in parallel / show status updates. It's not as easy as it should be

both of these are long-running and might benefit from the parallelism / `context` status updates

src/App/Fossa/Analyze.hs
```diff
@@ -119,6 +122,14 @@ data UnpackArchives = UnpackArchives
 
 data JsonOutput = JsonOutput
 
+-- | Collect analysis modes into a single type for ease of use.
+-- These modes are intended to be different options that alter how analysis is performed or what analysis steps are followed.
+data ModeOptions = ModeOptions
+  { modeVSIAnalysis :: VSIAnalysisMode
+  , modeIATAssertion :: IATAssertionMode
+  , modeBinaryDiscovery :: BinaryDiscoveryMode
+  }
```
cnr, line 131: [no action] 💡 

the `analyzeMain` args sure are getting unwieldy.. I wonder if we should pack all of the args into a record

cnr (approved):
LGTM 🖨️

src/App/Fossa/VSI/IAT/Types.hs
kitified: Oops, I added this for debugging and forgot to remove it. Will do.

src/Effect/ReadFS.hs
kitified: I'm OK to move `fileIsBinary` to `BinaryDeps`. I put it here because at least in monorepo land we've found it super useful to be able to avoid inspecting files for eg IPR data when they're binary, so I thought it may be similarly generally useful here, but definitely happy to only make it more general when/if that need actually shows up.

kitified:
FOSSA `3.37.2` supports this functionality and is now in production, merging for release.

PR #355, opened 2021-09-01
Adds apple m1 installation workaround in docs

meghfossa:
# Overview

Adds workaround for apple m1 silicon in the installation docs. 

## Acceptance criteria

- Provided install instruction works for m1 macbooks.

## Testing plan

- I tested on my personal device. This is the same workaround described and identified by kit and connor.

## Risks

N/A

## References

Works on https://github.com/fossas/team-analysis/issues/713

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,12 @@ more.
 curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/spectrometer/master/install.sh | bash
 ```
 
+### macOS with Apple M1 Silicon
+
+```bash
+arch -x86_64 /bin/bash -c "$(curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/spectrometer/master/install.sh)"
+```
+
 ### Windows with Powershell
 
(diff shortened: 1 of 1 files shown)


cnr:
I think we can avoid a separate command for M1 users; my bet is that `arch -x86_64` is effectively just changing the return value when we get the architecture with `uname -m`, which causes it to download the darwin+x86_64 zip file.

We should be able to hardcode a case for darwin+aarch64 where it downloads the darwin+x86_64 zip file instead

meghfossa:
> I think we can avoid a separate command for M1 users; my bet is that `arch -x86_64` is effectively just changing the return value when we get the architecture with `uname -m`, which causes it to download the darwin+x86_64 zip file.
> 
> We should be able to hardcode a case for darwin+aarch64 where it downloads the darwin+x86_64 zip file instead

I wanted to get quick and dirty workaround in. This ticket is very low in backlog order. But, when this ticket gets picked on, I agree on approach you mentioned.

liftM (approved):

PR #357, opened 2021-09-02
Debug bundles (part 0): improvements for `Control.Carrier.Simple`/`Replay`/`Record`

cnr:
# Overview

This PR is a prerequisite to upcoming debug-bundle PRs, and addresses a few issues:

1. replay-logging is slow. Replay mode in particular is very very slow when replaying `fossa.debug.json` files generated in large directory structures
2. The "simple effect" machinery isn't "safe"; `SimpleC` uses `unsafeCoerce` in its `Algebra` instance
3. "Simple effect" datatypes are unmanageable: because of the phantom `m` type in, e.g., `data Exec m a`, we can't derive `Eq`/`Ord`/`Show` for those datatypes

To address these issues, we:

- 1a: Directly record effect invocations into a `Map` that uses effect constructor values as keys, rather than using JSON `Value` and serializing effect constructors on each effect invocation 
- 1b: Directly decode effect Journals into a `Map` that uses effect constructor values as keys, rather than serializing effect constructors and return values into JSON `Value` on each effect invocation during replay mode
- 2/3: Change the simple effect machinery to use types `data SomeEffect a` rather than `data SomeEffect m a`

We also incidentally:

- Enforce at the type level that absolute paths are used for `ReadFS` and `Exec` operations

Some preliminary observations about performance improvements under `--record`/`--replay` (more profiling / benchmarking needs to be done):

- `--record` mode imposes a ~10-20% overhead vs analysis without `--record` (for the small-medium projects I've been testing with)
- `--replay` mode is substantially faster now than running without `--replay`, whereas before it was orders of magnitude slower than running without `--replay` in the worst case

## Risks

- This touches our core effect datatypes (`ReadFS`/`Exec`), which is inherently risky -- but these refactors should be type-safe (and notably for these core effects, we no longer use `unsafeCoerce` in their `SimpleC` carrier, so this is a strict improvement for those)


scruffystuffs:
Why the `F` suffix?  What does `F` stand for?

cnr:
at some point all of the effect datatypes instances of `Functor`, so the `F` was for that. I guess it stuck even when the functor instances went away. I can change that to something better

src/Control/Carrier/Simple.hs
```diff
@@ -120,46 +124,52 @@ import Control.Applicative
 import Control.Carrier.Reader
 import Control.Carrier.State.Strict
 import Control.Monad.Trans
-import Data.Kind (Type)
-import Unsafe.Coerce (unsafeCoerce)
+
+---------- The Simple effect
+
+-- | Simple effects.. as an effect
+data Simple e m a where
```
meghfossa, line 131: [no action] Why simple as name (can't imagine what would be opposite of it 😅)- Is this conventional?

cnr: the real term is "first-order" effects (as opposed to "higher-order").. could definitely rename and add docs for what it means. would probably be useful

cnr: the tl;dr is first-order effects don't contain effectful sub-expressions:

example first-order:
```hs
data Filesystem m a where
  ReadFile :: FilePath -> Filesystem m String
```

example higher-order:
```hs
data Error e m a where
  Catch :: m a -> (e -> m a) -> Error e m a

catch :: Has (Error e) sig m => m a -> (e -> m a) -> m a
catch act hdl = send (Catch act hdl)

-- or from our codebase
data Diagnostics m a where
  Context :: String -> m a -> Diagnostics m a

context :: Has Diagnostics sig m => String -> m a -> m a
context ctx act = send (Context ctx act)
```

"higher-order" effects are ones where `m` appears as an argument to an effect constructor.

`Simple` is a wrapper effect around first-order effects that doesn't even allow the effect type to mention `m`. e.g., in this PR:

```hs
data ReadFSF a where
  DoesDirExist :: ... -> ReadFSF Bool

type ReadFS = Simple ReadFSF

doesDirExist :: Has ReadFS sig m => ... -> m Bool
doesDirExist dir = send (Simple (DoesDirExist dir))
```

src/Control/Effect/Replay.hs
```diff
@@ -34,34 +34,43 @@ import Unsafe.Coerce
...
+keyBy f = Map.fromList . map (\v -> (f v, v))
+
+convertFromJournal :: Replayable e => Journal e -> Map (e Void) (EffectResult e)
+convertFromJournal (Journal mapping) =
+  case parse id (fmap (keyBy unsafeEffectResultToKey) . traverse (uncurry replayDecode) $ Map.toList mapping) of
+    -- FIXME: this error happens when we're not able to deserialize effect
+    -- constructors/results from a journal. Because replay mode is only used
+    -- by spectrometer contributors locally, plumbing the error through
+    -- our normal error machinery is not worth the effort
+    Error str -> error $ "Unable to deserialize effect Journal: the Journal was likely created on a different spectrometer version. Error: " <> str
```
scruffystuffs, line 72: I don't think this comment is a FIXME, it's just a valid usage of `error`.

src/App/Fossa/Configuration.hs
```diff
@@ -115,7 +116,10 @@ readConfigFile file = do
 
 readConfigFileIO :: IO (Maybe ConfigFile)
 readConfigFileIO = do
-  config <- Diag.runDiagnostics $ runReadFSIO $ readConfigFile defaultFile
+  -- FIXME: we probably want to read from the target directory of analysis, not
```
meghfossa, line 119: - [ ] TODO: Use target directory of analysis

src/App/Fossa/Compatibility.hs
```diff
@@ -27,9 +26,10 @@ compatibilityMain ::
   [Argument] ->
   IO ()
 compatibilityMain args = withDefaultLogger SevInfo . runExecIO . withCLIv1Binary $ \v1Bin -> do
+  dir <- sendIO getCurrentDir
   cmd <- runStickyLogger SevInfo $ do
     logSticky "[ Waiting for fossa analyze completion ]"
-    exec [reldir|.|] $ v1Command v1Bin args
+    exec dir $ v1Command v1Bin args
```
meghfossa, line 32: [nit] Should this be target directory?

meghfossa (approved):
LGTM on first pass. 

I did not perform any QA on my machine. I will do that second part of this PR since it's branched off from here. 

On Syft - should we using target directory instead of current?

src/App/Fossa/Compatibility.hs
scruffystuffs: When running syft, there is no target directory.  The positional argument to the command line is a a path or name that syft resolves itself.

scruffystuffs:
When running syft, there is no target directory.  The positional argument to the command line is a a path or name that syft resolves itself.

src/App/Fossa/Configuration.hs
cnr: Yeah, looks like you've [ticketed this already](https://github.com/fossas/team-analysis/issues/692) 😬

src/App/Fossa/Compatibility.hs
scruffystuffs: This is getting removed soon, let's not try to fix it.

scruffystuffs (approved):
This looks good to me.  I haven't taken a deep look at the TH code, but I figure that's the safest code, since if there's an issue, it'll cause a compile error.  And if it works, it's testable by running `--debug` mode.

PR #359, opened 2021-09-03
Minor: Includes lonely dependency in yarn v1 lock graphing

meghfossa:
# Overview

Yarn V1 lock analyses excluded dependency which had no outgoing, or incoming edges. This PR ensures, we include those dependencies in graphing analyses. 

This is prereq for: https://github.com/fossas/team-analysis/issues/723

## Acceptance criteria

- Test passes
- Dependency without any edges are included in the graphing

## Testing plan

- You will have to rely on unit tests. 
- Create yarn project, add a dependency which has no direct dependencies (e.g. https://github.com/sindresorhus/decamelize). Perform analyses. 

## Risks

N/A

## References

Works on https://github.com/fossas/team-analysis/issues/723

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Spectrometer Changelog
 
+## v2.15.14
+
+- Yarn: Fixes missing dependency from the analyses, when dependency has zero deep dependencies, and is not a deep dependency of any other dependency. ([#359](https://github.com/fossas/spectrometer/pull/359))
+
 ## v2.15.13
 
 Adds another closed beta feature around FOSSA C/C++ support.
--- a/src/Strategy/Yarn/V1/YarnLock.hs
+++ b/src/Strategy/Yarn/V1/YarnLock.hs
@@ -26,8 +26,16 @@ analyze lockfile = context "Lockfile V1 analysis" $ do
     Right parsed -> context "Building dependency graph" $ pure (buildGraph parsed)
 
 buildGraph :: YL.Lockfile -> Graphing Dependency
-buildGraph lockfile = Graphing.edges (concatMap (edgesForPackage . first NE.head) (MKM.toList lockfile))
+buildGraph lockfile =
+  Graphing.edges (concatMap edgesForPackage packagesList)
+    <> Graphing.deeps (map toDependency packageWithoutOutEdges)
   where
+    packagesList :: [(YL.PackageKey, YL.Package)]
+    packagesList = map (first NE.head) $ MKM.toList lockfile
+
+    packageWithoutOutEdges :: [(YL.PackageKey, YL.Package)]
+    packageWithoutOutEdges = filter (null . YL.dependencies . snd) packagesList
+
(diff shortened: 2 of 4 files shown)


scruffystuffs:
@cnr is this another case for not pruning unreachable deps globally?

cnr:
@scruffystuffs @meghfossa 
> is this another case for not pruning unreachable deps globally?

yes, but in this case graph pruning doesn't happen: [we only prune unreachable deps when there are direct deps in the graph](https://github.com/fossas/spectrometer/blob/61c4fc2e1f63710ebd083426d147b64c04d9af73/src/App/Fossa/Analyze/Project.hs#L25-L27), and yarn v1 doesn't set any deps as direct

so as far as I know
> BUT it has no impact on final graphing produced as we prune unreachable dependencies.

..is not correct

scruffystuffs:
Ah, it will have an effect when yarn + npm merge lands

meghfossa:
> > BUT it has no impact on final graphing produced as we prune unreachable dependencies.
> 
> ..is not correct

You are correct here Connor - it did produce lonely dependency in the final build graph. I've corrected the description to reflect this.

liftM (approved):
LGTM.

I agree with @cnr's assessment that the medium-term fix here is to refactor pruning to be an analyzer-specific utility. That might be a worthwhile follow-up PR or ticket @meghfossa (cc @zlav).

liftM:
On an unrelated note, MKMs are such a bizarre abstraction. Can't map keys _already_ point to the same value? What's the motivation for "linking" keys together?

:shrug:

liftM:
Wait, I was just looking over this a second time - why does this work? I don't see where dependencies actually get marked as direct. Why aren't all of the dependencies in the `Graphing` pruned?

liftM:
A more precise hack might be to mark all packages that have no predecessors (i.e. in-degree of zero) as direct. If the lockfile is in a good state, then I think this will never accidentally mark a deep dep as direct (although it will miss some direct deps that are in the middle of the graph).

Still, getting the graph edges perfect isn't that important as long as the list is complete - it's probably more worthwhile to fix the pruning sooner rather than be more precise with this fix.

cnr:
@liftM

> Wait, I was just looking over this a second time - why does this work? I don't see where dependencies actually get marked as direct. Why aren't all of the dependencies in the Graphing pruned?

we only prune dependencies iff direct dependencies exist in the graph (again adding to the confusion). in the yarn-lock case, _there are no direct deps_, because `yarn.lock` only gives us edges:

(pseudocode)
```
pkgA:
  deps:
    - pkgB

pkgB

pkgD
```

so to clarify what this PR fixes, because it's not related to graph pruning: there are some dangling dependencies without edges in or out (e.g., `pkgD` in the example above), and we previously missed them because we (read: I, the original dumb author of this strategy) were only constructing graphs with the `edge` operator

> A more precise hack might be to mark all packages that have no predecessors (i.e. in-degree of zero) as direct.

Agree that's a decent hack to surface many direct deps; would prefer to just do the right thing and look at package.json

PR #368, opened 2021-09-14
[DNM] Overhaul documentation

liftM:
This PR overhauls the documentation for FOSSA CLI.

Some goals:

- Any technical user should be able to understand most of FOSSA end-to-end after reading this.
- Common questions that we encounter during integrations are answered, and the answers are easy to find.
- Every piece of documentation is discoverable by clicking URLs from the main README.md.


liftM:
To get remaining TODOs, run `rg --type markdown TODO`.

docs/concepts/dependencies.md
```diff
@@ -0,0 +1,5 @@
+- What is a dependency?
+  - Dependency as compilation unit
+  - Dependency as license tuple
+  - Dependency as runtime component
```
scruffystuffs, line 4: Maybe add "Dependency as instruction to package manager" or similar

README.md
```diff
@@ -1,75 +1,168 @@
...
+- [Specifying vendored dependencies](./docs/walkthroughs/specifying-vendored-deps.md)
+- [Adding manual dependencies](./docs/walkthroughs/integrating-with-custom-tools.md)
 
 ## Reporting Issues
 
-If you are experiencing an issue related to the results on the FOSSA
-website/dashboard, please contact [support@fossa.com](mailto:support@fossa.com)
+If you've found a bug or need support, the best way to get support is to email [support@fossa.com](mailto:support@fossa.com).
+
+Make sure to include reproduction steps and any relevant project files (e.g. `pom.xml`s, `package.json`s, etc.).
```
kitified, line 162: Do we want to also suggest they preemptively provide a debug bundle or at least a `--debug` log?

docs/references/strategies/languages/python/pipenv.md
```diff
@@ -0,0 +1,339 @@
...
+dependencies. This will fail unless `pipenv install` has been run in that directory.
+
+## Limitations
+
+- Pipfile.lock does not report edges, if `pipenv graph --json-tree` fails, we will
+not be able to detect the edges between dependencies.
+
+  - `pipenv graph --json-tree` relies on `pipenv install` being run in the directory beforehand.
+  If that command was not run (such as in a freshly cloned repo), then the graph command will fail,
+  which means that pipenv analysis will succeed, but with incomplete data (missing edges)
```
kitified, line 33: I think this document largely predated this PR, but I think something that would be useful to add as a documentation guideline is to clarify to users what issues concretely mean for them when we talk about the issue.

For example, if I'm a user the result of "missing edges" is unclear. I think it means "all the dependencies will appear as direct in the UI"; if this is correct that is probably useful information to provide (either inline, or as a link to a troubleshooting document where we explain the behavior and underlying issue).

kitified (commented):
I really appreciate the new organization system and think it's really comprehensive!
Left a couple comments - I think this is great.

docs/walkthroughs/debugging-your-integration.md
```diff
@@ -0,0 +1,22 @@
+### `fossa analyze`: Why wasn't my project found?
```
meghfossa, line 1: [no action] This is great!

docs/glossary.md
```diff
@@ -0,0 +1,25 @@
+# Table of Contents
+
+# Dependency topics
+
+- Direct
+- Transitive
```
meghfossa, line 6: We should use the Deep term for consistency with UI

liftM: Maybe - I'm considering shifting the UI to use "transitive" instead. "Transitive" is more right, although "deep" tends to be more intuitive for non-technical users.

liftM:
Leaving some other notes from teammates here before I forget:

via @meghfossa:
- [ ] Add a language/tool table with direct links to documentation
- [ ] Add a walkthrough for generating debug logs for reporting issues (so we can send this to customers when we get tickets)

via @rlfagan:
- [ ] Mention that `fossa analyze` never uploads your code (this is important for some users)
- [ ] Add removal instructions

docs/glossary.md
zlav: Yea I would agree that "deep" is more intuitive for non-technical users.

README.md
```diff
@@ -1,75 +1,168 @@
...
+
+- **Double-check your results.** Some analysis methods may produce partial or unexpected results depending on what information was available when you ran the analysis. If something seems wrong, [our debugging guide](./docs/walkthroughs/debugging-your-integration.md) can help you diagnose and debug your integration.
+
+- **Scan for issues and generate a compliance report.** Once your analysis is ready, we'll automatically queue an issue scan and report the results in the web application. If you want to try out different issue scanning configurations, you can [change your policy](TODO) or [run a manual issue scan](TODO). Once an issue scan is complete, you can also [generate a report](TODO) from the web application.
+
+- **Set up FOSSA in your CI.** You can also use your issue scan results as inputs to CI scripts. For GitHub repositories, you can use FOSSA's [native GitHub integration](TODO) to report a status check on your PRs. For other CI integrations, you can also [use `fossa test`](TODO) to get programmatic issue status in CI.
+
+### Using Spectrometer as a standalone tool
```
zlav, line 135: I don't see a great reason to include this section. I'd prefer to omit this section rather than field any questions from customers who get the idea that they want to try this. I'm open to hearing why this is beneficial though.

PR #372, opened 2021-09-16
Binary deps: unpack jar

kitified:
# Overview

* Introduce the concept of different strategies that can resolve dependency information for binaries discovered in the source tree during a scan with `--experimental-enable-binary-discovery`.
* Introduce the first such strategy, which resolves JAR files.

### Binary resolution strategies

The idea here is that some binaries (JARs are a great example) are relevant for a user to discover as a binary dependency, but contain information we could use to pre-populate some or all of the dependency information.

Similarly, more advanced strategies in the future could potentially use our open source fingerprint database to resolve binaries we've fingerprinted, or use static binary analysis tools for other known languages.

Strategies are always optional, and eventually fall back to the current binary discovery functionality added in #353.

### JAR resolution

This is heavily based on the Ant analysis in CLIv1; I expect that as customers use this feature we will tweak it as needed.

For this strategy we:
1. Unpack the JAR
2. Search inside `META-INF` for a `pom.xml`; in the case of multiple we pick the one with the shortest path. If we find one, we parse it and use its data for the dependency.
3. Otherwise, search inside for `META-INF/MANIFEST.MF`. If we find one, we parse it and use its data for the dependency.

## Acceptance criteria

* Upgrade binary discovery to pre-populate information for discovered JAR files.

## Testing plan

Aside from automated testing, I:

* Placed several JARs in a project (pushed to [go/cpp-vsi-demo](http://go/cpp-vsi-demo) on the `with-jars` branch).
* Ran `fossa analyze --experimental-enable-binary-discovery` on the project.
* Validated that the results appear as expected in both the UI and with `--output`.

## Risks

This operation slows down `--experimental-enable-binary-discovery`, but since this is opt-in I think this is not very risky.

## Checklist


src/App/Fossa/BinaryDeps/Archive.hs
```diff
@@ -0,0 +1,34 @@
...
+import Control.Algebra (Has)
+import Control.Effect.Exception (bracket)
+import Control.Effect.Lift (Lift, sendIO)
+import Path (Abs, Dir, File, Path, fromAbsDir, fromAbsFile, mkRelDir, (</>))
+import Path.IO (getTempDir, removeDirRecur)
+
+-- | Use the provided extraction function to extract the zip, then run the callback.
+-- When the callback is complete, clean up the temporary directory holding the archive contents.
+withArchive :: Has (Lift IO) sig m => (Path Abs File -> Path Abs Dir -> m ()) -> Path Abs File -> (Path Abs Dir -> m c) -> m c
+withArchive extract archive = bracket (extractArchive extract archive) cleanupDir
```
kitified, line 15: This module shares some duplication with `Discovery.Archive`. I opted to extract this functionality because:

* The `Discovery.Archive` functionality seems particularly suited for expanding during walk operations; I wanted something that could more easily be called on on specific files.
* I wanted to be able to return data from within the closure, which `Discovery.Archive` doesn't currently do (and I didn't want to mess with something so central if I could avoid it).

That being said I'm definitely open to refactoring this into something better if anyone has strong opinions and/or suggestions how to overcome these concerns.

scruffystuffs: > and I didn't want to mess with something so central if I could avoid it

In haskell, due to the strong typing, this is often less of a concern.  Although it can be expensive, if you change the types on purpose to break things, you'll know exactly what to change in the rest of the codebase.  Like rust, haskell tries to be a "refactor without fear" language.  Especially the way we use it.

kitified: Fair enough. I'll try just modifying `Discovery.Archive` instead.

src/App/Fossa/BinaryDeps/Jar.hs
```diff
@@ -0,0 +1,106 @@
...
+    then logDebug $ "Found multiple pom.xml files: " <> pretty (Text.intercalate "; " $ map (renderRelative archive) poms)
+    else pure ()
+  pom <- fromMaybeText "No pom.xml files found" $ choosePom poms
+  logDebug $ "Chose representative pom.xml: " <> pretty (renderRelative archive pom)
+  parsePom pom
+
+choosePom :: [Path Abs File] -> Maybe (Path Abs File)
+choosePom [] = Nothing
+choosePom [pom] = Just pom
+choosePom poms = Just . head $ sortOn (length . toString) poms
```
meghfossa, line 82: [nit] avoid partial function: `head`. There is a ticket in backlog to remove head, tail, etc. 

https://wiki.haskell.org/Partial_functions

scruffystuffs: I think this whole function could be written as `choosePom = listToMaybe . sortOn (length . toString)`.

Though this usage is obviously safe upon reading, it would be nice if it were safe regardless of context.  That's the reasoning behind not using any partial functions.

kitified: That makes total sense. Thanks, updated!

meghfossa (approved):
LGTM. 

I did not test the result in web interface, but I did run it locally with few of provided examples in with-jar branch and inspected output. I tried tiny [jar bomb](https://www.bamsoftware.com/hacks/zipbomb/zbsm.zip) as well, it correctly fingerprinted the file. 

Before you merge, can you add docs on this? Maybe `docs/experimental/--experimental-enable-binary-discovery.md` This will make it easy to ctrl+f for on call, instead of looking at code.

kitified:
@meghfossa 

> Before you merge, can you add docs on this PR or add a ticket in backlog for its docs? Maybe `docs/experimental/--experimental-enable-binary-discovery.md` This will make it easy to ctrl+f for on call, instead of looking at code.

Excellent idea. Added documenting this and other experimental features to the backlog.

PR #373, opened 2021-09-17
Fix archive compression bug

kitified:
# Overview

Fix a bug when using `vendored-dependencies` in `fossa-deps.yml`, where the vendored project was in a subdirectory.

## Acceptance criteria

Support vendored projects in subdirectories.

## Testing plan

1. I created a project with vendored dependencies and a `fossa-deps.yml` file. View it at [go/cpp-vsi-demo](http://go/cpp-vsi-demo) on the branch `with-fossa-deps`.
2. I ran a scan against that project and confirmed that everything now works as expected.

## Risks

None that I can think of in particular.

## Checklist

### DIFF
--- a/src/App/Fossa/ArchiveUploader.hs
+++ b/src/App/Fossa/ArchiveUploader.hs
@@ -97,8 +97,6 @@ arcToLocator arc =
 
 compressFile :: Path Abs Dir -> Path Abs Dir -> FilePath -> IO FilePath
 compressFile outputDir directory fileToTar = do
-  -- Without using `fromAbsDir` for each of these directories, the conversion
-  -- is incorrect. `show outputDir` gives an incorrect result even though it typechecks.
   let finalFile = toString outputDir </> safeSeparators fileToTar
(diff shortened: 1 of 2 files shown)


src/App/Fossa/ArchiveUploader.hs
```diff
@@ -98,7 +99,7 @@ compressFile :: Path Abs Dir -> Path Abs Dir -> FilePath -> IO FilePath
 compressFile outputDir directory fileToTar = do
   -- Without using `fromAbsDir` for each of these directories, the conversion
   -- is incorrect. `show outputDir` gives an incorrect result even though it typechecks.
-  let finalFile = toString outputDir </> fileToTar
+  let finalFile = toString outputDir </> safeSeparators fileToTar
```
scruffystuffs, line 102: [nit]: Can we remove the above comment while we're here?  It was stale when we switched to using `toString`

scruffystuffs (approved):
Looks good to me, code-wise.  Haven't tested, but I'd probably defer to you for testing here, since you found the issue.

One unrelated nit.

PR #375, opened 2021-09-22
Fixes mvn:tree tactic to exclude project as direct dependency for single && multi-module projects

meghfossa:
# Overview

These changes ensures, we remove "project identifier" from dependency listings. 

Currently, for given `pom.xml` file:

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>
 
  <groupId>com.mycompany.app</groupId>
  <artifactId>my-app</artifactId>
  <version>1.0-SNAPSHOT</version>
 
  <properties>
    <maven.compiler.source>1.7</maven.compiler.source>
    <maven.compiler.target>1.7</maven.compiler.target>
  </properties>
 
  <dependencies>
    <dependency>
      <groupId>junit</groupId>
      <artifactId>junit</artifactId>
      <version>4.12</version>
      <scope>test</scope>
    </dependency>
  </dependencies>
</project>
```

We will include `com.mycompany.app:my-app` as root dependency. This is incorrect. Further, we will never be able to resolve the locator with this spec.

This PR aims to address this by shrinking project identifier dependencies. 

## Acceptance criteria

- For single module maven projects, when using `mvn:dependency` tactic, project is not considered a direct dependency
- For multi-module maven projects, when using `mvn:dependency` tactic, root project or sub projects are not considered dependencies (we can't resolve local sub-project as dependency). 

## Testing plan

- Ensure you have maven installed, and is accessible via PATH.

**Single module projects:**

- For single module project, analyze `pom.xml` from https://github.com/fossas/team-analysis/issues/738. Note that project identifier is not included as dependency. 

**Multi-module projects**

```
git clone https://github.com/mkyong/maven-examples.git
cd maven-examples
cd java-multi-modules
fossa analyze -o | jq 
```

Note that project or subproject identifiers are not included as dependency. 

## Risks

N/A

## References

Closes https://github.com/fossas/team-analysis/issues/738

## Checklist

### DIFF
--- a/src/Graphing.hs
+++ b/src/Graphing.hs
@@ -40,6 +40,7 @@ module Graphing (
   pruneUnreachable,
   stripRoot,
   promoteToDirect,
+  shrinkRoots,
 
   -- * Conversions
   fromAdjacencyMap,
@@ -159,6 +160,15 @@ stripRoot (Graphing gr) = Graphing $ AM.overlay newDirectEdges (AM.removeVertex 
 
     newDirectEdges :: AM.AdjacencyMap (Node ty)
     newDirectEdges = AM.edges $ map (Root,) newDirect
+
+-- | Shrinks all root nodes.
+-- Remove current direct nodes. It will promote their immediate children as directs.
+-- Unlike @stripRoot@, it removes them from graphing, instead of preserving them (as node), and their edges.
+shrinkRoots :: forall ty. Ord ty => Graphing ty -> Graphing ty
+shrinkRoots (Graphing gr) = Graphing $ foldr AME.shrinkSingle gr currentDirect
+  where
+    currentDirect :: [Node ty]
+    currentDirect = Set.toList $ AM.postSet Root gr
 
 -- | Add a direct node to this Graphing
 direct :: Ord ty => ty -> Graphing ty
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # Spectrometer Changelog
 
+## v2.15.23
+- Maven: Fixes `mvn:dependency` tactic to exclude root project as direct dependency. ([#375](https://github.com/fossas/spectrometer/pull/375))
+
 ## v2.15.22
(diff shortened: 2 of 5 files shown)


src/Graphing.hs
```diff
@@ -160,6 +161,15 @@ stripRoot (Graphing gr) = Graphing $ AM.overlay newDirectEdges (AM.removeVertex
     newDirectEdges :: AM.AdjacencyMap (Node ty)
     newDirectEdges = AM.edges $ map (Root,) newDirect
 
+-- | Shrinks all root nodes.
+-- Remove current direct nodes. It will promote their immediate children as directs.
+-- Unlike @stripRoot@, it removes them from graphing, instead of preserving them (as node), and their edges.
```
cnr, line 166: I bet this is the intended behavior for `stripRoot`, and it's just implemented wrong. It's probably another case where we implicitly expect `pruneUnreachable` to remove those nodes after `stripRoot`

We should probably replace the existing `stripRoot` with this implementation

cnr (approved):
approach lgtm, left one comment

src/Graphing.hs
meghfossa: Agreed - right now, yarnlock, and cargo is using this strategy. I have added tech-debt ticket in backlog for this https://github.com/fossas/team-analysis/issues/746 (prefer to keep the scope/blast-zone of this PR only related to mvn).

cnr: I'm concerned that by not doing it now we'll never get around to it, unless you plan on doing it as an immediate followup to this

It'll mean one fewer thing for us to worry about when dealing with the tech debt of removing the implicit `pruneUnreachable`

meghfossa: Yeah - I plan to have followup PR this week for this. I've assigned the ticket it to myself.

PR #376, opened 2021-09-23
Refines jsonDeps to support older gradle versions

meghfossa:
# Overview

It correctly queries dependencies of configurations, when working with gradle v3.3 or below.

## Acceptance criteria

- Successfully analyzes project with gradle version below v3.3

## Testing plan

1. Create example project with gradle 2
```
curl -s "https://get.sdkman.io" | bash
source "/Users/megh/.sdkman/bin/sdkman-init.sh"
sdk install gradle 2.7
sdk install java 8.0.292.hs-adpt
gradle init 
```

Add dependencies as you see fit. 

Few example deps:
```
'org.codehaus.groovy:groovy:3.+'
'org.codehaus.groovy:groovy-json:3.+'
'com.codevineyard:hello-world:1.0.1!!'
'org.neo4j.driver:neo4j-java-driver:4.3.4'
```

2. Create example project with gradle 7 (similar to aforementioned gradle2 procedure)

Analyze using `fossa analyze -o | jq`. 

## Risks

N/A. Everything is still under try/catch. I don't anticipate any issues here.

## References

Closes: https://github.com/fossas/team-analysis/issues/740

## Checklist

### DIFF
--- a/scripts/jsondeps.gradle
+++ b/scripts/jsondeps.gradle
@@ -115,7 +115,7 @@ allprojects {
             def configToKeyValue = { config ->
                 def jsonDeps = []
                 // config.resolvedConfiguration: https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/ResolvedConfiguration.html
-                config.resolvedConfiguration.firstLevelModuleDependencies.each { dep ->
+                config.resolvedConfiguration.getFirstLevelModuleDependencies().each { dep ->
                     println "DEBUG: Found direct dependency in configuration"
                     println dep
                     def result = depToJSON dep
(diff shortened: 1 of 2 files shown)


scripts/jsondeps.gradle
```diff
@@ -127,33 +127,53 @@ allprojects {
...
+                    if (config.respondsTo("isCanBeResolved")) {
+                        if (config.isCanBeResolved() == false) {
+                            println "DEBUG: Configuration is not resolvable"
+                            return null
+                        }
+                    }
+
+                    // At this point, we are working against gradle: 
+                    // - v3.3 or greater, and configuration is resolvable.
+                    // - v3.2.1 or lower, and configuration does not have canResolve property and are inherently resolvable.
```
meghfossa, line 156: Although there is no risk here (runtime error, etc.) I will not merge until I have rigorously confirmed this. All indication lead to this conclusion/

scripts/jsondeps.gradle
```diff
@@ -127,33 +127,53 @@ allprojects {
...
-                            println result
-                            jsonConfigs << result
-                        } else {
-                            println "DEBUG: Configuration is not resolvable"
-                        }
-                    } catch (Exception ignored) {
-                        println "DEBUG: An exception occurred"
-                        println ignored
-                        ignored.printStackTrace()
+                    def data = getResolvedConfigDepsOrEmptyOf config
```
meghfossa, line 174: - [ ] better naming

scruffystuffs:
Idk if I'm the right person to review this.  I've never read or written any gradle (groovy?) code, including this script.

meghfossa:
@liftM tagged you here for review since you made the edits most recently

scripts/jsondeps.gradle
```diff
@@ -127,33 +127,53 @@ allprojects {
...
-                            jsonConfigs << result
-                        } else {
-                            println "DEBUG: Configuration is not resolvable"
-                        }
-                    } catch (Exception ignored) {
-                        println "DEBUG: An exception occurred"
-                        println ignored
-                        ignored.printStackTrace()
+                    def jsonConfigWithDeps = getResolvedConfigDepsOrEmptyOf config
+                    if (jsonConfigWithDeps != null) {
```
liftM, line 175: [nit] Might be a tiny bit safer to return `[]` instead of `null` on empty, then we don't need to null-check the result.

scripts/jsondeps.gradle
liftM, line 156: (Sorry to delete your previous comment @meghfossa - GitHub code review otherwise has an irritating bug that does not let me create a multiline comment that ends on a previously commented line.)

Rewording suggestion for clarity:

```suggestion
                    // At this point, we know one of these must be true: 
                    // - We are on Gradle v3.3 or newer (because isCanBeResolved is present) and the configuration is resolvable (isCanBeResolved() == true, therefore did not exit early above).
                    // - We are on Gradle v3.2.1 or older (because isCanBeResolved is not present). Configurations in Gradle v3.2.1 or older are always resolvable.
```

scripts/jsondeps.gradle
```diff
@@ -127,33 +127,53 @@ allprojects {
...
+            // -----------
+            // - https://discuss.gradle.org/t/what-is-a-configuration-which-cant-be-directly-resolved/30721
+            // - https://docs.gradle.org/current/userguide/declaring_dependencies.html#sec:resolvable-consumable-configs
+            def getResolvedConfigDepsOrEmptyOf = { config -> 
+                def configsDeps = null              
+                try {
+                    // We know that we are working against gradle version greater or equal to v3.3
+                    // So, configuration has resolvable property. If the configuration is not resolvable return null.
+                    if (config.respondsTo("isCanBeResolved")) {
+                        if (config.isCanBeResolved() == false) {
```
liftM, line 148: [nit] I might combine these for clarity? I'm not sure though.

```suggestion
                    if (config.respondsTo("isCanBeResolved") && !config.isCanBeResolved()) {
```

liftM (approved):
LGTM. Test plan seems reasonable, although I did not try to replicate.

scripts/jsondeps.gradle
meghfossa: I'm leaning towards keeping `null` instead of using `flatten` and `filter`. I prefer not to do unnecessary transformations (if it can be avoided), and keeping the code for quick readability for even groovy.

PR #374, opened 2021-09-21
[WIP] Yarn npm merge

scruffystuffs:
# Overview

TBD

## Acceptance criteria

TBD

## Testing plan

TBD

## Risks

TBD

## References

closes fossas/team-analysis#707

## Checklist

- [ ] I added tests for this PR's change (or confirmed tests are not viable).
- [ ] If this PR introduced a user-visible change, I added documentation into `docs/`.
- [ ] I updated `Changelog.md` if this change is externally facing. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.
- [ ] I linked this PR to any referenced GitHub issues, if they exist.

PR #380, opened 2021-09-30
dep envs are now sets

scruffystuffs:
# Overview

Change `dependencyEnvironments` to be a `Set`, as opposed to a list.  This is needed for #374.

Part of the work for fossas/team-analysis#707

This change is REALLY simple, and only has one logical change, which is to use `Set` instead of list, but we use this everywhere, so the diff seems huge.

## Checklist


meghfossa (approved):
LGTM - looks straight forward.

PR #396, opened 2021-10-15
Discovered binaries reports origin path

kitified:
# Overview

Reports an origin for uploaded `SourceUserDefDep`s.

The background here is that Motorola requested a feature which we call "binary discovery" ([reference documentation](https://www.notion.so/fossa/Binary-Discovery-5a1afa2e9ee64c48ae822e870e7cf06b)). The TL;DR of this feature is that we, as a strategy in Spectrometer, identify files with binary content in the project and report them as unlicensed user-defined dependencies in Core.

This, maybe predictably, led to a follow up feature request from Motorola: "please display the path at which the binary is located _in the report under the `File Path(s)` column_". This PR makes that happen.

"Origin", in this context, means the path at which the user defined dependency was identified. It's optional on the type and if provided, hydrates `origin_paths` in the dependency lock.

This then is used to provide the actual point of the PR, which is hydrating `File Path(s)` in the report for the project when using binary discovery.

## Acceptance criteria

I booted a core instance that supports this information and ran a scan.
I then verified that:

* Origin paths are displayed in the UI (when they're clicked they just say "this file wasn't uploaded", but that's fine).
* The report correctly hydrates the `File Path(s)` column with this origin.

## Testing plan

1. Boot core with the branch `discovered-binary-origins`. Make sure the "File paths in reports" feature flag is enabled.
2. Create a project with a binary file inside (or add a binary file to a project you already have).
3. Run a scan with Spectrometer on this branch with the flag `--experimental-enable-binary-discovery`.
4. Run a report, selecting to display the file paths. Validate that the file path of the binary appears in that report.

## Screenshots

Showing this working:

Screen Shot 2021-10-15 at 2 15 43 PM

Screen Shot 2021-10-15 at 2 16 03 PM

And showing that user defined deps _without_ origins are still displayed properly:

Screen Shot 2021-10-15 at 2 17 33 PM

## Risks

Not particularly risky.

## Checklist

### DIFF
--- a/src/Path/Extra.hs
+++ b/src/Path/Extra.hs
@@ -1,6 +1,7 @@
 module Path.Extra (
   tryMakeRelative,
   renderRelative,
+  renderSomeBase,
 ) where
 
 import Data.String.Conversion (toText)
@@ -18,6 +19,10 @@ tryMakeRelative absDir absFile =
 -- | Render the relative path between a Path Abs Dir and a Path Abs File that is supposed to be in that dir.
 -- Intended for convenience when displaying the newly relative path; to interact with it use `tryMakeRelative` instead.
 renderRelative :: Path Abs Dir -> Path Abs File -> Text
-renderRelative absDir absFile = case tryMakeRelative absDir absFile of
+renderRelative absDir absFile = renderSomeBase $ tryMakeRelative absDir absFile
+
+-- | Render the file regardless of the kind of base.
+renderSomeBase :: SomeBase f -> Text
+renderSomeBase path = case path of
   Abs p -> toText . toFilePath $ p
   Rel p -> toText . toFilePath $ p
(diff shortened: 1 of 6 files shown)


src/Path/Extra.hs
```diff
@@ -18,6 +19,10 @@ tryMakeRelative absDir absFile =
 -- | Render the relative path between a Path Abs Dir and a Path Abs File that is supposed to be in that dir.
 -- Intended for convenience when displaying the newly relative path; to interact with it use `tryMakeRelative` instead.
 renderRelative :: Path Abs Dir -> Path Abs File -> Text
-renderRelative absDir absFile = case tryMakeRelative absDir absFile of
+renderRelative absDir absFile = renderSomeBase $ tryMakeRelative absDir absFile
+
+-- | Render the file regardless of the kind of base.
+renderSomeBase :: SomeBase f -> Text
+renderSomeBase path = case path of
```
scruffystuffs, line 26: [optional nit]: I think, for this, I would prefer if `SomeBase` had instances for `ToString` and `ToText` in `Data.String.Conversion`.  Then we could just use `toString` or `toText` everywhere.  But this isn't necessary to change in this PR.

scruffystuffs (approved):
Looks good, left one comment

PR #397, opened 2021-10-19
Compatibility script: `fossa report attribution --json`

jssblck:
# Overview

Adds a compatibility script for `fossa attribution --json`.

## Acceptance criteria

* The compatibility layer transforms the report to use the same key structure and content.

The sole exception to this is that FOSSAv1 sometimes outputs `Notes` keys (which are arrays of strings) with a single zero-value string, but sometimes not. The FOSSAv2 version of the report consistently reports an empty array in these cases, so we don't really have a way to reproduce this exactly.

Example:

Screen Shot 2021-10-18 at 5 09 40 PM

## Testing plan

* Run a scan on a project.
* Run `fossa report attribution --json` on that same project using both FOSSA v1 and v2. Save their outputs to disk and note the differences.
* Run `fossa report attribution --json | go build scripts/compat-attribution/main.go` using FOSSA v2. Save its output to disk.
* Note that the difference between the compat version of the output and the v1 output is within AC.

## Risks

Not very risky.

## Potentially Contentious Decisions

I set up a Go project at the root of Spectrometer to make this happen.

I originally spent a fair amount of time trying to hack this with something like `jq` in pure bash because I wanted to avoid this, but it quickly became a mess. Rather than go down that route I decided that this was the more reasonable option. Happy to discuss it with the team if there are strong objections though.

This is also why I moved the `vendor/` folder to `vendor-bins/`: the existence of a `vendor/` folder made Go very sad indeed.

## Checklist

Also I didn't include any tests because:

* The operation of this script is _extremely_ straightforward and mechanical.
* I don't see a good way for us to integration test against an actual FOSSA instance, but if I'm missing something I'd love to be shown how I can do it.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,8 +1,9 @@
 # Spectrometer Changelog
 
-## Unreleased
+## v2.17.3
 
-- Adds compatibility script for `fossa report attribution --json` ([#397](https://github.com/fossas/spectrometer/pull/397))
+ - Monorepo: adds some optimizations to reduce the amount of file buffering in memory during a scan, resulting in less memory pressure and faster scans. ([#402](https://github.com/fossas/spectrometer/pull/402))
+ - Adds compatibility script for `fossa report attribution --json` ([#397](https://github.com/fossas/spectrometer/pull/397))
 
 ## v2.17.2
 
--- a/vendor_download.sh
+++ b/vendor_download.sh
@@ -45,7 +45,7 @@ esac
 TAG="latest"
 echo "Downloading asset information from latest tag for architecture '$ASSET_POSTFIX'"
 
-WIGGINS_TAG="2021-10-14-943957b"
+WIGGINS_TAG="2021-10-21-8eb009f"
 echo "Downloading wiggins binary"
 echo "Using wiggins release: $WIGGINS_TAG"
 WIGGINS_RELEASE_JSON=vendor-bins/wiggins-release.json


kitified:
@meghfossa removed your request for review for now because these build scripts are causing me grief. Once I have it figured out I'll re-request from you!

meghfossa:
@kitified Does attribution require specific permission or FF enabled for the org? (I'm getting FOSSA API: no permission) error - this is with my free seat account on app.fossa.com.

kitified:
@meghfossa you have to run it with a full token, otherwise I'm not aware of anything special.

.gitignore
```diff
@@ -9,7 +9,7 @@ dist-newstyle/
 *.hp
 
 # Development files/directories
-vendor/
+vendor-bins/
 .vscode/
```
meghfossa, line 13: should vendor directory be ignored as well?

Changelog.md
```diff
@@ -1,5 +1,9 @@
 # Spectrometer Changelog
 
+## Unreleased
```
meghfossa, line 3: Let's have a dedicated release since, we are adding new executable in distribution

.github/workflows/build.yml
```diff
@@ -174,14 +224,17 @@ jobs:
         chmod +x Linux-binaries/*
         zip -j release/fossa_${{ steps.get-version.outputs.VERSION }}_linux_amd64.zip Linux-binaries/fossa
         zip -j release/pathfinder_${{ steps.get-version.outputs.VERSION }}_linux_amd64.zip Linux-binaries/pathfinder
+        zip -j release/compat-attribution_${{ steps.get-version.outputs.VERSION }}_linux_amd64.zip Linux-binaries/compat-attribution
 
         chmod +x macOS-binaries/*
         zip -j release/fossa_${{ steps.get-version.outputs.VERSION }}_darwin_amd64.zip macOS-binaries/fossa
         zip -j release/pathfinder_${{ steps.get-version.outputs.VERSION }}_darwin_amd64.zip macOS-binaries/pathfinder
+        zip -j release/compat-attribution_${{ steps.get-version.outputs.VERSION }}_linux_amd64.zip macOS-binaries/compat-attribution
```
meghfossa, line 232: I have not tested this bit - but should this end with `_darwin_amd64.zip macOS-binaries/compat-attribution` instead?

.github/workflows/build.yml
```diff
@@ -174,14 +224,17 @@ jobs:
...
 
         chmod +x Windows-binaries/*
         zip -j release/fossa_${{ steps.get-version.outputs.VERSION }}_windows_amd64.zip Windows-binaries/fossa.exe
         zip -j release/pathfinder_${{ steps.get-version.outputs.VERSION }}_windows_amd64.zip Windows-binaries/pathfinder.exe
+        zip -j release/compat-attribution_${{ steps.get-version.outputs.VERSION }}_linux_amd64.zip Windows-binaries/compat-attribution.exe
```
meghfossa, line 237: I have not tested this bit - but should this end with `_windows_amd64.zip Windows-binaries/compat-attribution.exe` instead?

meghfossa (approved):
Please look at comments on Github Build workflow before merging!

Other than that LGTM. 

- Can you also add issue in backlog to remove this mode post April 2022?
- It would nice if you can modify Makefile to build `compat-attribution` in this PR as well.

PR #398, opened 2021-10-20
Add Scoop install instructions

rolodato:
Available on the main bucket after https://github.com/ScoopInstaller/Main/pull/2778 was merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,12 @@ Set-ExecutionPolicy Bypass -Scope Process -Force; iex  ((New-Object System.Net.W
 **NOTE**: You may need to add the executable to your PATH. The installer
 reports the installation path of the executable.
 
+Alternatively, install using [Scoop](https://scoop.sh/):
+
+```
+scoop install fossa
+```
+
 ## Using Spectrometer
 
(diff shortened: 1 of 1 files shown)


meghfossa:
[nit] Can we add test in CI to ensure, if scoop integration breaks for whatever reason (different uri scheme, etc), we have some visibility, since this change adds it to public docs.

I have not tested this (may need to modify it)

```yaml
jobs:
  scoop-install-test:
    name: Scoop Install Test
    runs-on: windows-latest
    steps:
      - name: Install using Scoop
        shell: powershell
        run: |
          scoop install fossa

      - name: Test Fossa Install
        shell: powershell
        run: |
          fossa --version

```

rolodato:
@meghfossa I agree that we should have some kind of check, I'd need to do some research on how it works (never actually ran Windows on GHA before).

That being said, we hardly ever modify the Windows installation or build so it's unlikely that this integration will break any time soon. If it does it will most likely affect only desktop users and not CI/CD pipelines. Would you be OK with merging this now to make it known to users ASAP and we can create a backlog item for verifying this in CI? Or if you think you can figure out how to verify this in CI easier, feel free to have a go :)

PR #402, opened 2021-10-21
Bump Wiggins Tag: Optimization in Indexers during scan

gtulipani:


Changelog.md
```diff
@@ -1,5 +1,9 @@
 # Spectrometer Changelog
 
+## v2.17.3
+
+ - Monorepo: Optimization during Scan to avoid passing the whole file content through the code. ([#402](https://github.com/fossas/spectrometer/pull/402))
```
kitified, line 5: We try to make the changelog read like a sentence (refer to other releases for examples of this).

[optional] Also, I think (personal opinion) it's great if changelogs don't get too technical, but do include some detail for technical minded folks when possible. A great thing to keep in mind is all users should be able to answer "why do I care about this release", and interested users should be able to answer "in a very broad sense, how does this release accomplish this"

As a suggestion, consider this:
```suggestion
 - Monorepo: adds some optimizations to reduce the amount of file buffering in memory during a scan, resulting in less memory pressure and faster scans. ([#402](https://github.com/fossas/spectrometer/pull/402))
```

gtulipani: Amazing, thanks!

PR #403, opened 2021-10-21
Limit capabilities to respect cgroups cpu quota

cnr:
# Overview

## Capabilities

For speed, we poll the environment for the number of available "capabilities", which GHC defines as "the number of Haskell threads that can run truly simultaneously (on separate physical processors)", and we use that number to spawn worker threads for discovery/analysis tasks.

Previous to this PR, we were letting the GHC runtime automatically set the number of available capabilities [with the `-N` flag](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/using-concurrent.html#rts-options-for-smp-parallelism).

## Containers

Containers, especially those run in a cluster environment, are often run with limits for the amount of CPU they're allowed to consume. Linux cgroups ("control groups") presents these limits to the container as `cpu.cfs_quota_us` and `cpu.cfs_period_us`, [documented here](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/resource_management_guide/sec-cpu).

In short, `quota / period` forms the ratio of cpu time we're allowed to consume, e.g.,

| quota | period | description |
| --- | | |
| 10 | 10 | we can consume up to one core |
| 20 | 10 | we can consume up to two cores
| 5 | 10 | the scheduler will give us one cpu core for up to 50% of the time |
| -1 | 10 | we can use all of the machine's available cpu resources | 

## The problem

Unfortunately, rather than propagating cgroup cpu limits to the container's view of the hardware of the machine, the container will happily report that it has _all_ of the available cores on the host machine.

As such, when GHC automatically sets the number of capabilities for us in a cgroups-cpu-limited container on a large machine, we spawn _way_ too many worker threads relative to the amount of CPU time allocated to us. When the disparity is particularly large, we end up with a lot of context-switching and contention. Threads spend a lot of time getting woken up before momentarily getting throttled and yielding to the scheduler.

## The solution

Rather than letting the GHC runtime set the number of capabilities for us with the `-N` flag, this PR adds a new `initCapabilities` that sets the number of available capabilities while respecting cgroup limits, when applicable 

## Acceptance criteria

- Within a linux container, cgroups cpu limits are respected, and the number of capabilities is set based on cgroups limits, when applicable
- Outside of a linux container, CLI behavior is unchanged
- On macOS and Windows, CLI behavior is unchanged

## Testing plan

- Manual testing for (1), (2), (3) - in particular with print debugging on `getNumCapabilities` to ensure they're getting set correctly
- (1) was additionally tested within a customer's target environment
- Tests for cgroup/mountinfo parser logic and cgroups quota logic (TODO: need to add this)

## Risks

N/A

## References

https://fossa.zendesk.com/agent/tickets/3198

## Checklist


test/System/CGroup/TypesSpec.hs
```diff
@@ -0,0 +1,53 @@
+{-# LANGUAGE CPP #-}
```
scruffystuffs, line 1: I'd really like to avoid doing anything with CPP if possible.  Maybe we could use a wrapper around `it` which does nothing on windows.

PR #428, opened 2021-11-02
Adds perl support

meghfossa:
# Overview

Adds basic perl support.

## Acceptance criteria

- _Built_ perl project can be analyzed, using  `(MYMETA|META).{json|yml}`

## Testing plan

- Get example perl project (https://github.com/codecov/example-perl)
- Build perl project (mentioned steps on repository)
- Run fossa cli (`fossa analyze example-perl -o`)

## Risks

N/A

## References

Closes https://github.com/fossas/team-analysis/issues/794

## Checklist

- [x] I updated `*schema.json` if I have made changes for `.fossa.yml`, `fossa-deps.{json, yaml, yml}`. You may also need to update these if you have added/removed new dependency (e.g. pip) or analysis target type (e.g. poetry).


docs/references/strategies/languages/perl/perl.md
```diff
@@ -0,0 +1,38 @@
...
+1. Parse `MYMETA.{yml, json}` or `META.{yml, json}` to identify dependencies.
+
+## Limitation
+
+- Dependency required for `runtime` only will be reported.
+- Reported analysis will not have any edges.
+
+## Example 
+
+1. Build using: `perl Makefile.PL`. When you do this, you'll see that `MYMETA.yml` and `MYMETA.json` are generated.
```
meghfossa, line 22: - [ ] To update with more scenario or to remove the build instruction

src/Strategy/Perl.hs
```diff
@@ -0,0 +1,202 @@
...
+            version
+            runtimeRequires
+            buildRequires
+            testRequires
+            developRequires
+            configureRequires
+
+      -- Reference: http://module-build.sourceforge.net/META-spec-v1.4.html
+      parseBelowV1_5 obj version = do
+        runtimeRequires1 <- obj .:? "requires"
```
zlav, line 169: I don't think you need to make this `runtimeRequires1` since it's in a different do block from the one above.
```suggestion
        runtimeRequires <- obj .:? "requires"
```

test/Perl/PerlSpec.hs
```diff
@@ -0,0 +1,70 @@
...
+  shouldBe,
+ )
+
+perl :: (PackageName, Maybe Text)
+perl = (PackageName "perl", Just "5.006")
+
+expectedContentFromJson :: PerlMeta
+expectedContentFromJson =
+  PerlMeta
+    { version = 2.0
```
zlav, line 27: Would it make sense to have a test for the 2.0 parser for JSON and Yaml? As well as for 1.4?

zlav (approved):
LGTM. These parsers are all starting to look the same to me which is a really good thing.

test/Perl/PerlSpec.hs
meghfossa: In current test suite, we have Meta.yml for 1.4 version, and Meta.json for 2.0 The way Data.Yaml works, is that it's very thin layer on top of aeson, so I don't anticipate any issues. 

Nonetheless, I've added test for posterity.

PR #431, opened 2021-11-02
add checksums for zip files

scruffystuffs:
Update GH actions release flow to post checksums of the zip files we host.  This can help confirm that the version a customer downloads is the same version that was posted in the release.  While this does not mitigate our anything in our main threat model (which is a malicious actor gaining access to the repo on Github itself, from which they could spoof these checksums), the posted checksum can be saved, and later used as proof that the release artifacts have not changed.


zlav (approved):
This LGTM. I tested `sha256sum` and validated that it produces the output we need.

PR #432, opened 2021-11-02
Run the checksum from inside the release dir

scruffystuffs:
sha256sum outputs checksums in the format of `somehash path/to/file`.
When a user downloads `somezipfile` and `somezipfile.sha256`, we want to be able to run `sha256sum somezipfile.sha256`.  However, to get the correct path name in the checksum file, we have to run from the downloader's POV, which means that `somezipfile` needs to be in the current directory when the sum is calculated.

closes fossas/team-analysis#801


zlav (approved):
LGTM

PR #434, opened 2021-11-02
Compute checksums in binary mode and sanity-check them

scruffystuffs:
Binary mode only makes a difference on platforms with a real difference between binary and test files.  MS-DOS (not windows!) is cited as an example in multiple articles, but if anyone is using spectrometer on MS-DOS I think we can safely say it's not supported.


zlav (approved):
LGTM

PR #436, opened 2021-11-04
Docs: Adds test-development walkthrough documentation

meghfossa:
# Overview

Adds walkthrough for test and development dependencies.

### DIFF
--- a/docs/walkthroughs/test-dev-dependency.md
+++ b/docs/walkthroughs/test-dev-dependency.md
@@ -1,0 +1,113 @@
+# Working with Test and Development Dependency
+
+FOSSA CLI, by default, does not include the unused dependency in analysis. This is expected for scenarios in which, you have a dependency for development, testing, and other purposes - which ultimately are not used in your runtime. FOSSA CLI can infer dependency's inclusion for runtime or distribution by package manager's manifest file. When we are unable to infer if the dependency is only used in runtime, we choose the conservative approach and include them in our analysis reporting. 
+
+CLI's language support documentation (in [strategy's section](./../references/strategies/README.md)), includes whether an analysis of a particular project will by default unused dependency.
+
+## Example 
+
+For the following `package.json` file:
+
+```json
+{
+    "name": "example-project",
+    "dependencies": {
+        "pngjs": "6.0.0",
+        "once": "1.4.0"
+    },
+    "devDependencies": {
+        "mocha": "*"
+    }
+}
+```
+
+### When running `fossa analyze`:
+
+We will report the dependency tree as follows:
+
+```tree
+.
+ |-npm+once$1.4.0 (direct)
+ | \-npm+wrappy$1 (transitive)
+ \-npm+pngjs$6.0.0 (direct)
+```
+
+Note that, we are not reporting "mocha" dependency or any other transitive dependency of mocha. This is intended behavior, as aforementioned, FOSSA CLI, by default only report dependency which is used in your runtime or distribution.
+
+### How to include all dependencies?
+
+You can use the `--include-unused-deps` flag to explicitly include all _discovered_ dependencies found. This will include all unused dependencies (development, configuration, etc.)
+
+You can run analysis like this:
+```bash
+fossa analyze --include-unused-deps
+```
+
+With the flag, it will report the following dependency graph now (notice the inclusion of mocha and its dependencies)
+
+```tree
+.
+ |-npm+mocha$9.1.3 (direct)
+ | |-npm+@ungap/promise-all-settled$1.1.2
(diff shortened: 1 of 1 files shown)


docs/walkthroughs/test-dev-dependency.md
```diff
@@ -0,0 +1,113 @@
...
+
+You can run analysis like this:
+```bash
+fossa analyze --include-unused-deps
+```
+
+With the flag, it will report the following dependency graph now (notice the inclusion of mocha and its dependencies)
+
+```tree
+.
```
meghfossa, line 49: Blocked on:
1. https://github.com/fossas/team-analysis/issues/811
2. https://github.com/fossas/team-analysis/issues/810

docs/walkthroughs/test-dev-dependency.md
```diff
@@ -0,0 +1,113 @@
+# Working with Test and Development Dependency
```
zlav, line 1: [nit] Should be "dependencies".

docs/walkthroughs/test-dev-dependency.md
```diff
@@ -0,0 +1,113 @@
...
+
+FOSSA CLI, by default, does not include the unused dependency in analysis. This is expected for scenarios in which, you have a dependency for development, testing, and other purposes - which ultimately are not used in your runtime. FOSSA CLI can infer dependency's inclusion for runtime or distribution by package manager's manifest file. When we are unable to infer if the dependency is only used in runtime, we choose the conservative approach and include them in our analysis reporting.
```
zlav, line 3: [grammar] I don't think "dependency" can be used as a plural noun.

```suggestion
FOSSA CLI, by default, does not include unused dependencies in analysis. This is expected for scenarios in which, you have a dependency for development, testing, and other purposes - which ultimately are not used in your runtime. FOSSA CLI can infer dependency's inclusion for runtime or distribution by package manager's manifest file. When we are unable to infer if the dependency is only used in runtime, we choose the conservative approach and include them in our analysis reporting. 
```

docs/walkthroughs/test-dev-dependency.md
```diff
@@ -0,0 +1,113 @@
...
+We will report the dependency tree as follows:
+
+```tree
+.
+ |-npm+once$1.4.0 (direct)
+ | \-npm+wrappy$1 (transitive)
+ \-npm+pngjs$6.0.0 (direct)
+```
+
+Note that, we are not reporting "mocha" dependency or any other transitive dependency of mocha. This is intended behavior, as aforementioned, FOSSA CLI, by default only report dependency which is used in your runtime or distribution.
```
zlav, line 35: [nit] confusing sentence
```suggestion
Note that, we are not reporting the "mocha" dependency or any other transitive dependencies of mocha as it is a development dependency. By default, the FOSSA CLI only reports dependencies used in your runtime or distribution.
```

docs/walkthroughs/test-dev-dependency.md
```diff
@@ -0,0 +1,113 @@
...
+
+### How to include all dependencies?
+
+You can use the `--include-unused-deps` flag to explicitly include all _discovered_ dependencies found. This will include all unused dependencies (development, configuration, etc.)
```
zlav, line 39: What does discovered found? I think this would be better if you said something like the following.
```suggestion
You can use the `--include-unused-deps` flag to explicitly include all dependencies, regardless of how they are used. This will include all dependencies (development, configuration, etc.)
```

docs/walkthroughs/test-dev-dependency.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+### When running `fossa analyze`:
+
+We will report the dependency tree as follows:
+
+```tree
+.
+ |-npm+once$1.4.0 (direct)
+ | \-npm+wrappy$1 (transitive)
```
zlav, line 31: Where is `wrappy` in the `package.json`? You may want to make the example file a `package-lock.json` file. I don't think the jump to where the transitive deps are coming from will be immediately obvious to most people reading this doc. This also goes for the dependency tree towards the bottom with dev deps in it. I was confused at first about where all of the other deps came from.

zlav (approved):
Looks pretty good, awesome walkthrough!
My main comments are around using "Dependency" as a plural (although I think we may do this elsewhere) and using a better example file.

meghfossa:
Closing this since, I don't think we will be able to resolve this prior to switch over:

- fossas/team-analysis#811
- fossas/team-analysis#810

Thanks for the review on this, I can apply feedback to the changes, and once blockers are resolved, we should be good to go!

PR #438, opened 2021-11-05
[DNM]: Explicit imports

scruffystuffs:
# Overview

Statically enforcing explicit imports might be okay, except that includes the `import Module (SomeLargeType (..))` form, and theres no exception for the stdlib prelude, which prevents us from doing `import Prelude hiding (filter)` if we wanted to define our own function named filter.  In these cases, we either have to rename the function, or import every used item from the prelude explicitly.

This PR took hours of work, it should have taken minutes.  The tooling only gets us so far, and we have to expand inner `(..)` forms everywhere.  

I would like to get team agreement that this is a bad direction to go in, until either GHC provides a more fine-grained import lint, or HLS gets better import expansion.

This doesn't even fix the test code and still hit 168 files!

### DIFF
--- a/src/Effect/Logger.hs
+++ b/src/Effect/Logger.hs
@@ -22,10 +22,27 @@ module Effect.Logger (
   module X,
 ) where
 
-import Control.Algebra as X
-import Control.Carrier.Simple
-import Control.Effect.ConsoleRegion
-import Control.Effect.Exception
+import Control.Algebra as X (
+  Algebra (alg),
+  Handler,
+  Has,
+  run,
+  send,
+  thread,
+  (~<~),
+  type (:+:) (L, R),
+ )
+import Control.Carrier.Simple (
+  Simple,
+  SimpleC,
+  interpret,
+  sendSimple,
+ )
+import Control.Effect.ConsoleRegion (
+  displayConsoleRegions,
+  withConcurrentOutput,
+ )
+import Control.Effect.Exception (Lift)
 import Control.Effect.Lift (sendIO)
 import Control.Effect.Record (RecordableValue)
 import Control.Effect.Record.TH (deriveRecordable)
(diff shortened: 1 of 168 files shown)


src/Strategy/Haskell/Stack.hs
```diff
@@ -11,22 +11,74 @@ module Strategy.Haskell.Stack (
...
+import Types (
+  DepType (HackageType),
+  Dependency (
+    Dependency,
+    dependencyEnvironments,
+    dependencyLocations,
+    dependencyName,
+    dependencyTags,
+    dependencyType,
+    dependencyVersion
```
scruffystuffs, line 64: This type is EVERYWHERE in the codebase, and we have to manually expand it every time.

src/Effect/Logger.hs
```diff
@@ -22,23 +22,57 @@ module Effect.Logger (
...
+  AnsiStyle,
+  Color (Cyan, Green, Red, White, Yellow),
+  color,
+  renderStrict,
+ )
 import System.Console.ANSI (hSupportsANSI)
 import System.Console.Concurrent (errorConcurrent, outputConcurrent)
 import System.IO (stderr)
-import Prelude hiding (log)
+import Prelude (Applicative, Eq, IO, Ord, Show, String, pure, ($), (.), (<=), (<>))
```
scruffystuffs, line 75: This is disgusting.  We should never have to import any of these.

src/Graphing.hs
```diff
@@ -55,7 +55,25 @@ import Algebra.Graph.AdjacencyMap.Algorithm qualified as AMA
...
+  Ord,
+  Semigroup ((<>)),
+  Show,
+  Traversable (sequenceA, traverse),
+  map,
+  ($),
+  (++),
+  (.),
+  (<$>),
+ )
```
scruffystuffs, line 76: Bruh.

src/Effect/ReadFS.hs
```diff
@@ -38,17 +38,37 @@ module Effect.ReadFS (
...
+import Control.Algebra as X (
+  Algebra (alg),
+  Handler,
+  Has,
+  run,
+  send,
+  thread,
+  (~<~),
+  type (:+:) (L, R),
+ )
```
scruffystuffs, line 50: Also, these were weirdly hard to fix.  When we re-export, we have to check the module that imports this to see if there's an issue.

liftM:
Yeah, I think this might be better as a separate lint than turning on the GHC warning, since I think the `Type(..)` form and `import Prelude` are both acceptable.

zlav:
I agree that we shouldn't merge this PR. I'm going to close it. Thanks for putting the work in to understand what fully requiring explicit imports would require.

PR #445, opened 2021-11-10
Remove `packageVersion` from `NpmPackageJson`

kitified:
# Overview

No longer attempts to parse the `version` field in `package-lock.json`.

This field denotes the version of the project being scanned, which we don't care about. However, `package-lock.json` files without this field fail parsing.

## Acceptance criteria

1. `package-lock.json` files without the `version` field present now parse correctly.
2. The `--output` is the same between a project parsed with the `version` field intact and without the `version` field.

## Testing plan

I used the example `package.json` and `package-lock.json` in the [attached archive file](https://github.com/fossas/spectrometer/files/7515353/Example-Project.zip).

I then:

* Verified that the CLI fails to parse the project.
* Inserted a `version` field into `package-lock.json` and recorded the result of running analysis with `--output`.
* Removed the `version` field, applied the changes in this PR, and then recorded the result of running analysis with `--output`.
* Compared the two recorded results (recorded in [pkglock-outputs.zip](https://github.com/fossas/spectrometer/files/7515361/pkglock-outputs.zip)), and observed they are identical.

## Risks

I don't think this is very risky, as far as I can tell this field is totally unused.

## References

Closes https://github.com/fossas/team-analysis/issues/810

## Checklist

- [x] I updated `*schema.json` if I have made changes for `.fossa.yml`, `fossa-deps.{json, yaml, yml}`. You may also need to update these if you have added/removed new dependency (e.g. pip) or analysis target type (e.g. poetry).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Spectrometer Changelog
 
+## v2.19.7
+
+- Resolves a regression when parsing npm `package-lock.json` files that do not contain a `version` field ([#445](https://github.com/fossas/spectrometer/pull/445))
+
 ## v2.19.6
 
 - Special cases scans with a single VSI only filter to skip other analysis strategies ([#407](https://github.com/fossas/spectrometer/pull/407))
--- a/src/Strategy/Node/Npm/PackageLock.hs
+++ b/src/Strategy/Node/Npm/PackageLock.hs
@@ -29,7 +29,6 @@ import Strategy.Node.PackageJson (FlatDeps (directDeps), NodePackage (pkgName), 
 
 data NpmPackageJson = NpmPackageJson
   { packageName :: Text
-  , packageVersion :: Text
   , packageDependencies :: Map Text NpmDep
   }
   deriving (Eq, Ord, Show)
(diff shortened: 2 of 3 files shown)


liftM:
It might be more accurate to make this an optional field instead.

zlav:
@liftM I don't think we gain anything by parsing this field, at the moment it's useless metadata so I'm not convinced we need it.

zlav (approved):
LGTM! And we even have a test file that already has the version for parsing.

liftM:
> @liftM I don't think we gain anything by parsing this field, at the moment it's useless metadata so I'm not convinced we need it.

Yeah, I just worry that we make the intent of the code ("be a parser for package.json") less clear by removing this field than by parsing but not using the field.

kitified:
I also considered making it optional instead, but I ultimately decided against it because we don't parse _most_ of the fields inside of `package-lock.json`. Of the available fields (`name`, `version`, `lockfileVersion`, `packages`, `dependencies`) we only parse `name`, ~~`version`,~~ and `dependencies`.

I'm happy to go either way though.

kitified:
Talked with Leo, this isn't a blocker. Merging.

zlav:
@liftM I think the name of the type followed by a `FromJson` makes it clear that we're parsing a package.json file.

I don't know why, but I feel very strongly about not parsing fields that aren't being used. I think it comes from looking back into CLI v1 code where I was parsing metadata and getting frustrated trying to find if the fields were important. Personally, I think parsers should parse only what is needed with no exceptions.

